PR #17, opened 2015-12-13
Use `ulong` instead of `long` to store message id

jhgg:
Or you'll get an overflow error!


jhgg:
note: probably have to replace long w/ ulong everwhere else you're using it to track an ID too.

PR #26, opened 2016-01-25
Change built in help & Custom command activation

Googie2149:
Finally removed the placeholders...


RogueException:
Thanks for the PR~

We still need some sort of distinction between the different types though. Something like [{arg.Name}..] is probably good for multiple, but I have no idea how to show unparsed.

Also you only removed one square bracket for unparsed :P

PR #29, opened 2016-02-10
Fix Server GetIconUrl string

jarveson:
Small fix to return the correct server icon url


RogueException:
Whoops... nice catch - thanks for the PR :D

PR #33, opened 2016-02-13
Added User.HasRole(ulong roleId)

SSStormy:
Allows O(1) lookup for user roles without having a reference to a Role class.


SSStormy:
Hopefully the other accidental commits that I made don't interfere. I really should not be allowed to use git.

RogueException:
We'll discuss this more sometime in my Discord channel

PR #34, opened 2016-02-14
fix for MessageQueue error logging, per request of @bloodskilled

420foxbot:


RogueException:
I'll do a more complete solution, but thanks for the suggestion :D

PR #49, opened 2016-04-12
Removed hack

khionu:
Code now assumes client is logged in as Bot Account, without 1 voice channel limit


RogueException:
Thanks for the PR Khio~

khionu:
x33 Happy to help

PR #52, opened 2016-04-24
Add `User.IsBot` field, bump version number to 0.9.1

420foxbot:
A few users requested an IsBot field for marking Bot Accounts;

Bumping Version number because it has been 0.9-rc3 forever and that's caused a bit of confusion in the channel with what version users are on.


RogueException:
Looks good to me

PR #54, opened 2016-04-30
Add Nicknames to User Model; Bump README to 0.9.1

420foxbot:


src/Discord.Net/API/Client/Common/Member.cs
```diff
@@ -10,5 +10,7 @@ public class Member : MemberReference
         public DateTime? JoinedAt { get; set; }
         [JsonProperty("roles"), JsonConverter(typeof(LongStringArrayConverter))]
         public ulong[] Roles { get; set; }
+        [JsonProperty("nick")]
+        public string Nick { get; set; } = "";
```
420foxbot: Fix because Discord is a dolt about sending payloads; `""` will never be a valid payload for nicknames.

If key is present in JSON:
    - if value is not null: the user's nick was changed
    - if value is null: the user's nick was reset
If key is not present in JSON:
    - the user's nick was unchanged

src/Discord.Net/Models/User.cs
```diff
@@ -175,6 +177,8 @@ internal void Update(APIMember model)
 				JoinedAt = model.JoinedAt.Value;
 			if (model.Roles != null)
 				UpdateRoles(model.Roles.Select(x => Server.GetRole(x)));
+            if (model.Nick != "")
```
420foxbot: See above note.

RogueException:
LGTM, thanks fox.

PR #55, opened 2016-04-30
Added SelfBot Functionality by Config

khionu:
Adds `CommandConfigBuilder.IsSelfBot`, default as false
When true, the bot ignores everyone but itself, as far as commands. This is the behavior expected of SelfBots, as inferred from the Discord stance.


RogueException:
I'll add this for now, but 1.0 will likely have a different approach.

khionu:
<3 Understood. As long as the issue is addressed, I'm happy

PR #56, opened 2016-04-30
Support Role Mentions

Lirusaito:
`Message.MentionedRoles` is now populated properly with roles.
`Role.IsMentionable` has been added to determine whether
`Role.Mention` now mentions roles that are `IsMentionable` as expected (otherwise returns empty string)
`Role.Edit` now takes parameter `bool? isMentionable`
`Server.CreateRole` now takes parameter `bool isMentionable`
`IsMentionable` added to `UpdateRoleRequest`


RogueException:
Looks good, thanks Liru.

PR #57, opened 2016-05-04
Add support for changing Nicknames!

Lirusaito:
```cs
User.Edit(nickname: "foo");
```

### DIFF
--- a/src/Discord.Net.Net45/API/Client/Rest/UpdateOwnNick.cs
+++ b/src/Discord.Net.Net45/API/Client/Rest/UpdateOwnNick.cs
@@ -1,0 +1,23 @@
+﻿using Newtonsoft.Json;
+
+namespace Discord.API.Client.Rest
+{
+    [JsonObject(MemberSerialization.OptIn)]
+    public class UpdateOwnNick : IRestRequest
+    {
+        string IRestRequest.Method => "PATCH";
+        string IRestRequest.Endpoint => $"guilds/{GuildId}/members/@me/nick";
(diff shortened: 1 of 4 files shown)


RogueException:
Does this work for yourself? The client hits a members/@me/nick endpoint instead.

Lirusaito:
Works for me, I have a command
```cs

            group.CreateCommand("nickname")
                .MinPermissions(10)
                .Parameter("nick", Commands.ParameterType.Unparsed)
                .Do(e => e.User.Edit(nickname: string.Join(" ", e.Args)));
```
And here's a log from its use:
```
[12:14 AM] Liru: ~nickname woof
[12:14 AM] BOTDog: Liru: Error: Command Error: The server responded with error 403 (Forbidden)
[12:15 AM] Liru: ~nickname woof
[12:17 AM] woof: ~nickname
[12:17 AM] Liru: Yay
```
(The first time it didn't work because of the new role hierarchy mess.

Lirusaito:
OH! You mean editing the bot's own nickname? I hadn't tried that, gimme a moment.

Lirusaito:
Ah, you are right!

Lirusaito:
Any ideas for how we can have an Edit with nickname with other params set too that can work on self?

Lirusaito:
Alright, tell me what you think of this... not sure I like it myself.

RogueException:
Looks good. I also cleaned up User.cs a bit to match 1.0's logic.

Thanks for the PR, Liru~

PR #96, opened 2016-06-19
BulkMessageDelete Exception

khionu:
Added OutOfBounds exception for DeleteMessages, for when Message Count is less than 2 or greater than 100


RogueException:
This isn't really user friendly. If less than 2, it should redirect to deleting a single message.

RogueException:
LGTM

PR #105, opened 2016-07-05
Added XML Doc summaries to the Client Events

AdenFlorian:
Added XML Doc summaries to the Client Events to help users of the library decide which events to use. Based off of [event descriptions in the Discord API](https://discordapp.com/developers/docs/topics/gateway)


RogueException:
Looks good except for the not implemented placeholders - those are actually events that were removed. We should probably Obsolete attrib those instead.

AdenFlorian:
@RogueException Done in 5195e6b

RogueException:
LGTM, thanks!

PR #106, opened 2016-07-05
Created events for when a user joins or leaves a voice channel

AdenFlorian:
`UserJoinedVoiceChannel` and `UserLeftVoiceChannel`


AdenFlorian:
Why only one?

RogueException:
I'm unsure if this is needed, we already have an event for general user updates. If we add this, where do we draw the line? What about username changes, game changes, etc.

I merged everything into a single event because I thought it made it easier having all user changes in one place.

AdenFlorian:
@RogueException Ok, cool, that's kind of what I was wondering I guess (where y'all wanted to draw the line).

RogueException:
Closing this, unless someone has a good argument of why it should be added

PR #110, opened 2016-07-09
Fix unintended avatar reset

khionu:
I like having an avatar after changing my username ^p^


khionu:
Please. I had already thought of it, I just didn't think it would work, for some reason.

RogueException:
The point of Optional in the model is to avoid having to add null or default value handler, which was causing problems before (see nicknames). This issue looks like its related to a different bug in Optional.

RogueException:
Underlying issue fixed in https://github.com/RogueException/Discord.Net/commit/953f3b033da8c233cf6c7a124454eebe7d2fe6d7

PR #111, opened 2016-07-09
Add null-conditional to MessageManager

420foxbot:
Downloading Messages from a DM Channel was causing a cast on Line 64 to throw a nullref.


RogueException:
LGTM <3

PR #124, opened 2016-07-15
Add Support for Gateway v6

### LINKED ISSUE #123: Add Support for Gateway v6

420foxbot:
Resolves #123 

Adds basic support for API/Gateway v6.

**This does not add support for groups;** per design I thought it would be best to implement that in a different PR.

As a result, all events received from a group will throw a warning on the logger.


RogueException:
Merging these model changes into #125

PR #139, opened 2016-07-20
Add Dependency Injection and Autoloading to the Commands service

### LINKED ISSUE #128: CommandsService.LoadAssembly should not attempt to install modules without a valid constructor.
In the current behavior, if a module has an invalid constructor (e.g. requiring a DiscordSocketClient), the `LoadAssembly` method will throw an `InvalidOperationException` and crash.

In some cases, users may intend to have only certain modules use custom constructors, while leaving the rest of their modules default.

Projected Change: If the assembly crawler fails to find a valid constructor for a class, skip over that module, rather than hard-failing.

I can create a fix if this issue is deemed unintended behavior.

DigiTechs:
Resolves #128 and improves on foxbot's existing code.

Modules can now be constructed using something like this:
```cs
public Module(ClassA a, ClassB b, CommandService service) { }
```
If the code has any other parameters other than CommandService, a DependencyMap is required to be passed to LoadAssembly; otherwise the code will throw an exception stating that dependency injection is required.

### DIFF
--- a/src/Discord.Net.Commands/ReflectionUtils.cs
+++ b/src/Discord.Net.Commands/ReflectionUtils.cs
@@ -17,25 +17,50 @@ namespace Discord.Commands
             if (constructor == null)
                 throw new InvalidOperationException($"Found no constructor for \"{typeInfo.FullName}\"");
 
-            object[] parameters;
-            try
+            object[] arguments = null;
+
+            ParameterInfo[] parameters = constructor.GetParameters();
+
+            // TODO: can this logic be made better/cleaner?
+            if (parameters.Length == 1)
             {
-                // TODO: probably change this ternary into something sensible
-                parameters = constructor.GetParameters()
-                    .Select(x => x.ParameterType == typeof(CommandService) ? service : map.Get(x.ParameterType)).ToArray();
+                if (parameters[0].ParameterType == typeof(IDependencyMap))
+                {
+                    if (map != null)
+                        arguments = new object[] { map };
+                    else
+                        throw new InvalidOperationException($"Could not find a valid constructor for \"{typeInfo.FullName}\" (an IDependencyMap is required)");
+                }
             }
-            catch (KeyNotFoundException ex) // tried to inject an invalid dependency
+            else if (parameters.Length == 2)
             {
-                throw new InvalidOperationException($"Could not find a valid constructor for \"{typeInfo.FullName}\" (could not provide parameter)", ex);
+                if (parameters[0].ParameterType == typeof(CommandService) && parameters[1].ParameterType == typeof(IDependencyMap))
+                    if (map != null)
+                        arguments = new object[] { service, map };
+                    else
+                        throw new InvalidOperationException($"Could not find a valid constructor for \"{typeInfo.FullName}\" (an IDependencyMap is required)");
             }
-            catch (NullReferenceException ex) // tried to find a dependency
+
+            if (arguments == null)
             {
-                throw new InvalidOperationException($"Could not find a valid constructor for \"{typeInfo.FullName}\" (type requires dependency injection)", ex);
+                try
+                {
+                    // TODO: probably change this ternary into something sensible?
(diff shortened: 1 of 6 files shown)


src/Discord.Net.Commands/ReflectionUtils.cs
```diff
@@ -6,18 +6,33 @@ namespace Discord.Commands
...
+                    return constructor.Invoke(new object[1] { map });
+                }
+                else
+                    throw new InvalidOperationException($"Could not find a valid constructor for \"{typeInfo.FullName}\"");
             }
-            catch (Exception ex)
+            catch
             {
-                throw new InvalidOperationException($"Failed to create \"{typeInfo.FullName}\"", ex);
+                throw new InvalidOperationException($"Could not find a valid constructor for \"{typeInfo.FullName}\"");
```
RogueException: Needs a better msg, also make sure to include the caught exception as an inner exception.

src/Discord.Net.Commands/ReflectionUtils.cs
```diff
@@ -6,18 +6,33 @@ namespace Discord.Commands
...
-                throw new InvalidOperationException($"Failed to find a valid constructor for \"{typeInfo.FullName}\"");
+            if (typeInfo.DeclaredConstructors.Count() > 1)
+                throw new InvalidOperationException($"Found too many constructors for \"{typeInfo.FullName}\"");
+            var constructor = typeInfo.DeclaredConstructors.FirstOrDefault();
             try
             {
-                return constructor.Invoke(null);
+                if (constructor.GetParameters().Length == 0)
+                    return constructor.Invoke(null);
+                else if (constructor.GetParameters().Length > 1)
```
RogueException: Why does it exception on more than one parameter?

src/Discord.Net.Commands/ReflectionUtils.cs
```diff
@@ -6,18 +6,33 @@ namespace Discord.Commands
...
+                    throw new InvalidOperationException($"Could not find a valid constructor for \"{typeInfo.FullName}\"");
+                var parameter = constructor.GetParameters().FirstOrDefault();
```
RogueException: Should probably cache the result of .GetParameters(), you use it in several places and reflection is slow.

src/Discord.Net.Commands/ReflectionUtils.cs
```diff
@@ -1,23 +1,70 @@
...
+
             try
             {
-                return constructor.Invoke(null);
+                return constructor.Invoke(arguments);
             }
             catch (Exception ex)
             {
-                throw new InvalidOperationException($"Failed to create \"{typeInfo.FullName}\"", ex);
+                throw new InvalidOperationException($"Could not find a valid constructor for \"{typeInfo.FullName}\" (Error invoking constructor)", ex);
```
RogueException: I'd just simplify this to "Failed to create object", since the exception could come from user code.

src/Discord.Net.Commands/ReflectionUtils.cs
```diff
@@ -1,23 +1,70 @@
...
+                        throw new InvalidOperationException($"Could not find a valid constructor for \"{typeInfo.FullName}\" (an IDependencyMap is required)");
+                }
+            }
+            else if (parameters.Length == 2)
+            {
+                if (parameters[0].ParameterType == typeof(CommandService) && parameters[1].ParameterType == typeof(IDependencyMap))
+                    if (map != null)
+                        arguments = new object[] { service, map };
+                    else
+                        throw new InvalidOperationException($"Could not find a valid constructor for \"{typeInfo.FullName}\" (an IDependencyMap is required)");
```
RogueException: I thought we were going to inject parameters directly?

RogueException:
Removed my old comments, was confused by your code until I realized you're flattening the parameters *and* allowing to accept the dependency map directly

RogueException:
Just gotta fix up that one exception.

I'll probably do a quick sweep after and clean things up, but LGTM.

PR #145, opened 2016-07-25
Changed CommandServiceConfig constructor to public

jankcat:
Changed CommandServiceConfig constructor to public, so that it can be instantiated by projects implementing the library... so that commands can be registered.


khionu:
This is unnecessary, you can look at [this example](https://github.com/RogueException/DiscordBot/blob/master/src/DiscordBot/Program.cs#L48:L54) on how to do it properly. 0.9.x is soon to be phased out, as 1.0 is nearing Beta. At this stage of 0.9.x's lifespan, I think this serves no good.

jankcat:
Oh great. Yeah I saw this documentation after the fact. http://rtd.discord.foxbot.me/en/legacy/features/commands.html

Thanks

PR #149, opened 2016-07-28
Ignore CHANNEL_PINS_UPDATE

### LINKED ISSUE #135: Support CHANNEL_PINS_UPDATE

420foxbot:
This gateway event is raised when a message in a GuildChannel is pinned or unpinned. Unfortunately, this event only contains a timestamp of when the pin was changed, making it effectively useless. As a result, it will be dropped.

Resolves #135


RogueException:
LGTM

PR #156, opened 2016-07-29
Added Sharding support

Carbonitex:


RogueException:
When is InformationVersion not available?

RogueException:
I'm just going to handle both PRs here lol

src/Discord.Net/DiscordConfig.cs
```diff
@@ -5,7 +5,7 @@ namespace Discord
     public class DiscordConfig
     {
         public const int APIVersion = 6;        
-        public static string Version { get; } = typeof(DiscordRestConfig).GetTypeInfo().Assembly?.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion ?? "Unknown";
+        public static string Version { get; } = typeof(DiscordRestConfig).GetTypeInfo().Assembly?.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? "Unknown";
```
RogueException: When is InformationVersion not available? Rather than returning Unknown, it should fall back to the old way of getting version: Assembly.GetName().Version.ToString(3)

src/Discord.Net/API/Gateway/IdentifyParams.cs
```diff
@@ -14,5 +14,7 @@ public class IdentifyParams
         public int LargeThreshold { get; set; }
         [JsonProperty("compress")]
         public bool UseCompression { get; set; }
+        [JsonProperty("shard", NullValueHandling = NullValueHandling.Ignore)]
```
RogueException: We avoid using Null/Default ValueHandling - use the Optional<T> type instead which on serializes if it was assigned a value.

src/Discord.Net/API/DiscordSocketApiClient.cs
```diff
@@ -190,8 +190,10 @@ internal override void Dispose(bool disposing)
                 Token = _authToken,
                 Properties = props,
                 LargeThreshold = largeThreshold,
-                UseCompression = useCompression
+                UseCompression = useCompression,
+                ShardingParams = totalShards > 1 ? new int[] { shardID, totalShards } : null
```
RogueException: See other comment about Optional<T>s, dont assign this value at all if it isn't needed.

PR #157, opened 2016-07-30
Remove all references to XKCD URLs in Invites

### LINKED ISSUE #153: Xkcd should be removed from IInvite
It looks like the client has removed Xkcd URLs, and they can't be generated anymore. I have not checked to see if the API reflects this yet.

420foxbot:
Resolves #153

This is a breaking change!

With the introduction of Vanity URLs and the shortening of Invite URLs, the API appears to have removed the "xkcdpass" field of the Invite. 

Since this functionality was removed from the Discord API and thus obsolete, it has been removed from the Invite models.


RogueException:
LGTM

PR #161, opened 2016-07-30
Add support for 'params' in commands

DigiTechs:
This should work with any existing code and adds support for commands which have arrays as their last argument. Unfortunately it doesn't seem possible to require usage of 'params', hence the TODO comment in the Command class.

I also added a new property to CommandParameter which may not be needed (UnderlyingType) so that may be able to be cleaned up more.


khionu:
LGTM, but I'm not certain about those TODOs :/

PR #162, opened 2016-07-30
Added Extension Methods for WebSocket

### LINKED ISSUE #146: Add extension methods specific to the type of connection
While not required, these methods should really help out users by not requiring async for all actions.

We should also look into moving GetCachedMessage from IMessageChannel to an extension method for WS.

420foxbot:
Resolves #146.

Adds extension methods to `IGuildChannel` and `IChannel` types for WebSocket users.


khionu:
LGTM

RogueException:
What about adding a sync way to get the existing collections?

Channel.GetUsers instead of Channel.GetUsersAsync

RogueException:
ICollection~

RogueException:
LGTM

PR #163, opened 2016-07-30
Added `Command.Synopsis`

khionu:
Added `Command.Synopsis` for separation of short and long descriptions, this being intended for long.


khionu:
Good point.... I would make that a 1.1 issue, though, unless further changes make it convenient to add in.

PR #164, opened 2016-07-30
Don't load modules that are already loaded

### LINKED ISSUE #160: Commands should not be added again if module autodetection is done twice

420foxbot:
Resolves #160.

Previously, if a user autoloaded commands more than once, commands that were already in the command map would be readded. 

If the module list already contains a module with the same type as the module being loaded, it will not load the new instance of this module.

### DIFF
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -12,7 +12,7 @@ namespace Discord.Commands
     public class CommandService
     {
         private readonly SemaphoreSlim _moduleLock;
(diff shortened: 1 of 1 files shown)


khionu:
LGTM

RogueException:
Maybe we should make modules key on their type instead of instance? This check was already supposed to exist: https://github.com/RogueException/Discord.Net/blob/9348e087b016acd87721fbb69ce0588f598d1bb4/src/Discord.Net.Commands/CommandService.cs#L103

RogueException:
LGTM

PR #167, opened 2016-07-31
Finish SocketVoiceChannel.ConnectAsync

DigiTechs:
This implements the Promise we were talking about and allows an AudioClient to successfully connect.

### DIFF
--- a/src/Discord.Net/WebSocket/Entities/Guilds/SocketGuild.cs
+++ b/src/Discord.Net/WebSocket/Entities/Guilds/SocketGuild.cs
@@ -24,7 +24,7 @@ namespace Discord
         internal override bool IsAttached => true;
 
         private readonly SemaphoreSlim _audioLock;
-        private TaskCompletionSource<bool> _syncPromise, _downloaderPromise, _audioPromise;
+        private TaskCompletionSource<bool> _syncPromise, _downloaderPromise, _audioConnectPromise;
         private ConcurrentHashSet<ulong> _channels;
         private ConcurrentDictionary<ulong, SocketGuildUser> _members;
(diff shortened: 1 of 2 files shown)


src/Discord.Net/WebSocket/Entities/Guilds/SocketGuild.cs
```diff
@@ -307,6 +308,7 @@ public VoiceState AddOrUpdateVoiceState(VoiceStateModel model, DataStore dataSto
                 _audioLock.Release();
             }
             await audioClient.ConnectAsync(url, CurrentUser.Id, voiceState.VoiceSessionId, token).ConfigureAwait(false);
```
RogueException: Should also have a try-catch around .ConnectAsync. If it exceptions, feed it to the promise.

src/Discord.Net/WebSocket/Entities/Guilds/SocketGuild.cs
```diff
@@ -307,6 +308,7 @@ public VoiceState AddOrUpdateVoiceState(VoiceStateModel model, DataStore dataSto
...
+            _audioConnectPromise.TrySetResult(true);
```
RogueException: Use the TrySetResultAsync helper method to ensure the continuation runs on a non-gateway task.

src/Discord.Net/WebSocket/Entities/Guilds/SocketGuild.cs
RogueException: Should also have a try-catch around .ConnectAsync. If it exceptions, feed it to the promise.

PR #178, opened 2016-08-02
Command tweaks

DigiTechs:
Most of these tweaks make documentation of commands easier for bot developers who want to do things their own way.

### DIFF
--- a/src/Discord.Net.Commands/Attributes/DescriptionAttribute.cs
+++ b/src/Discord.Net.Commands/Attributes/DescriptionAttribute.cs
@@ -2,6 +2,7 @@
 
 namespace Discord.Commands
 {
+    // Full summary of method/parameter
     [AttributeUsage(AttributeTargets.Method | AttributeTargets.Parameter)]
     public class DescriptionAttribute : Attribute
     {
@@ -12,7 +13,8 @@ namespace Discord.Commands
         }
     }
 
-    [AttributeUsage(AttributeTargets.Method)]
+    // Brief summary of method/module
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class)]
     public class SynopsisAttribute : Attribute
     {
         public string Text { get; }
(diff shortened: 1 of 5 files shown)


src/Discord.Net.Commands/Attributes/DescriptionAttribute.cs
```diff
@@ -12,7 +13,8 @@ public DescriptionAttribute(string text)
         }
     }
 
-    [AttributeUsage(AttributeTargets.Method)]
+    // Brief summary of method/module
```
khionu: Uhhhh.... Synopsis was intended to be the Longer summary, as Description is for Parameters, too.... A Synopsis is something intended to give an overview of something, beginning to end.

DigiTechs: Synopsis is usually defined as 'a brief summary' (in contrast to description, which is defined as 'a spoken or written account of a person, object, or event' - basically a 'full summary') which is why I changed the meanings about.

khionu:
LGTM

RogueException:
Modules now being accessible via `Commands` now makes it less clear imo
Also why are we using `Synopsis`? Was this a discussion I missed? Feels less clear than something like `Summary`, especially when you guys already arent able to decipher its meaning at first glance.

khionu:
I suggested it, asked twice if anyone had any issues with the phrasing :/ I thought I was clear at intended usage....

DigiTechs:
I can add the Modules enumerable back if that would be preferred. I also renamed Synopsis to Summary to make the differences between it and Description clear.

RogueException:
I'd say revert https://github.com/RogueException/Discord.Net/pull/178/commits/3856d435876b4aab3833ba09fe2c878071be3a0a and it should be good.

We dont use ILookup anywhere in the lib, having it here isn't consistent (and I dont like it much anyway :P).

After that's fixed, LGTM.

PR #179, opened 2016-08-03
Preconditions for commands

### LINKED ISSUE #172: PreConditions for Commands
What we have already agreed on:
- To provide a Levels system, and a replacement for `IPermissionChecker`s

What I would recommend:
- Having the checks be done before `CommandService.Execute`, not as a part of, so Users have flexibility in that regards. Maybe as a part of `CommandService.Search`?
- Making use of our DI, again, for the IPermissionChecker's, providing `IMessage` and `IDependencyMap`

DigiTechs:
Resolves #172 after our discussion about it. Feedback is welcome.

*(if anybody asks about the .gitignore, I accidentally forgot to unstage it)*


RogueException:
Looking good so far, a few thoughts:

- Preconditions should be able to be set at the group and module levels
- Unsure how I feel about RequireRole, name-based permissions feel dirty, and using an id doesnt upport multiple guilds
- RequireDM has the wrong check in it
- We need a precondition for checking actual discord permissions
- Doesn't make sense to have the precondition check in Parse(), leave it up to bot dev if doing it manually, or they can rely on the auto check you added to the shorthand Execute

RogueException:
Also unsure about the naming of RequireDM/RequireGuild - what does everyone think?

khionu:
Re RequireDM/RequireGuild, how about `DMCommand`/`GuildCommand`?

Also, I agree about the RequireRole, that's not something we should do. Yes, there are several bots that require a role of a particular name, but they should handle that, it's easy enough.

DigiTechs:
Okay, I've added the precondition to check individual permissions, and tweaked the existing preconditions a little. (adding an AttributeUsage attribute to them)

We've already discussed RequireDM and it does have the correct check, and the precondition check isn't in Parse. (it's GitHub's diff viewer showing it wrong)

I've left the RequireRole attribute for now (we can always delete it later) as we should have a discussion on that.

RogueException:
Oh so it is, my bad. Either way, I think Preconditions shouldn't be in Execute(context, result) and be it's own separate step. This gives people more control with logging, or if they want to add a secondary precondition system alongside ours.

RogueException:
Other than that, LGTM.

src/Discord.Net.Commands/Attributes/Preconditions/RequirePermission.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Threading.Tasks;
+
+namespace Discord.Commands.Attributes.Preconditions
+{
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = true, Inherited = true)]
+    public class RequirePermission : RequireGuildAttribute
+    {
+        public GuildPermission? GuildPermission { get; set; }
+        public ChannelPermission? ChannelPermission { get; set; }
+
```
Anonbun: I don't think it makes a ton of sense to have properties on attributes have a public set.

DigiTechs: I decided to leave it this way for developers who wanted to define their permission using both; something like this:

```cs
[RequirePermission(GuildPermission = GuildPermission.ABC, ChannelPermission = ChannelPermission.DEF)]
```

Having a public setter allows properties to be set by name; like the AllowMultiple and Inherited properties in AttributeUsage.

Anonbun: It just seems odd to have constructors for one property or the other but not both. This is pretty insignificant though, so it's fine if it's left like this.

RogueException:
Is this ready to merge?

DigiTechs:
I'd say this is as ready as it'll ever be - I don't think there's anything we're obviously missing nor anything that needs to be changed before merging.

PR #181, opened 2016-08-04
Add ContainsId<T>(T id) extension to IEnumerable<IEntity<T>>

ghost:
This should allow bot devs to check directly if IEntity types like IRole exist within an IEnumerable based on their id. For example; `GuildUser.Roles.ContainsId(12345)`

### DIFF
--- a/src/Discord.Net/Entities/Roles/IRole.cs
+++ b/src/Discord.Net/Entities/Roles/IRole.cs
@@ -5,7 +5,7 @@ using Discord.API.Rest;
 
 namespace Discord
 {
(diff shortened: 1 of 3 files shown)


DigiTechs:
After a discussion, it seems like the original approach was a Bad Idea (tm) - the new approach of adding an extension method to `IEnumerable<IEntity<T>>` seems to be the better approach which should still work for these people.

RogueException:
I think we're going to have another talk about this, postponed til we do

FiniteReality:
Closing since #912 is a better place to discuss this, and this hasn't been touched in forever.

PR #186, opened 2016-08-07
Add Sync Extensions for DiscordSocketClient

### LINKED ISSUE #185: Sync methods should be exposed on DiscordSocketClient
Unsure if we want to make these extensions or not, but DiscordSocketClient is missing some sync calls like CurrentUser, GetVoiceRegion, PrivateChannels, etc

420foxbot:
Resolves #185

### DIFF
--- a/src/Discord.Net/WebSocket/Extensions/SocketClientExtensions.cs
+++ b/src/Discord.Net/WebSocket/Extensions/SocketClientExtensions.cs
@@ -1,0 +1,49 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+namespace Discord.WebSocket
+{
+    public static class SocketClientExtensions
+    {
+        public static IVoiceRegion GetVoiceRegion(this DiscordSocketClient client, string id)
+        {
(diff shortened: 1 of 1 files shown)


src/Discord.Net/WebSocket/Extensions/SocketClientExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        public static GuildEmbed? GetGuildEmbed(this DiscordSocketClient client, ulong id)
+        {
+            var guild = client.DataStore.GetGuild(id);
+            if (guild != null)
+                return new GuildEmbed(guild.IsEmbeddable, guild.EmbedChannelId);
+            return null;
+        }
+
+        public static IReadOnlyCollection<IUserGuild> GetGuildSummaries(this DiscordSocketClient client) =>
```
RogueException: Unsure if we should expose this one, especially since it's the same as GetGuilds

RogueException:
I'd say remove GetGuildSummaries, and you can merge.

PR #193, opened 2016-08-12
Added a flag "ForceWhitespace" on module attribute

Kwoth:
In order to opt out of forced whitespaces after module prefix.

Its default is set to true, in order to not break older code. I think this should be default behavior though

### DIFF
--- a/src/Discord.Net.Commands/Attributes/ModuleAttribute.cs
+++ b/src/Discord.Net.Commands/Attributes/ModuleAttribute.cs
@@ -6,16 +6,14 @@ namespace Discord.Commands
     public class ModuleAttribute : Attribute
     {
         public string Prefix { get; }
-        public bool AutoLoad { get; set; }
-        public ModuleAttribute()
-        {
-            Prefix = null;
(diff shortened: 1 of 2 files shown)


FiniteReality:
Why did you get rid of the default constructor for ModuleAttribute? What you probably should have done is something like this:

```cs
public ModuleAttribute()
    : this(null)
{
}
public ModuleAttribute(string prefix)
{
    Prefix = prefix;
    AutoLoad = true;
    ForceWhitespace = true;
}
```

That way users can still override their own properties, e.g.:

```cs
[Module("name", ForceWhitespace = false)]
//...
[Module(AutoLoad = false, Prefix = "omg", ForceWhitespace = true)]
//etc.
```

Also, I'm not sure about the name of ForceWhitespace. Doesn't make sense to me... Maybe `AppendSpace`?

Kwoth:
That is odd, i didn't notice. I will fix it asap.

RogueException:
:-1: I see this as being a bad pattern for both us and users, and creates a disconnect between command definitions and the command map

PR #208, opened 2016-08-18
Fixed AddBanAsync(IUser,int) calling itself

Kwoth:
... instead of AddBanAsync(ulong,int)

Fixes StackOverflowException which occurs when AddBanAsync is called.


COREISSOPRO:
How do you get nadeko flowers sorry for posting on a random conversation @Kwoth

RogueException:
LGTM

PR #209, opened 2016-08-18
Add owner/administrator check to RequirePermissionAttribute

### LINKED ISSUE #207: RequirePermission fails for guild owner
We should probably have an explicit check for the guild owner to bypass the entire permissions check, as guild owners have all permissions anyway.

FiniteReality:
Resolves #207.

A 'Fuzzy' property has been introduced for people who want the old behavior of ignoring owner/administrator checks, which can be used something like this:

``` cs
[RequirePermission(permission, Fuzzy = false)]
```


Kwoth:
I'd drop the fuzzy flag. It just wasn't working correctly before, so no need for backwards compat. Server owner has all permissions anyway.

RogueException:
This is pointing to a deeper issue and shouldnt be patched at the attribute level

RogueException:
Was this fixed by https://github.com/RogueException/Discord.Net/commit/24ed9c6c09ce4f748f6e64fb81570708ded252b6 ?

PR #222, opened 2016-08-22
Document public APIs

FiniteReality:
This is an in-progress PR which I'll use to track the state of my work on this.

Feedback on the documentation would be nice as to make sure I'm documenting these correctly.

### DIFF
--- a/src/Discord.Net/Entities/Channels/ChannelType.cs
+++ b/src/Discord.Net/Entities/Channels/ChannelType.cs
@@ -1,10 +1,15 @@
 ﻿namespace Discord
 {
+    /// <summary> Specifies the type of channel a message was sent to or eceived from. </summary>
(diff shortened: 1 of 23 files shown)


src/Discord.Net/Entities/Channels/ChannelType.cs
```diff
@@ -1,10 +1,15 @@
 ﻿namespace Discord
 {
+    /// <summary> Specifies the type of channel a message was sent to or eceived from. </summary>
     public enum ChannelType
     {
+        ///<summary> A text channel </summary>
         Text = 0,
+        ///<summary> A direct-message text channel </summary>
         DM = 1,
+        ///<summary> A voice channel channel </summary>
```
khionu: > voice channel channel

PR #230, opened 2016-08-26
Implement Command Aliases

FiniteReality:
This implementation could probably be more efficient, but I opted for clarity and simplicity where possible.

### DIFF
--- a/src/Discord.Net.Commands/Attributes/AliasAttribute.cs
+++ b/src/Discord.Net.Commands/Attributes/AliasAttribute.cs
@@ -1,0 +1,18 @@
+using System;
+
+namespace Discord.Commands
+{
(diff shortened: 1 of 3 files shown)


FiniteReality:
I don't see why that would be too useful - by the time the user does something with it, it's probably already registered internally. Let's see what Khionu/Voltana think about it before I change anything.

khionu:
I would say leave it. If people are relying on the Attributes rather than the accumulated data in the Command/Module objects, they're coding badly anyways.

RogueException:
I dont really care if it's immutable in the attribute itself. Follow whatever style we have elsewhere, if we use array elsewhere.

You forgot to use the aliases on removing a command.

PR #237, opened 2016-08-27
Split IMessage into IUserMessage and ISystemMessage

RogueException:


RogueException:
whelp, I botched that one :P

FiniteReality:
it gits gud

PR #246, opened 2016-08-29
Fix command aliases (#244)

### LINKED ISSUE #244: Aliases don't actually do anything
Aliases don't appear to work, latest build.

command
alias
command

FiniteReality:
¯\\\_(ツ)\_/¯


FiniteReality:
Okay, scratch the fact that this is in progress. VS Code didn't run the correct version of the test project which made me think the bug still existed. I rebuilt and it's fine.

PR #250, opened 2016-08-29
Add Sanitize function to escape Markdown

### LINKED ISSUE #249: Auto escaping on Format
I would be nice if Fomat also escapes those chararacters which would end the markdown so if i use Format.Bold("test** test") it will display **test\*\* test** instead **test** test**.

FiniteReality:
Resolves #249 

Automatically escaping in the individual formatting methods would be a bad idea because it would prevent chaining, e.g:

```cs
var message = $"Tried to ban user {Format.Bold(Format.Italics(user.Name))}";
```

Instead, the developer should be escaping manually, like so:

```cs
var message = $"Tried to ban user {Format.Bold(Format.Italics(Format.Sanitize(user.Name)))}";
```

### DIFF
--- a/src/Discord.Net/Format.cs
+++ b/src/Discord.Net/Format.cs
@@ -2,6 +2,9 @@
 {
     public static class Format
     {
+        // Characters which need escaping
+        private static string[] SensitiveCharacters = { "*", "_", "~", "`", "\\" };
+
         /// <summary> Returns a markdown-formatted string with bold formatting. </summary>
         public static string Bold(string text) => $"**{text}**";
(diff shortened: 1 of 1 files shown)


Confruggy:
You can also do italics with _ so i dont think this is necessary but of course you should edit Format.Italics then so it does markdown with _ instead *
**_test\*\*test_**
`**_test\*\*test_**`

RogueException:
What's the status of this PR?

FiniteReality:
Last I checked, this was ready to merge. I don't think it's missing anything that people want, but I am open to suggestions.

PR #260, opened 2016-08-31
Allow parameterless commands to build correctly

### LINKED ISSUE #253: Parameterless Command in Group/Module injects extra space
```csharp
[Module("prune"}]
public class PruneCommands {
    [Command]
    public async Task Prune(IUserMessage message) {
    }
}
```

Produces a Command with Text `prune `. This is required in addition to the normal space separator and registers an unknown command error if the extra space is not present.

The same happens with GroupAttribute as well.

FiniteReality:
Resolves #253 and works correctly with aliases/etc.

### DIFF
--- a/src/Discord.Net.Commands/Module.cs
+++ b/src/Discord.Net.Commands/Module.cs
@@ -48,8 +48,6 @@ namespace Discord.Commands
 
         private void SearchClass(TypeInfo parentType, object instance, List<Command> commands, string groupPrefix, IDependencyMap dependencyMap)
         {
-            if (groupPrefix != "")
-                groupPrefix += " ";
             foreach (var method in parentType.DeclaredMethods)
             {
                 var cmdAttr = method.GetCustomAttribute<CommandAttribute>();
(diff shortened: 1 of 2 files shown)


src/Discord.Net.Commands/Module.cs
```diff
@@ -63,9 +61,15 @@ private void SearchClass(TypeInfo parentType, object instance, List<Command> com
                 {
                     string nextGroupPrefix;
                     if (groupAttrib.Prefix != null)
+                    {
+                        if (groupPrefix != "")
```
RogueException: This is inside the type loop. Wont this add multiple spaces if there are multiple commands in a class?

FiniteReality: Oh boy, it probably will. Let me fix that.

src/Discord.Net.Commands/Module.cs
```diff
@@ -63,9 +61,16 @@ private void SearchClass(TypeInfo parentType, object instance, List<Command> com
                 {
                     string nextGroupPrefix;
                     if (groupAttrib.Prefix != null)
```
FiniteReality: Wait, this still introduces a bug. Whoops.

FiniteReality:
I think there's still a bug in Module.cs, or at least a logic error. I'll look into it.

FiniteReality:
That should fix that error; everything should be fine now.

PR #267, opened 2016-09-03
Optimize rest of lib for release builds

Mushroom:
Applies https://github.com/RogueException/Discord.Net/commit/ed7710fbef94ebc0a561b2d4a682026b2ffdad59 to the main project as well as the commands namespace


RogueException:
Optimize should be restricted to the Release build, otherwise it'll interfere with debugging

PR #271, opened 2016-09-06
Fixed Optional argument in Message Updated event

Kwoth:
This is the way MessageDeleted works. Just making it consistent.

Right now, first argument is always returning true on IsSpecified with null as the value if message isn't in the cache. With this fix will return IsSpecified=false when it is null.


khionu:
LGTM

PR #278, opened 2016-09-07
Implemented GetChannels for Discord.DiscordClient.

CodeSurfing145:
I wasn't sure if there is any other option for this but this seemed logical to add.
Useful for looping through the channels like so:

```
foreach (KeyValuePair<ulong, Channel> c in botClient.GetChannels())
{
    Console.WriteLine(c.Value.Name);
}
```

### DIFF
--- a/src/Discord.Net/DiscordClient.cs
+++ b/src/Discord.Net/DiscordClient.cs
@@ -380,6 +380,12 @@ namespace Discord
             }
             return channel;
         }
+
+        public ConcurrentDictionary<ulong, Channel> GetChannels()
+        {
+            return _channels;
+        }
+
         public Channel GetChannel(ulong id)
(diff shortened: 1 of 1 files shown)


CodeSurfing145:
@420foxbot Should I close this one and commit a new one with a conversion to IReadOnlyDict?

Mushroom:
You could just commit again to this PR, no need to make another one

CodeSurfing145:
We probably better wait, as ConcurrentDictionary has no interface implemented yet for IReadOnlyDictionary. As of .NET 4.6 it has this.

http://stackoverflow.com/questions/35970151/unable-to-cast-net-concurrentdictionaryk-v-to-ireadonlydictionaryk-v-interf

PR #287, opened 2016-09-16
Override type reader

Joe4evr:
This is the biggest contribution I've made yet, and I don't know if I got all the cases. Please do mention if I missed anything.

### DIFF
--- a/src/Discord.Net.Commands/Attributes/TypeReaderAttribute.cs
+++ b/src/Discord.Net.Commands/Attributes/TypeReaderAttribute.cs
@@ -1,7 +1,7 @@
 ﻿using System;
 using System.Reflection;
 
-namespace Discord.Commands.Attributes
+namespace Discord.Commands
 {
     [AttributeUsage(AttributeTargets.Class | AttributeTargets.Parameter, AllowMultiple = true)]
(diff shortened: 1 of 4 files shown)


src/Discord.Net.Commands/Module.cs
```diff
@@ -16,6 +19,7 @@ public class Module
         public string Remarks { get; }
         public IEnumerable<Command> Commands { get; }
         internal object Instance { get; }
+        public ConcurrentDictionary<Type, TypeReader> OverridingTypeReaders { get; }
```
RogueException: No need for this to be concurrent, it should be immutable after the module has been created

Joe4evr: Roger. Will change that in the morning.

RogueException:
The rest looks LGTM, has this been tested?

Joe4evr:
*Can* I even test?

khionu:
How could you *not*? O.o

Joe4evr:
Pinging @RogueException, I updated it to fit the new stuff.

Joe4evr:
Closing, as @FiniteReality is taking over.

PR #292, opened 2016-09-27
Added support for Do Not Disturb UserStatus

Sentinent:
This should be all the references to UserStatus. Added DoNotDisturb after Offline to hopefully break as little bots as possible.


RogueException:
Thanks for the PR~

PR #308, opened 2016-10-14
Exclude abstract types from being loaded as modules.

Joe4evr:


FiniteReality:
Loading abstract classes should definitely throw an exception. Looks good to me.

PR #312, opened 2016-10-15
Allow Preconditions on parameters.

Joe4evr:


Joe4evr:
Hmmm.... May need some more work on this, as it doesn't seem like the *value* of the parameter is available. I primarily thought about using this for a `[LimitTo(params string[])]` check so I can do `[LimitTo("foo", "bar")] string someParam` and the command would only be valid if the param is one of those values.

Joe4evr:
Closing, as @FiniteReality is taking over.

PR #315, opened 2016-10-16
Fix #311

### LINKED ISSUE #311: CommandService is not listing all installed Commands or Modules
When you add a new Module to the `CommandService` the field `Modules` only contains the last module and `Commands` only the commands from the last module

james7132:
TypeInfo.BaseType will likely return the same Type if all modules derive from
ModuleBase or some common subclass of it. Making it appear as if only one module
is registered.

Changed to TypeInfo.AsType for expected behavior.


FiniteReality:
LGTM - I'm surprised something like this wasn't picked up on earlier.

RogueException:
Gah, I remember bringing up that property too. Thanks for the PR!

PR #329, opened 2016-10-22
Raise UserUpdated for Avatar/Username changes; GuildMemberUpdated for Nickname

### LINKED ISSUE #274: UserUpdated doesn't trigger on username and avatar changes. Nor does UserPresence
Voltana said
"They give username changes to us through presence updated, but it's not a part of presence"

I'd personally like it to trigger on UserUpdated : )

foxbot:
This should resolve #274, although I opted to raise `GuildMemberUpdated` over `UserUpdated`.

Since nicknames are guild-specific, it would make more sense to raise them in the context of a guild, rather than a global context.

With this change, any previously attached `SocketGuildUser` objects have their `Nickname` field modified.

This also raises `UserUpdated` for Username or Avatar Changes; any previously attached `SocketUser` objects are updated.

### DIFF
--- a/src/Discord.Net.WebSocket/DiscordSocketClient.cs
+++ b/src/Discord.Net.WebSocket/DiscordSocketClient.cs
@@ -1451,7 +1451,7 @@ namespace Discord.WebSocket
                                         }
 
                                         await _userPresenceUpdatedEvent.InvokeAsync(guild, user, before, user.Presence).ConfigureAwait(false);
-                                        if (data.User.Username.IsSpecified || data.Roles.IsSpecified)
+                                        if (data.User.Username.IsSpecified || data.Roles.IsSpecified || data.Nick.IsSpecified)
                                         {
                                             var before2 = user.Clone();
                                             await _guildMemberUpdatedEvent.InvokeAsync(before2, user).ConfigureAwait(false);
--- a/src/Discord.Net.WebSocket/DiscordSocketClient.cs
+++ b/src/Discord.Net.WebSocket/DiscordSocketClient.cs
@@ -1440,13 +1440,13 @@ namespace Discord.WebSocket
 
                                         SocketPresence beforePresence;
                                         SocketGuildUser beforeGuild;
-                                        SocketGlobalUser before;
+                                        SocketGlobalUser beforeGlobal;
                                         var user = guild.GetUser(data.User.Id);
                                         if (user != null)
                                         {
 
                                             beforePresence = user.Presence.Clone();
-                                            before = user.GlobalUser.Clone();
+                                            beforeGlobal = user.GlobalUser.Clone();
                                             beforeGuild = user.Clone();
                                             user.Update(State, data);
                                         }
@@ -1454,7 +1454,7 @@ namespace Discord.WebSocket
                                         {
                                             beforePresence = new SocketPresence(UserStatus.Offline, null);
                                             user = guild.AddOrUpdateUser(data);
-                                            before = user.GlobalUser.Clone();
+                                            beforeGlobal = user.GlobalUser.Clone();
                                             beforeGuild = user.Clone();
                                         }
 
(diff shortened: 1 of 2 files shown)


src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -1355,24 +1356,35 @@ public Task DownloadUsersAsync(params IGuild[] guilds)
                                             return;
                                         }
 
-                                        SocketPresence before;
+                                        SocketPresence beforePresence;
+                                        SocketGuildUser beforeGuild;
+                                        SocketGlobalUser before;
```
420foxbot: I'm not very satisfied with the number of allocations I have to make here, but I currently don't see any other way to do this. The GlobalUser must be cloned individually, because MemberwiseClone doesn't seem to create a new reference when cloning a reference type.

src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -1358,21 +1358,21 @@ public Task DownloadUsersAsync(params IGuild[] guilds)
 
                                         SocketPresence beforePresence;
                                         SocketGuildUser beforeGuild;
-                                        SocketGlobalUser before;
+                                        SocketGlobalUser beforeGlobal;
```
420foxbot: I'm not very satisfied with the number of allocations I have to make here, but I currently don't see any other way to do this. The GlobalUser must be cloned individually, because MemberwiseClone doesn't seem to create a new reference when cloning a reference type.

src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -1355,24 +1356,35 @@ public Task DownloadUsersAsync(params IGuild[] guilds)
...
+                                        SocketGuildUser beforeGuild;
+                                        SocketGlobalUser beforeGlobal;
                                         var user = guild.GetUser(data.User.Id);
                                         if (user != null)
                                         {
-                                            before = user.Presence.Clone();
+
+                                            beforePresence = user.Presence.Clone();
+                                            beforeGlobal = user.GlobalUser.Clone();
+                                            beforeGuild = user.Clone();
```
420foxbot: I'm not very satisfied with the number of allocations I have to make here, but I currently don't see any other way to do this. The GlobalUser must be cloned individually, because MemberwiseClone doesn't seem to create a new reference when cloning a reference type.

RogueException: MemberwiseClone is a shallow clone, which is the main reason we use it - deep clones would be expensive. Having to clone global is one of the side effects of that.

Memory-wise the operation is really cheap since it's just a memcpy iirc. It does involve memory pinning, but I've yet to see an impact.

src/Discord.Net.WebSocket/Entities/Users/SocketGuildUser.cs
```diff
@@ -72,6 +72,10 @@ internal override void Update(ClientState state, PresenceModel model)
                 UpdateRoles(model.Roles.Value);
             if (model.Nick.IsSpecified)
                 Nickname = model.Nick.Value;
+            if (model.User.Username.IsSpecified)
+                GlobalUser.Username = model.User.Username.Value;
+            if (model.User.Avatar.IsSpecified)
+                GlobalUser.AvatarId = model.User.Avatar.Value;
```
RogueException: Lets not repeat logic here, just pass model.User to GlobalUser.Update

src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -1355,24 +1356,35 @@ public Task DownloadUsersAsync(params IGuild[] guilds)
...
                                         else
                                         {
-                                            before = new SocketPresence(UserStatus.Offline, null);
+                                            beforePresence = new SocketPresence(UserStatus.Offline, null);
                                             user = guild.AddOrUpdateUser(data);
+                                            beforeGlobal = user.GlobalUser.Clone();
+                                            beforeGuild = user.Clone();
+                                        }
+
+                                        await _userPresenceUpdatedEvent.InvokeAsync(guild, user, beforePresence, user.Presence).ConfigureAwait(false);
```
RogueException: Are there cases where presence wouldn't have changed in this event? Username change? We should avoid raising the event if there are no changes.

Auralytical (changes requested):

Issue #332, opened 2016-10-23
ReplyAsync should return the message sent


Sentinent:

Joe4evr:
I respectfully disagree, and I posted my reasoning in #333.

PR #333, opened 2016-10-23
Update ReplyAsync Task to return the sent message.

LassieME:
As requested in issue #332.

### DIFF
--- a/src/Discord.Net.Commands/ModuleBase.cs
+++ b/src/Discord.Net.Commands/ModuleBase.cs
@@ -6,9 +6,9 @@ namespace Discord.Commands
     {
         public CommandContext Context { get; internal set; }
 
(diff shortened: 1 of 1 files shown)


Joe4evr:
LTTP, but I respectfully disagree with this change. This will prevent returning the `ReplyAsync()` Task itself without needing to `await` it (and therefor without an additional allocation), like so:
```cs
public override Task NextTurnCmd()
    => GameInProgress ? Game.NextTurn() : ReplyAsync("No game in progress.");
```
If you want the sent message, `Context.Channel.SendMessageAsync()` is still there to use.

RogueException:
@Joe4evr The function returns a message, and as such should return that obj. Omitting it because it slightly complicates one liners is not a good defense :P.

I think that one-liner would still work if you cast the result to a non-generic Task.

Issue #263, opened 2016-09-01
Commands don't inherit their Groups Preconditions


khionu:
![](https://i.lithi.io/ti7p.png)

Here, you can see a command that should inherit `PermissionLevel(5)` from a Group, and a command that has it directly. The command that has it directly respects the PermissionChecker while the command that should inherit it doesn;t.

RogueException:
May be fixed by #361, testing needed

foxbot:
This is resolved as of #365 (was unable to test groups on #361)

foxbot:
Closed as fixed.

PR #334, opened 2016-10-24
Fix #328 by excluding nested types in CommandService

### LINKED ISSUE #328: Commands in groups with similar names are not selected properly for execution.
I have a few commands/groups set up like so:

``` csharp
[Group("temp")]
public class Temp : ModuleBase {
  [Group("role")]
  public class Role : ModuleBase {
     [Command("add")]
     public async Task Add(...) {
     }
  }
}

[Group("role")]
public class Role : ModuleBase {
   [Command("add")]
   public async Task Add(...) {
   }
}
```

When trying to execute `role add`, `temp role add` is found and executed instead.

FiniteReality:
Apparently a nested type counts as an exported type and it is returned by Assembly.ExportedTypes - so we do a check to ensure that the type isn't nested before adding it ourselves.

### DIFF
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -95,6 +95,16 @@ namespace Discord.Commands
             {
                 foreach (var type in assembly.ExportedTypes)
                 {
+                    //Ensure that we weren't declared as a submodule
+                    if (type.DeclaringType != null)
+                    {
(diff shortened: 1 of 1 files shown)


RogueException:
I'm unsure if this is how I want to approach this. We'll discuss soon.

RogueException:
Rather than do a top-down approach, we're going to search for public classes, and then work our way upward. This will simplify search code and also fix #263 in the process.

Issue #248, opened 2016-08-29
Hierarchy Helper?


khionu:
The hierarchy is messy business, and I think a function that takes two IGuildUsers and returns true if the first is above the second would be an extremely helpful addition.

To note. according to Meew0, hierarchy is determined by Role Position (higher), and tie breakers are by Role Id (smaller).

RogueException:
Has Discord officially doc'd this yet? I'm wary of adding this since it could just up and change again :P

khionu:
No, and idk if they plan to. Afaicr, they didn't discuss much after they made the change, And as much as I would like it to change, they made it pretty clear that they won't

RogueException:
This would be good to have

PR #336, opened 2016-10-25
Implement IComparable<T> on IRole, add extension IRole#CompareTo

james7132:
Allows GuildUsers and Roles to be compared with each other in a reasonable manner. Addresses the issue mentioned in #248.


RogueException:
Yeah, I agree that this shouldn't be added as part of comparables. In particular, it's dangerous how this redefines equality. Does `u1.CompareTo(u2) == 0` mean that users are the same, or that they have the same relative power?

There's no harm in adding a new method to users though.

james7132:
Perhaps keep it on IRole where it makes sense, and change it to a ``CompareRoles`` extension method?

Issue #262, opened 2016-08-31
Quoted parameters are not properly parsed for params arguments


foxbot:
**Test case:**
`command "this is one param" another andanother "and another"`

**Expected Result:**
`-> "this is one param", another, andanother, "and another"`

**Actual Result:**
Command fails to parse, `A quoted parameter is incomplete`.

PR #343, opened 2016-10-30
Fix parsing double quoted parameters (#262)

AdamPatenaude:
This fixes (or adds support for, i'm not really sure, looking at the code, that it was originally intended) parsing for parameters in double quotes as mentioned in #262.
This doesn't include the quotes with the parameter and also doesn't work with single quotes but both could be fixed easily.

### DIFF
--- a/src/Discord.Net.Commands/CommandParser.cs
+++ b/src/Discord.Net.Commands/CommandParser.cs
@@ -66,6 +66,7 @@ namespace Discord.Commands
 			int endPosition = startPos;
 			int inputLength = input.Length;
 			bool isEscaped = false;
+			bool isDoubleQuoted = false;
 
 			var expectedArgs = command._parameters;
 			List<string> argList = new List<string>();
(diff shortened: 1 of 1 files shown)


khionu:
Umm..... #262 is referring to the Dev Branch..... this functionality already works in 0.9 (Master Branch)

AdamPatenaude:
Well then, I tried. 

But from what I've seen, the algorithm in the Master branch simply ignores quotes and doesn't actually consider everything in them as a parameter.
I've done my own testing, but not in the "official" tests.

khionu:
It's been a while since I used 0.9, but I recall being able to use these with no issue. Given there is no open issue for this for 0.9, I don't think this should be accepted. We don't know it's broken, so let's leave it as is

RogueException:
I'm avoiding touching the 0.9 branch unless absolutely necessary, and I don't remember this ever being an issue o.O.

khionu:
Agreed. This wasn't an issue for 0.9. Cited Issue is for 1.0

PR #344, opened 2016-11-02
Fix EnumTypeReader missing equivalent Enum values by name.

AntiTcb:
Enumerating over the Enum values would prevent declarations such as this: `enum Foo {A, B, AliasForA = A}` from being parsed properly as AliasForA would not be contained within _enumsByName. By all means, if there's a cleaner implementation rather than having to check if _enumsByValue already contains the key, spruce up my changes.

Also opted to have the error message say exactly which parse failed.

### DIFF
--- a/src/Discord.Net.Commands/Readers/EnumTypeReader.cs
+++ b/src/Discord.Net.Commands/Readers/EnumTypeReader.cs
@@ -32,10 +32,12 @@ namespace Discord.Commands
             var byNameBuilder = ImmutableDictionary.CreateBuilder<string, object>();
             var byValueBuilder = ImmutableDictionary.CreateBuilder<T, object>();
             
-            foreach (var v in Enum.GetValues(_enumType))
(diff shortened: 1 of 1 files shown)


src/Discord.Net.Commands/Readers/EnumTypeReader.cs
```diff
@@ -59,7 +61,7 @@ public override Task<TypeReaderResult> Read(CommandContext context, string input
                 if (_enumsByName.TryGetValue(input.ToLower(), out enumValue))
                     return Task.FromResult(TypeReaderResult.FromSuccess(enumValue));
                 else
-                    return Task.FromResult(TypeReaderResult.FromError(CommandError.ParseFailed, $"Value is not a {_enumType.Name}"));
+                    return Task.FromResult(TypeReaderResult.FromError(CommandError.ParseFailed, $"Name is not a {_enumType.Name}"));
```
RogueException: I'd leave this with the old error. It's talking about the value provided by the user.

PR #346, opened 2016-11-07
Make header dictionary ignore case

ObsidianMinor:
Sometimes "Date" is "date". This fixes it.


RogueException:
Thanks for the PR~

PR #353, opened 2016-11-13
Support rich Embeds

420foxbot:
This PR adds support for Rich Embeds.

**This can be a breaking change!**

The `IMessageChannel.SendMessage` method has been modified to include an optional `API.Embed` parameter, which will be sent to Discord with the message. You can produce an `API.Embed` object through the `EmbedBuilder` utility, or by creating it yourself.

This feature has been tested.

Since, with this update, messages no longer need to have any content to be valid, I have removed any preconditions that ensure messages have content. However, **sending messages with no content will crash the ratelimiter.** Still looking into a cause.

### DIFF
--- a/src/Discord.Net.Commands/ModuleBase.cs
+++ b/src/Discord.Net.Commands/ModuleBase.cs
@@ -6,9 +6,9 @@ namespace Discord.Commands
     {
         public CommandContext Context { get; internal set; }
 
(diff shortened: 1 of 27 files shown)


FiniteReality:
Personally, I think that using types from the API namespace in public code is a bad idea. I'd prefer making `EmbedBuilder.Build` (and friends) internal and simply passing the EmbedBuilder to the `SendMessage` method.

Other than that, the EmbedBuilder class looks fine to me.

PR #356, opened 2016-11-17
Unbreaking enum reader

aequasi:


aequasi:
Oh yeah. didnt even see that one

PR #358, opened 2016-11-17
Add Channel to MessageDeleted event

LassieME:
Adds channel to the MessageDeleted event. Getting the channel before required the message to be in cache, which it not always is.

### DIFF
--- a/src/Discord.Net.WebSocket/DiscordSocketClient.Events.cs
+++ b/src/Discord.Net.WebSocket/DiscordSocketClient.Events.cs
@@ -59,12 +59,12 @@ namespace Discord.WebSocket
             remove { _messageReceivedEvent.Remove(value); }
         }
         private readonly AsyncEvent<Func<SocketMessage, Task>> _messageReceivedEvent = new AsyncEvent<Func<SocketMessage, Task>>();
-        public event Func<ulong, Optional<SocketMessage>, Task> MessageDeleted
+        public event Func<ulong, SocketChannel, Optional<SocketMessage>, Task> MessageDeleted
         {
             add { _messageDeletedEvent.Add(value); }
             remove { _messageDeletedEvent.Remove(value); }
         }
-        private readonly AsyncEvent<Func<ulong, Optional<SocketMessage>, Task>> _messageDeletedEvent = new AsyncEvent<Func<ulong, Optional<SocketMessage>, Task>>();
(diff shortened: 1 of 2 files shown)


src/Discord.Net.WebSocket/DiscordSocketClient.Events.cs
```diff
@@ -59,12 +59,12 @@ public partial class DiscordSocketClient
             remove { _messageReceivedEvent.Remove(value); }
         }
         private readonly AsyncEvent<Func<SocketMessage, Task>> _messageReceivedEvent = new AsyncEvent<Func<SocketMessage, Task>>();
-        public event Func<ulong, Optional<SocketMessage>, Task> MessageDeleted
+        public event Func<ulong, SocketChannel, Optional<SocketMessage>, Task> MessageDeleted
```
Nikey646: Shouldn't this be `ISocketMessageChannel` instead of `SocketChannel`? Other events that receive a channel type object receive the interface not an object.

foxbot (approved):
Not sure how I feel about the arrangement of parameters in the event signature, but this should be an okay change.

src/Discord.Net.WebSocket/DiscordSocketClient.Events.cs
LassieME: As far as I can see only the UserIsTyping event uses ISocketMessageChannel. I'd not be opposed to that change tho.

Nikey646: MessageCreated and MessageUpdated receive `SocketMessage` objects which have a `ISocketMessageChannel Channel` property. I believe that is the reason why only UserIsTyping receives the `ISocketMessageChannel` object directly?

I mean it doesn't really matter because you can cast it into any of the required objects, or even into the interface directly. Just seemed more appropriate?

RogueException: Yeah, this should be ISocketMessageChannel

Auralytical (changes requested):

RogueException:
@420foxbot maybe we should look into a special type just for messages that combines Id, HasValue and Value. I think these events are the only places we publicly expose Optional<> other than API types, and it feels a bit weird to have them here.

LassieME:
Someone put a On Hold label on this while #359 is figured out.

Auralytical (approved):

LassieME:
Was combined into #395 . Closing this.

Issue #238, opened 2016-08-27
Update documentation for IUserMessage/ISystemMessage


420foxbot:
Depends on #237

420foxbot:
Resolved in 8406bd8f73de926630a921cac2bee451198af0e4

Issue #304, opened 2016-10-10
Should Command Groups be separated by a user-defined character, rather than whitespace?


foxbot:
In 1.0's current implementation of Commands, Modules. Groups, and Commands are separated from eachother by whitespace. For someone's general usecase, this would not be an issue.

Consider the following scenario:

```
Module "module"
    Group "group"
        Command "command"
            Remainder String "args"
    Command "group"
        Remainder String "args"
```

If a user executes `~module group command`, which command should the bot execute? In the current implementation, by default, the bot will execute the first match it finds, which would be `module -> group`. However, what if the user wanted to execute `module -> group -> command`?

With this proposal, the bot developer will be able to specify a string that should be expected between tokens when searching for a command. For example, the user could now discretely see a difference between:
`~module.group.command args`
and
`~module.group command`

Thoughts?

khionu:
I'm all for it. Could make for some very creative uses.

DjRand:
IMO, this seems like a useful feature to have.   Can't say for sure if I'd use it, but it's something I'd keep in mind if it was available to be used, as it definitely offers a way to add a little more clarity to your command style.

khionu:
Clarity or artistic. For instance, this could be used to imitate a filesystem, if you use slashes.

420foxbot:
this could be used to imitate a **programming language** (hell yeah)
~please.kill.me()

FiniteReality:
This definitely seems like a good option to include.

RogueException:
Looks good

Issue #192, opened 2016-08-12
Add command builders


RogueException:
The existing system should be migrated over to using this. ModuleBuilder should be injected into DI so a module can do post-magic alterations before data becomes immutable.

PR #361, opened 2016-11-18
Implement command builders

FiniteReality:
This system should allow users to build commands manually using a fluent interface similar to 0.9, or using the current type-based system.

This hasn't been tested extensively, as I'm not sure of all the use cases I need to test against. I have, however, tested against some common cases (see [this](https://gist.github.com/FiniteReality/935b22cd18901faa68e9bf8d24354402)) which I think should cover most cases.


FiniteReality:
This should resolve #238, #263 and #192 as well, and after this #304 should be fairly easy to implement.

PR #364, opened 2016-11-19
Parameter preconditions and typereader overriding

FiniteReality:
Using the new command builder system, since I'm not sure whether Joe4evr is going to update his system.

Parameter preconditions had to be done as part of the execute phase as they require the value passed to the command after parsing - I wasn't sure about the best way to include it as a separate phase as it would have meant changing how the parameter list was built and passed around heavily.

### DIFF
--- a/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
@@ -183,7 +183,9 @@ namespace Discord.Commands
                 if (attribute is SummaryAttribute)
                     builder.Summary = (attribute as SummaryAttribute).Text;
                 else if (attribute is OverrideTypeReaderAttribute)
-                    builder.TypeReader = service.GetTypeReader((attribute as OverrideTypeReaderAttribute).TypeReader);
+                {
+                    builder.TypeReader = GetTypeReader(service, paramType, (attribute as OverrideTypeReaderAttribute).TypeReader);
+                }
                 else if (attribute is ParameterPreconditionAttribute)
                     builder.AddPrecondition(attribute as ParameterPreconditionAttribute);
                 else if (attribute is ParamArrayAttribute)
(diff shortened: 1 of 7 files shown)


FiniteReality:
So, I'm a derp and didn't test properly; the latest commit fixes the mistakes I made.

src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
```diff
@@ -193,23 +199,47 @@ private static void BuildParameter(ParameterBuilder builder, System.Reflection.P
...
             }
 
-            var reader = service.GetTypeReader(paramType);
-            if (reader == null)
+            if (builder.TypeReader == null)
             {
-                var paramTypeInfo = paramType.GetTypeInfo();
-                if (paramTypeInfo.IsEnum)
+                var readers = service.GetTypeReaders(paramType);
+                var reader = readers?.FirstOrDefault();
```
RogueException: Is it safe to always assume the first reader is the default? Maybe we should have a separate field for the default/global reader?

RogueException: It seems you're using ConcurrentBags which, by definition, do not guarantee any ordering.

src/Discord.Net.Commands/Info/CommandInfo.cs
```diff
@@ -125,9 +125,26 @@ public Task<ExecuteResult> Execute(CommandContext context, ParseResult parseResu
             if (map == null)
                 map = DependencyMap.Empty;
 
+            object[] args = null;
+
+            try
```
RogueException: It's safe to leave this try-catch around the whole op, instead of splitting it into two sections.

Auralytical (changes requested):

Auralytical (approved):

RogueException:
If this has been retested, PR is approved once conflicts are resolved.

FiniteReality:
@RogueException Any news on merging this? As far as I'm aware this is ready

Issue #365, opened 2016-11-19
Nameless commands triggering NullReferenceException on AddModulesAsync


james7132:
Setup:
```csharp
[Group("test")]
public class TestModule : ModuleBase {
  [Command]
  public async Task TestCommand() {
    ...
  }
}
```
Result of using ``CommandService.AddModulesAsync``:
```
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at Discord.Commands.CommandMapNode.NextWhitespace(String text, Int32 startIndex)
   at Discord.Commands.CommandMapNode.AddCommand(String text, Int32 index, CommandInfo command)
   at Discord.Commands.CommandMap.AddCommand(CommandInfo command)
   at Discord.Commands.CommandService.LoadModuleInternal(ModuleInfo module)
   at Discord.Commands.CommandService.LoadModuleInternal(ModuleInfo module)
   at Discord.Commands.CommandService.<AddModulesAsync>d__12.MoveNext()
```

FiniteReality:
Mind running a debugger and seeing what the command's aliases are? This may be a result of my command builders code or volt's reworked command map - if we know what the command's aliases are it'll be easier to narrow down.

james7132:
The command only has one element under alias, and it is indeed null from the null initialization in CommandAttribute.

FiniteReality:
Okay, that means it's a problem on my side. I'll fix it.

james7132:
Fixed as of  #367

PR #367, opened 2016-11-20
Fix a bunch of command/alias issues

### LINKED ISSUE #366: Submodules ModuleInfo.Aliases collection does not contain Group name
```cs
[Group("primary")]
public class PrimaryModule : ModuleBase { 
    [Group("sub")]
    public class SubModule : ModuleBase {}
}
```
"primary" will be found in PrimaryModule's aliases, while "sub" will not be in SubModule's aliases (collection will be empty in this case). The behavior only happens to submodules.

FiniteReality:
According to my testing, resolves #366, #365 as well as CommandService.Commands/Modules not returning all registered commands/modules respectively.

### DIFF
--- a/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
@@ -118,7 +118,6 @@ namespace Discord.Commands
 
             builder.Name = method.Name;
 
-            var setName = false;
             foreach (var attribute in attributes)
             {
                 // TODO: C#7 type switch
@@ -127,12 +126,11 @@ namespace Discord.Commands
                     var cmdAttr = attribute as CommandAttribute;
                     builder.AddAliases(cmdAttr.Text);
                     builder.RunMode = cmdAttr.RunMode;
-                    builder.Name = setName ? builder.Name ?? cmdAttr.Text : cmdAttr.Text ?? builder.Name;
+                    builder.Name = builder.Name ?? cmdAttr.Text;
                 }
                 else if (attribute is NameAttribute)
                 {
                     builder.Name = (attribute as NameAttribute).Text;
-                    setName = true;
                 }
                 else if (attribute is PriorityAttribute)
                     builder.Priority = (attribute as PriorityAttribute).Priority;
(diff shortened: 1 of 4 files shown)


src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
```diff
@@ -127,7 +129,9 @@ private static void BuildCommand(CommandBuilder builder, TypeInfo typeInfo, Meth
                     builder.Name = builder.Name ?? cmdAttr.Text;
                 }
                 else if (attribute is NameAttribute)
+                {
```
RogueException: Style consistency: remove braces

src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
```diff
@@ -140,6 +144,9 @@ private static void BuildCommand(CommandBuilder builder, TypeInfo typeInfo, Meth
                     builder.AddPrecondition(attribute as PreconditionAttribute);
             }
 
+            if (builder.Name == null)
+                builder.Name = method.Name;
```
RogueException: We already have this on line 119

Auralytical (changes requested):

james7132:
Just tested this on my current setup, and it does not resolve #365: there is still a null alias included. Will report back why as soon as possible.

james7132:
The following code still triggers a NullReferenceException as described in #365:

```csharp
public class Admin : ModuleBase {

  [Group("prune")]
  public class Prune : ModuleBase {

    [Command]
    public Task Default(int count) {
      ...
    }

  }

}
```

The builder has a single null alias and the module has no aliases in this case.

FiniteReality:
@james7132 Interesting; it appears to be fixed for me:
screenshot_20161125_204315

Mind showing your code?

james7132:
For the exact code, check [here](https://github.com/james7132/Hourai/blob/master/src/Modules/Admin/Prune.cs#L18).

james7132:
Yes I am currently on commit 05fb81c.

FiniteReality:
@foxbot I can reproduce the issue; my Permutate code is returning an empty result which is incorrect

FiniteReality:
Okay, it should be fixed now. I'll work on fixing the reviewed issues now.

james7132:
Can confirm that this resolves  #365 now.

Auralytical (approved):

PR #369, opened 2016-11-21
Add relationship features

### LINKED ISSUE #64: Add Friends List Support
An important feature we're missing on the client API side of things. Should get this is in for 1.1.

ObsidianMinor:
Opening this pull request for when work starts on 1.1

Adds methods to get relationships, add users as friends, block users, and remove relationship for a specified user.

When pulled, this will close #64


src/Discord.Net.Core/API/DiscordRestApiClient.cs
```diff
@@ -972,6 +972,34 @@ internal virtual void Dispose(bool disposing)
...
             catch (HttpException ex) when (ex.StatusCode == HttpStatusCode.NotFound) { return null; }
         }
+        
+        //Relationships
+        public async Task<IReadOnlyCollection<Relationship>> GetRelationshipsAsync(RequestOptions options = null)
+        {
+            options = RequestOptions.CreateOrClone(options);
+            return await SendAsync<IReadOnlyCollection<Relationship>>("GET", () => "users/@me/relationships", new BucketIds(), options: options).ConfigureAwait(false);
+        }
+        public async Task AddFriendAsync(ulong userId, RequestOptions options = null)
```
ObsidianMinor: If the current relationship is IncomingRequest, this method accepts the request. If there is no relationship, it will send a friend request

ObsidianMinor:
When pulled, this will close #64

src/Discord.Net.WebSocket/ClientState.cs
```diff
@@ -10,18 +10,21 @@ internal class ClientState
         private const double AverageChannelsPerGuild = 10.22; //Source: Googie2149
         private const double AverageUsersPerGuild = 47.78; //Source: Googie2149
         private const double CollectionMultiplier = 1.05; //Add 5% buffer to handle growth
+        private const double AverageRelationshipsPerUser = 30;
```
khionu: Did you get data for this?

khionu (changes requested):
You alternate between the noun "relations" and "relationships". "Relations" is not context-correct, I would recommend refactoring to use "relationships".

That, and the AverageRelationshipsPerUser is an optimizer, it should be a statistically-accurate number. 30 may sound reasonable (it does to me, too), but should get more data on that, if you don't have it already.

Sentinent:
You should also initially populate relationships from the READY event.

binn:
You know @ObsidianMinor until they merge this pull request you can always make it a separate class or something so people like us who can't wait can use it :P if you can please do!! and also if you could make a version for 0.9.x it'd be amazing!

ObsidianMinor:
@Sentinent [I already do](https://github.com/RogueException/Discord.Net/pull/369/files#diff-714cba74b189aefe65e5bd00bde121a2R491)

PR #373, opened 2016-11-22
Added ColorPresets struct, moved Color.Default to ColorPresets.

AntiTcb:
https://github.com/RogueException/Discord.Net/issues/227

Color are left-to-right, top then bottom row when comparing the order from the modal. Very iffy on names.

### DIFF
--- a/src/Discord.Net.Core/Entities/Roles/Color.cs
+++ b/src/Discord.Net.Core/Entities/Roles/Color.cs
@@ -5,10 +5,7 @@ namespace Discord
 {
     [DebuggerDisplay(@"{DebuggerDisplay,nq}")]
     public struct Color
-    {
-        /// <summary> Gets the default user color value. </summary>
-        public static readonly Color Default = new Color(0);
-
+    { 
         /// <summary> Gets the encoded value for this color. </summary>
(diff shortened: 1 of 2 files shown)


foxbot (approved):
Not sure where you sourced the color names for these from, but everything else looks fine.

khionu:
This idea was already discussed months ago, and rejected.

khionu:
Whoops.... w/e, the discussion I recall was in the Design Guild, but before that issue.

RogueException:
I wanted to bring in the standard set of .NET color names, or just use the drawing.color primitive

FiniteReality:
To be fair, System.Drawing.Primitives can be used on any system to gain access to the System.Drawing.Color type and its presets. Maybe we should include a method to create a Discord.Net color from a System.Drawing.Color?

RogueException:
Do we even need a custom color type if we're importing drawing primitives?

FiniteReality:
I've just looked around, and it seems that System.Drawing.Primitives is a
netstandard2.0 project so we can't import it for Color primitives.

AntiTcb:
Redone in https://github.com/RogueException/Discord.Net/pull/725

PR #375, opened 2016-11-25
Adds Timestamps to embeds, and removes Image and Thumbnail-Builders

LassieME:
Not sure about the ImageUrl and ThumbnailUrl implementation, but this works.

### DIFF
--- a/src/Discord.Net.Core/Entities/Messages/EmbedBuilder.cs
+++ b/src/Discord.Net.Core/Entities/Messages/EmbedBuilder.cs
@@ -20,12 +20,13 @@ namespace Discord
             _fields = new List<Field>();
         }
 
-        public string Title { get { return _model.Title; }  set { _model.Title = value; } }
+        public string Title { get { return _model.Title; } set { _model.Title = value; } }
         public string Description { get { return _model.Description; } set { _model.Description = value; } }
         public string Url { get { return _model.Url; } set { _model.Url = value; } }
         public string ThumbnailUrl { get; set; }
         public string ImageUrl { get; set; }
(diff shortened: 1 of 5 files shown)


foxbot (approved):

src/Discord.Net.Core/Entities/Messages/EmbedBuilder.cs
```diff
@@ -44,6 +45,26 @@ public EmbedBuilder WithUrl(string url)
...
+        {
+            ThumbnailUrl = thumbnailUrl;
+            return this;
+        }
+        public EmbedBuilder WithImageUrl(string imageUrl)
+        {
+            ImageUrl = ImageUrl;
+            return this;
+        }
+        public EmbedBuilder WithTimestamp()
```
RogueException: This is unclear. should be more obvious that it's setting Timestamp to Now.
`WithNowTimestamp()` maybe? Unless someone has a better name.

Auralytical (changes requested):

src/Discord.Net.Core/Entities/Messages/EmbedBuilder.cs
foxbot: `WithCurrentTimestamp()`

LassieME: Sounds good to me.

RogueException: That works

Auralytical (approved):

Issue #211, opened 2016-08-19
RequirePermissions precondition for the bot itself


Kwoth:
RequirePermissions right now has guild and channel permissions which are checked on the user running the command.

For commands like ban,prune, channel manipulation... it would be nice to have preconditions which are compared against the **CurrentUser** (the bot itself) to avoid 403s and other failures in the command's code.

FiniteReality:
This is definitely something we should include, but I'm going to say it needs to be put on hold until we figure out a nice way to pass around information like the bot's user to the precondition.

A current workaround for you would be to use the 'moduleInstance' parameter in your own precondition to pass around the client instance and use that to check your permissions.

foxbot:
Preconditions now have a CommandContext passed into them, this should be an easy addition.

foxbot:
Fixed as of ea148db08b2c2fa643167a4d0703113ee04a44ab

PR #376, opened 2016-11-25
Improve preconditions

foxbot:
**This is a breaking change.**

This PR implements a RequireOwner precondition, which will require that the user invoking a command is the owner of the bot.

A RequireBotPermission precondition is introduced (per #211), which will require that the bot has a specified permission in the context of the command. As previously discussed, RequirePermission was renamed to RequireUserPermission.

All preconditions now include docstrings.

### DIFF
--- a/src/Discord.Net.Commands/Attributes/Preconditions/RequireOwnerAttribute.cs
+++ b/src/Discord.Net.Commands/Attributes/Preconditions/RequireOwnerAttribute.cs
@@ -13,12 +13,9 @@ namespace Discord.Commands
     [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
     public class RequireOwnerAttribute : PreconditionAttribute
     {
-        private IApplication application;
-
         public override async Task<PreconditionResult> CheckPermissions(CommandContext context, CommandInfo command, IDependencyMap map)
         {
-            if (application == null)
-                application = await context.Client.GetApplicationInfoAsync();
+            var application = await context.Client.GetApplicationInfoAsync();
             if (context.User.Id == application.Owner.Id) return PreconditionResult.FromSuccess();
             return PreconditionResult.FromError("Command can only be run by the owner of the bot");
         }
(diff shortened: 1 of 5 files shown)


src/Discord.Net.Commands/Attributes/Preconditions/RequireOwnerAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// <remarks>This precondition will only work if the bot is a bot account.</remarks>
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
+    public class RequireOwnerAttribute : PreconditionAttribute
+    {
+        private IApplication application;
+
+        public override async Task<PreconditionResult> CheckPermissions(CommandContext context, CommandInfo command, IDependencyMap map)
+        {
+            if (application == null)
+                application = await context.Client.GetApplicationInfoAsync();
```
RogueException: Ratelimit risk, we should make sure we cache this info in DiscordSocketClient.

Auralytical (changes requested):

src/Discord.Net.Commands/Attributes/Preconditions/RequireOwnerAttribute.cs
foxbot: @RogueException are preconditions created every time a command is invoked? I was under the impression that they were only created once, which was why I already cache this.

FiniteReality: The precondition instance is cached in the CommandInfo/ModuleInfo type - so this should not be an issue as long as the command isn't loaded/unloaded multiple times.

However, I think that caching ApplicationInfo on the DiscordSocketClient would be a good idea for the lib overall - I would use it in my preconditions to prevent ratelimits from occuring.

RogueException: I didn't notice the caching here already... whoops. But yeah, never cache at the precondition level since it won't be updated by logouts.

PR #377, opened 2016-11-25
Add CommandServiceConfig, DefaultRunMode

### LINKED ISSUE #368: CommandService should expose a DefaultRunMode option
Currently, commands are set to 'RunMode.Sync' by default. This should work fine for most bots, but for any bots receiving high-traffic, the 200-300ms required for a single reply will cause a noticeable lag and backup for the bot.

By changing the default RunMode to Mixed, the light work in a command (e.g. string manipulation) will still be left on the gateway thread, but the bot will not wait for callbacks from the Discord API on the gateway thread.

foxbot:
This adds an (optional) CommandServiceConfig, as well as a DefaultRunMode for commands.

This resolves #368 (for commands where a RunMode is not explicitly specified, a custom default value should be used)

### DIFF
--- a/src/Discord.Net.Commands/Attributes/CommandAttribute.cs
+++ b/src/Discord.Net.Commands/Attributes/CommandAttribute.cs
@@ -6,7 +6,7 @@ namespace Discord.Commands
     public class CommandAttribute : Attribute
     {
         public string Text { get; }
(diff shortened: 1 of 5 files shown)


AntiTcb:
Also sounds like a good place to stick MultiMatchHandling in as well.

PR #382, opened 2016-11-26
Allow command nodes to be separated by a configurable string

### LINKED ISSUE #304: Should Command Groups be separated by a user-defined character, rather than whitespace?

foxbot:
Resolves #304

### DIFF
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -31,7 +31,6 @@ namespace Discord.Commands
             _moduleLock = new SemaphoreSlim(1, 1);
             _typedModuleDefs = new ConcurrentDictionary<Type, ModuleInfo>();
             _moduleDefs = new ConcurrentBag<ModuleInfo>();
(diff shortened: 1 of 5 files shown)


FiniteReality:
Did you change the command map? More specifically, `CommandMapNode.NextWhitespace` looks for spaces to build commands.

PR #386, opened 2016-11-27
Resolve #384: Add upward reference for submodules

### LINKED ISSUE #384: Upward reference for submodules
Right now ModuleInfo currently has a `Submodules` property that provides access to modules nested within other modules, but there is no way for a obtain the parent module from the submodule. This is an issue if trying to obtain the full hierarchy from a given CommandInfo. Additionally, with `CommandService.Modules` enumerating all modules, whether they're a top level module or not, there is no way to tell if a module is a submodule. 

A solution that I propose:
 * Provide a `Parent` property on ModuleInfo, referring to the parent module, left null for top-level modules.
* Limit `CommandService.Modules` to only top level modules.

james7132:
This currently has the proposed solution mentioned in #384.

### DIFF
--- a/src/Discord.Net.Commands/Builders/ModuleBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/ModuleBuilder.cs
@@ -14,8 +14,7 @@ namespace Discord.Commands.Builders
         public CommandService Service { get; }
         public ModuleBuilder Parent { get; }
         public string Name { get; set; }
-        public string Summary { get; set; }
-        public string Remarks { get; set; }
+        public string Summary { get; set; } public string Remarks { get; set; }
 
         public IReadOnlyList<CommandBuilder> Commands => _commands;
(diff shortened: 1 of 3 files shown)


RogueException:
@FiniteReality you've been championing commands lately, mind taking a look at this?

FiniteReality:
Oh hey, I forgot about email replies. I'm fine with these changes; though I
think it would be nice if there were still a way to access all modules as
we do have the data internally. Maybe we should split it into
`CommandService.AllModules` and `CommandService.TopLevelModules`?

james7132:
It may be better and more succinct to revert CommandService.Modules to what it was before, and let devs apply the filter as needed.

FiniteReality:
Yeah, that would probably be the best approach. Other than that, LGTM

james7132:
Done. PTAL.

FiniteReality:
Nothing else needs to be done here, right? This looks ready to merge @RogueException

PR #387, opened 2016-11-27
Add TimeSpan TypeReader

james7132:
Added a TimeSpan TypeReader. Useful for taking in a measure of time for a command (i.e. a temporary ban).

### DIFF
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -14,7 +14,7 @@ namespace Discord.Commands
     public class CommandService
     {
         private readonly SemaphoreSlim _moduleLock;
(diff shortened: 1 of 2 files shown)


FiniteReality:
I think we decided against adding typereaders for things like this as bot devs may want to do it in different ways.
For example, your typereader parses a timespan more like a date/time - the time span reader I have reads more like `10h5m3s` and the like.

khionu:
Custom TypeReaders overwrite the Defaults, so this wouldn't interfere.... I think the expansion of what defaults are provided was under consideration, though

RogueException:
Anything that has a default .NET TryParse method is allowed. Devs can override defaults.
LGTM.

Issue #268, opened 2016-09-04
Option for commands to be case insensitive


Sentinent:
Right now, commands are case sensitive as defined in their attributes. There should be a field or config for CommandService to make commands case insensitive.

khionu:
We'll discuss this. Reminder, 1.0 is still in Beta, things might, and probably will, change

Joe4evr:
Should commands be case-insensitive by default? That's how they are in 0.9.

khionu:
1.0 Commands was largely about putting more control in the bot dev's hands. I think maybe a boolean in the Search function parameters would be a more wise move.

foxbot:
Resolved in 1406943bf9a5bd18779f5c509dc02834a81d3105

PR #391, opened 2016-11-28
Resolves #390

### LINKED ISSUE #390: CaseSensitiveCommands lowercasing arguments bug
When using the CommandService with config parameter `CaseSensitiveCommands = true`, all parameters of all commands are lowercased as a side effect.

Aux:
Fix case insensitive commands forcing parameters to return lowercase in SearchResult as specified in #390


Nikey646:
This should also solve the base problem of #268 right? The option to toggle case sensitivity already exists which should mean that it can close both issues.

PR #393, opened 2016-12-01
EmbedBuilder.WithImageUrl() now uses the value passed

Nikey646:
This is arguably the silliest Pull Request. But better than just whining about it.


LassieME:
Nope, totally valid.

Nikey646:
It is currently `this.ImageUrl = this.ImageUrl`... It is setting the value to itself.

LassieME:
Not entirely how this slipped past me tbh. I must have done something funky with got.

PR #395, opened 2016-12-03
Resolves #359: Combines event arguments into one struct.

### LINKED ISSUE #359: Gateway Events should be raised with a single object, rather than a (ulong, Optional<T>)
For any events that involve pulling an object from the cache, we raise the event with a signature of `(ulong, Optional<T>)`, where `ulong` represents the ID of the object, and the optional encapsulates the cached object, if it exists; otherwise an empty optional is created.

We should instead create a single object, which contains all relevant information known about the entity (Entity ID, Entity Channel ID), as well as a cached version of the object, if it exists. For entities where we have enough information available to download the object, we should also provide a download method, directly on the container.

Potential Design
```
public abstract class Some<T> where T : IEntity<TId>
{
    public readonly TId Id;
    public bool IsCached => Value != null;
    public T Value;
    public async Task<T> Download();
    public async Task<T> GetOrDownload();
}
```

Implementation
```
_client.ReactionAdded += async (someMessage, reaction) =>
{
    var message = await someMessage.GetOrDownload();
    if (reaction.Emoji.Name != "[unicode for star emoji]") return;
    await message.Channel.SendMessageAsync($"{message.Author}'s message got a :star:!";
};
```

This concept still needs another name, I believe there is already a primitive named `Some`.

LassieME:
Makes #358 obsolete and should resolve #359.

Honestly not sure I like this solution using only one struct, having a check for IsDownloadable and throwing an exception for MessageDeleted seems a bit, idk, janky. 

Might make more sense to either split it up into multiple structs or take the GC penalty and make it an abstract class. That would also make it applicable to use in the PresenceUpdate event, which with this current setup is the only event that still uses Optional. Looking forward to reading your thoughts on this one.

Still left to do:
- [x] Docstrings for events and cached struct.


FiniteReality:
I think we should make the restriction for `Cached<T>` simply be `IEntity` - that way it works for all entity types without changes in the future. Also, I don't think we need the `IsDownloadable` flag; entities which use the `Cached<T>` type should always be able to be downloaded.

LassieME:
So your suggestion is to keep the Optional on the MessageDeleted event? I guess that's OK, but it sounded like Voltana didn't like exposing Optional outside of APIClient.

FiniteReality:
Huh, good point - I forgot that message delete events don't always have the message cached.

Also, there's one issue with the `Cached<T>` type: what happens when the entity id is no longer a ulong? `Entity<TId>` supports multiple values for entity ids, so we should probably change the definition to be something like this:

```cs
public struct Cached<TEntity> where TEntity : Entity<TId>
```

That way we have `TEntity` for the entity type, and `TId` for the id type.

Furthermore, I think it would be useful to have some static helper methods to create `Cached<T>` types from an entity, and an entity id.

LassieME:
I assume you meant `IEntity<TId>`?

LassieME:
How would that work with the message deleted event? @foxbot I'm kinda blanking on how that would even work. If you could expand on your idea a little that would be great.

LassieME:
I think this is ready for review.

LassieME:
Sounds good, invite sent.

src/Discord.Net.Core/Utils/Cacheable.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// <typeparam name="TEntity">The type of entity that is cached</typeparam>
+    /// <typeparam name="TId">The type of this entity's ID</typeparam>
+    public struct Cacheable<TEntity, TId>
+        where TEntity : IEntity<TId>
+        where TId : IEquatable<TId>
+    {
+        /// <summary>
+        /// Is this entity cached?
+        /// </summary>
+        public bool HasValue => !EqualityComparer<TEntity>.Default.Equals(Value, default(TEntity));
```
Auralytical: Integer == 0 does not mean it has no value. This should be a bool passed to the constructor

Auralytical (changes requested):

Auralytical:
Sorry this took so long, had to figure out how this lined up with other event changes. After the one fix, LGTM.

LassieME:
Fixed.

Auralytical (approved):

foxbot (approved):

src/Discord.Net.Core/Utils/Cacheable.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// <typeparam name="TEntity">The type of entity that is cached</typeparam>
+    /// <typeparam name="TId">The type of this entity's ID</typeparam>
+    public struct Cacheable<TEntity, TId>
+        where TEntity : IEntity<TId>
+        where TId : IEquatable<TId>
+    {
+        /// <summary>
+        /// Is this entity cached?
+        /// </summary>
+        public bool HasValue { get; }
```
AntiTcb, line 19: Shouldn't this be `public bool HasValue => Value != null;`?

Auralytical: No, because there's no class constraint. We can't be sure we wont use this helper in other places with value type entities, so best to be safe.

AntiTcb: Wouldn't the value types be wrapped in a nullable anyway? Probably making a bigger deal out of this than necessary, but it seems worrying for the potential of HasValue to be false if Value actually exists.

Auralytical: Nullable still isnt a reference type. I'll also be doing a cleanup of this class sometime soon.

PR #412, opened 2016-12-11
Implement configurable command node separators

### LINKED ISSUE #304: Should Command Groups be separated by a user-defined character, rather than whitespace?

FiniteReality:
This fixes #304 over foxbot's PR for the same issue.

### DIFF
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -75,7 +75,6 @@ namespace Discord.Commands
             };
             _caseSensitive = config.CaseSensitiveCommands;
             _defaultRunMode = config.DefaultRunMode;
-
             _splitCharacter = config.CommandSplitCharacter;
         }
 
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -22,6 +22,7 @@ namespace Discord.Commands
 
         internal readonly bool _caseSensitive;
         internal readonly RunMode _defaultRunMode;
+        internal readonly char _splitCharacter;
 
         public IEnumerable<ModuleInfo> Modules => _moduleDefs.Select(x => x);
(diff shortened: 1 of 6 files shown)


src/Discord.Net.Commands/CommandService.cs
```diff
@@ -74,6 +75,8 @@ public CommandService(CommandServiceConfig config)
             };
             _caseSensitive = config.CaseSensitiveCommands;
             _defaultRunMode = config.DefaultRunMode;
+
```
foxbot: Why is there a blank line here?

foxbot (approved):
LGTM

src/Discord.Net.Commands/CommandService.cs
FiniteReality: That shouldn't be the case, whoops.

AeonLucid:
@RogueException @foxbot  When are bleeding edge MyGet builds published? Need to have this fix for my commands. 😄

AeonLucid:
@foxbot I don't think it pushed this commit, either that or I still have this issue. 

Version: 1.0.0-beta2-00512
Code: https://github.com/keyphact/ShiroBot/blob/4d9bd5d70bcb599b1d33c95cac7ce6e3fe436fcc/src/ShiroBot/Commands/InfoModule.cs

PR #417, opened 2016-12-14
Enhance Dependency Injection

foxbot:
The DependencyMap is modified to now accept:
- Transient objects (objects that will be instantiated each time they are loaded from the map)
- Factories (custom logic for injecting a type)

You may now inject types to your module by flagging a **settable property** with the InjectAttribute.


Auralytical:
I don't like the use of reflection here. All the fields/properties that are going to be set should be compiled into a reusable Func.

Auralytical:
Also docstrings would be good - I've never actually known what Transient deps were until you mentioned it.

PR #423, opened 2016-12-18
Allow module builders to have empty aliases

FiniteReality:
Fixes an issue foxbot found on discord

An empty alias is an empty string or whitespace: previously it didn't pick these things up which would cause commands to build incorrectly.

This commit should also fix an issue where commands would not get removed correctly from the command map when unloading modules.


Auralytical:
Went with https://github.com/RogueException/Discord.Net/commit/c1a9bdc142bc216490b5c00cb87c8b973f4ebaeb instead

PR #430, opened 2016-12-19
Fixed UserTypeReader from throwing NullReferenceException

Confruggy:
When using the Name#1234 format with a user that doesn't exist, the UserTypeReader throws a NullReferenceException. This fix it by using a null conditional operator.


foxbot (approved):
LGTM

Issue #427, opened 2016-12-19
Modules without aliases causes CommandInfo to lose alias information


AntiTcb:
```cs
Aliases = module.Aliases
                .Permutate(builder.Aliases, (first, second) =>
                {
                    if (first == "")
                        return second;
                    else if (second == "")
                        return first;
                    else
                        return first + service._separatorChar + second;
                })
                .Select(x => service._caseSensitive ? x : x.ToLowerInvariant())
                .ToImmutableArray();
```

This line in `CommandInfo`'s ctor will cause `Aliases` to remain as an empty collection if the module doesn't have any aliases. Normal functionality for nameless/prefixless modules is restored if module.Aliases contains an empty string, but, should every module have that alias as a default?

PR #431, opened 2016-12-19
Issues/427 + AliasAttribute targets classes

AntiTcb:
#427 
Tested, aliasless & aliased modules work fine.


foxbot (approved):
Tested; LGTM.

PR #432, opened 2016-12-21
DefaultRetryMode never assigned in DiscordRestAPIClient

BoGuu:
As the title states, DefaultRetryMode is never being assigned, thus defaults to AlwaysFail. Causes exceptions such as RateLimited to always be thrown. This also doesn't occur if you pass your own RequestOptions.

Commit where the problem was introduced: 1efcd3daf699b99330862364006ec1269f534dc7


Auralytical:
Good catch, thanks~

PR #440, opened 2016-12-24
Add command suggestions for incorrect commands

FiniteReality:
Allows bot devs to get a list of similar commands from the command map when users try to execute an invalid command.

Example Usage:
```cs
var result = await service.ExecuteAsync(context, 0, maxDifferences: 3);
if (result is SearchResult)
{
    var sResult = (SearchResult)result;
    if (sResult.Commands != null)
    {
        var commands = sResult.Commands.Select(x => x.Alias);
        await message.Channel.SendMessageAsync($"Invalid command - Did you mean:\n{string.Join("\n", commands)}");
    }
}
```

I'm fairly sure the code in CommandMapNode.cs could be cleaned up more, but I'm not sure about the best way to do that without making changes that would probably break everything.

### DIFF
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -222,26 +222,28 @@ namespace Discord.Commands
         }
 
         //Execution
-        public SearchResult Search(ICommandContext context, int argPos) 
-            => Search(context, context.Message.Content.Substring(argPos));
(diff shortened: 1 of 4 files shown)


FiniteReality:
I forgot to request a review for this, so cc @RogueException

Auralytical:
I dont like the idea of putting this in the core lib, should be an extension if anything

PR #449, opened 2016-12-28
Enable pre-releases not enabled by default

ghost:
WebSockets won't show up unless Enable pre-releases is ticked, which it isn't

[Discord.Net 1.0]


khionu:
You need Pre-Releases Enabled to get ANY of the 1.0 builds....

khionu:
@RogueException This should be a quicky, what do you think?

Auralytical:
All of the 1.0 builds are prereleases, do we really need to explain that to people? I mean... it's in the giant header at the top :P

PR #450, opened 2016-12-29
Add `IsPrivate` to `ICommandContext`

Joe4evr:
All implementations of `ICommandContext` already define this, and since preconditions now take in the interface, it makes sense to have this on there as well.


Auralytical:
Not really, as this implementation will be the same across all command contexts. It makes more sense as an extension method.

PR #457, opened 2017-01-03
Allow arbitrary attributes to be added to commands

FiniteReality:
I still don't approve adding type info back into commands, so I decided to use this solution for allowing arbitrary attributes to be added to commands.


FiniteReality:
cc @RogueException

FiniteReality:
One moment, I missed some files out while creating the patch. I'll re-open the PR once I've added them.

PR #458, opened 2017-01-03
Allow arbitrary attributes to be added to commands

FiniteReality:
I still don't approve adding type info back into commands, so I decided to use this solution for allowing arbitrary attributes to be added to commands.

This time I added all the files, and squashed into a single commit.

### DIFF
--- a/src/Discord.Net.Commands/Builders/CommandBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/CommandBuilder.cs
@@ -10,6 +10,7 @@ namespace Discord.Commands.Builders
     {
         private readonly List<PreconditionAttribute> _preconditions;
         private readonly List<ParameterBuilder> _parameters;
+        private readonly List<Attribute> _attributes;
         private readonly List<string> _aliases;
 
         public ModuleBuilder Module { get; }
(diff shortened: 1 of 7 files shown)


FiniteReality:
I don't seem to be able to request a review, so cc @RogueException

thedraccx:
Would really love to see this!

Auralytical:
Why leave this at attributes? We could have a typed tag collection, which would let people attach any data they want to a command.

Joe4evr:
How would you attach such arbitrary data to a method command without an attribute?

Joe4evr:
So presuming that would be some sort of `IReadOnlyCollection<TagAttribute>` ***how*** exactly is that any different from just an `IReadOnlyCollection<Attribute>`?

Joe4evr:
So I see your point, but what I initially meant was
> Why leave this at attributes? We could have a typed tag collection, which would let people attach any data they want to a command.

It sounded like Volt's implying to use a system *other* than Attributes, which would work for the builders, but we also have the Class Modules to think about. That's what I was pointing out; as far as I'm aware, you can't attach arbitrary metadata to a method without an attribute.

FiniteReality:
From what I'm understanding, 'typed tag collection' is essentially synonymous with what's in this PR, except valid for any type. I chose Attribute as the type to use since it can be used in the Fluent interface as well as data annotations (using EF terminology here since it's somewhat similar) without making the code stand out as 'strange' compared to the rest of the code.

Joe4evr:
Did you not read the chat backlog? Volt said:
> I said forever ago that I wanted to expose command builders for normal methods so people could do last minute dynamic changes
While attributes are fine, I see no reason why that be there limitation for tagged data
Mayswell allow any additional data of any type. Normal methods would have their attributes auto added to this collection. Builders could do more custom things

FiniteReality:
Oh yeah, I remember now. That still doesn't really change the original thought, it just means I need to change a few names.

PR #460, opened 2017-01-03
Add embed builder extensions

FiniteReality:
People in #dotnet_discord-net suggested that this should be part of the lib after I demonstrated it.

### DIFF
--- a/src/Discord.Net.Rest/Extensions/EmbedBuilderExtensions.cs
+++ b/src/Discord.Net.Rest/Extensions/EmbedBuilderExtensions.cs
@@ -1,0 +1,43 @@
+namespace Discord
+{
+    public static class EmbedBuilderExtensions
+    {
+        public static EmbedBuilder AddField(this EmbedBuilder builder, string title, string text, bool inline = false) =>
+            builder.AddField(field =>
+            {
+                field.Name = title;
+                field.Value = text;
+                field.IsInline = inline;
+            });
+
+        public static EmbedBuilder WithFooter(this EmbedBuilder builder, string text, string iconUrl = null) =>
+            builder.WithFooter(footer =>
+            {
+                footer.Text = text;
+                footer.IconUrl = iconUrl;
+            });
+
+        public static EmbedBuilder WithColor(this EmbedBuilder builder, uint rawValue) =>
+            builder.WithColor(new Color(rawValue));
+
+        public static EmbedBuilder WithColor(this EmbedBuilder builder, byte r, byte g, byte b) =>
+            builder.WithColor(new Color(r, g, b));
+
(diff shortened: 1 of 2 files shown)


khionu (approved):
LGTM

Auralytical:
Should these be extensions? Or part of the EmbedBuilder?

khionu:
Might make more sense as overloads

Erlite:
Looks good 👍

RyadaProductions:
Looking good.

foxbot (approved):

AntiTcb:
Mind throwing in 
```cs
public static EmbedBuilder AddFields(this EmbedBuilder builder, IEnumerable<EmbedFieldBuilder> fields)
        {
            foreach (var field in fields)
                builder.AddField(field);

            return builder;
        }
```
really quickly?

FiniteReality:
Why would that be useful? It seems like something that would have a very specific use-case.

AntiTcb:
This is probably the one specific use case, but I ran into writing that up after opting for a `yield return` for generating fields.  It honestly doesn't add much, but, it doesn't take away anything either by including it.

FiniteReality:
If that's the only use case, you should be writing the extension yourself. There's no real point adding extension methods for one specific use case when they're not even that hard to add yourself

Joe4evr:
@AntiTcb I'm more a fan of this one, really:
```cs
public static EmbedBuilder AddFieldSequence<T>(
    this EmbedBuilder builder,
    IEnumerable<T> seq,
    Action<EmbedFieldBuilder, T> action)
{
    foreach (var item in seq)
    {
        builder.AddField(efb => action(efb, item));
    }

    return builder;
}
```

Auralytical:
Do we still want sub-builders? They don't seem to really add much.

Joe4evr:
I don't think I can do `AddFieldWhen()`/`AddFieldSequence()` nicely without sub-builders, though.

src/Discord.Net.Rest/Extensions/EmbedBuilderExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public static class EmbedBuilderExtensions
+    {
+        public static EmbedBuilder WithColor(this EmbedBuilder builder, uint rawValue) =>
+            builder.WithColor(new Color(rawValue));
+
+        public static EmbedBuilder WithColor(this EmbedBuilder builder, byte r, byte g, byte b) =>
+            builder.WithColor(new Color(r, g, b));
+
+        public static EmbedBuilder WithColor(this EmbedBuilder builder, float r, float g, float b) =>
```
Auralytical, line 11: Needs int overload too

Auralytical (changes requested):

PR #461, opened 2017-01-04
Resolve #446

### LINKED ISSUE #446: RestUserMessage.Reactions throws NRE when no reactions are present
**Affected versions:** 1.0.0-beta2-00523

When accessing the `Reactions` property on a `RestUserMessage` object when no reactions were present in the model (`model.Reactions.IsSpecified` was false), a NullReferenceException is thrown. In this case (no reactions present), I would expect an empty dictionary to be returned, but instead an exception is being thrown.

**Exception details:**
```
   at System.Linq.ImmutableArrayExtensions.ToDictionary[TKey,TElement,T](ImmutableArray`1 immutableArray, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
   at Discord.Rest.RestUserMessage.get_Reactions()
   at ... // my code
```

**Code:**
```c#
foreach (string emote in reactMsg.Reactions) usedEmotes.Add(emote);
```            
`reactMsg` is an `IUserMessage` from a command context, in this case it is a `RestUserMessage`. I have confirmed reactMsg is not null.

**Potential fix:**
The [relevant assignment of the local variable](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Rest/Entities/Messages/RestUserMessage.cs#L105) is only done when reactions are present, thus resulting in a NRE when [extension methods are called to get the property value](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Rest/Entities/Messages/RestUserMessage.cs#L32). To fix, one could assign a default value of an empty array to the underlying field:

```c#
 private ImmutableArray<RestReaction> _reactions = ImmutableArray.Create<RestReaction>();
```
If a default value is used, [this else block](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Rest/Entities/Messages/RestUserMessage.cs#L115) can also be removed as it is redundant.

Alternatively, an else block could be added to the `IsSpecified` if statement which assigns the field to an empty value. Either way, it looks like a relatively easy fix.

firebingo:
Fixes issue #446. 
Fixed RestUserMessage so that if there are no reactions it still initializes the array so _reactions.ToDictionary() on get Reactions does not create an exception.

### DIFF
--- a/src/Discord.Net.Rest/Entities/Messages/RestUserMessage.cs
+++ b/src/Discord.Net.Rest/Entities/Messages/RestUserMessage.cs
@@ -118,6 +118,8 @@ namespace Discord.Rest
                 else
                     _reactions = ImmutableArray.Create<RestReaction>();
             }
+            else
+                _reactions = ImmutableArray.Create<RestReaction>();
 
             if (model.Content.IsSpecified)
             {


foxbot (approved):
LGTM

PR #464, opened 2017-01-05
Fixed guild member's joinedat reseting after certain events

Sentinent:
Changed `JoinedAt` to an optional field, fixing a bug where a guild member's joinedat would reset after certain events (namely GUILD_MEMBER_UPDATE).


foxbot (approved):
LGTM

Issue #407, opened 2016-12-07
Avatar and Username Update


samdivaio:
I found that change of user avatars or usernames are not updated until the bot is restarted which results the commands return the old avatar and username info however nicknames changes and other guild related changes like roles works just fine. Is this a known bug?

foxbot:
Seems like we're probably not handling the state change properly for avatar changes; i'll look into this with #338

AntiTcb:
Was not able to reproduce username issue, I get the following errors with avatar.
`Error: Error getting value from 'WriteTimeout' on 'System.Net.Http.StreamContent+ReadOnlyStream'.`
Using a `Stream` returned from `HttpClient.GetStreamAsync()`

`Error: Error getting value from 'ReadTimeout' on 'System.IO.FileStream'.`
Using a `FileStream` for a local file.

ghost:
Just to add: Not sure if it should happen, but on either username/avatar changes, UserUpdated event seems to be raised for every guild. First argument being the user, and second is the guild user.

khionu:
That's normal. Thank Discord for that

foxbot:
Resolved as of 4665b63d276a8ddd8be01106634c9531b2dcd08d

Issue #272, opened 2016-09-06
Cache Integrity


khionu:
Opening this issue to track known issues with Cache Integrity. I've noticed several, plus we have at least two separate issues that are related.
- GuildPermissions
  - When a Role gains a new permission
- Users"
  - Avatar Change, retains old hash
  - Username Change, retains old username

To Verify
- Discriminators

Pre-RC Update: Someone should do a series of tests to ensure the integrity of all entities and their properties.

FiniteReality:
Do usernames fall under this issue too? 

![](https://cloud.githubusercontent.com/assets/21000186/19127255/270ea99a-8b36-11e6-87ad-181192661536.png)

Between the eval commands, I changed my username to FiniteTesting and the username failed to update. Earlier I tried running UpdateAsync but that threw a NotSupportedException.

khionu:
Yes, it does

RogueException:
Cache logic was redone in beta2, let me know if there are any issues

khionu:
\o/

khionu:
Added amendment to OP

khionu:
I think this is worked on enough to be closed, but I'll leave that to the current maintainers

PR #465, opened 2017-01-06
Fixed users not being updated on PRESENCE_UPDATEs

Sentinent:
Previously, the `user` object in the USER_PRESENCE events were not being used, leading to #407. Also fixes some of the problems in #272.


Sentinent:
The events were already previous raised, but with inaccurate information. At [line 1467](https://github.com/RogueException/Discord.Net/blob/cf2596db5881f67b0e1ea8b46404521a7cc34882/src/Discord.Net.WebSocket/DiscordSocketClient.cs#L1467), the client checks to see if fields in the payload were specified and raised the UserUpdated event. The problem was that the user object from the event were never used after receiving them, making the client not update its user cache.

On rc00544 after changing username:
image

PR #466, opened 2017-01-06
Allow commands to return a Task<RuntimeResult>

FiniteReality:
This allows bot developers to centralize command result logic by using result data whether the command was successful or not.

Example usage:
```csharp
var _result = await Commands.ExecuteAsync(context, argPos);
if (_result is RuntimeResult result)
{
    await message.Channel.SendMessageAsync(result.Reason);
}
else if (!_result.IsSuccess)
{
    // Previous error handling
}
```

The RuntimeResult class can be subclassed too, for example:
```csharp
var _result = await Commands.ExecuteAsync(context, argPos);
if (_result is MySubclassedResult result)
{
    var builder = new EmbedBuilder();
    for (var pair in result.Data)
    {
        builder.AddField(pair.Key, pair.Value, true);
    }
    await message.Channel.SendMessageAsync("", embed: builder);
}
else if (_result is RuntimeResult result)
{
    await message.Channel.SendMessageAsync(result.Reason);
}
else if (!_result.IsSuccess)
{
    // Previous error handling
}
```

### DIFF
--- a/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
@@ -191,22 +191,34 @@ namespace Discord.Commands
 
             var createInstance = ReflectionUtils.CreateBuilder<IModuleBase>(typeInfo, service);
 
-            builder.Callback = async (ctx, args, map, cmd) => 
+            async Task<IResult> ExecuteCallback(ICommandContext context, object[] args, IServiceProvider services, CommandInfo cmd)
             {
-                var instance = createInstance(map);
-                instance.SetContext(ctx);
+                var instance = createInstance(services);
+                instance.SetContext(context);
+
                 try
                 {
                     instance.BeforeExecute(cmd);
+
                     var task = method.Invoke(instance, args) as Task ?? Task.Delay(0);
-                    await task.ConfigureAwait(false);
+                    if (task is Task<RuntimeResult> resultTask)
+                    {
+                        return await resultTask.ConfigureAwait(false);
+                    }
+                    else
+                    {
+                        await task.ConfigureAwait(false);
+                        return ExecuteResult.FromSuccess();
+                    }
                 }
                 finally
                 {
                     instance.AfterExecute(cmd);
                     (instance as IDisposable)?.Dispose();
                 }
-            };
+            }
+
+            builder.Callback = ExecuteCallback;
         }
 
         private static void BuildParameter(ParameterBuilder builder, System.Reflection.ParameterInfo paramInfo, int position, int count, CommandService service)
@@ -284,7 +296,7 @@ namespace Discord.Commands
         private static bool IsValidCommandDefinition(MethodInfo methodInfo)
         {
             return methodInfo.IsDefined(typeof(CommandAttribute)) &&
-                   (methodInfo.ReturnType == typeof(Task) || methodInfo.ReturnType == typeof(void)) &&
+                   (methodInfo.ReturnType == typeof(Task) || methodInfo.ReturnType == typeof(Task<RuntimeResult>)) &&
                    !methodInfo.IsStatic &&
                    !methodInfo.IsGenericMethod;
         }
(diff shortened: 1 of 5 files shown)


khionu:
This was discussed in the Dev Guild.

khionu:
TL;DR, custom errors, or passing any data from the commands. Good for any PostCommand Logic

src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
```diff
@@ -249,7 +249,7 @@ private static bool IsValidModuleDefinition(TypeInfo typeInfo)
         private static bool IsValidCommandDefinition(MethodInfo methodInfo)
         {
             return methodInfo.IsDefined(typeof(CommandAttribute)) &&
-                   methodInfo.ReturnType == typeof(Task) &&
+                   (methodInfo.ReturnType == typeof(Task) || methodInfo.ReturnType == typeof(Task<RuntimeResult>)) &&
```
Joe4evr: ~~Since this PR (allegedly) still needs a discussion, I'm throwing this note in here: this check won't work *entirely* well when people want to return a `Task<CustomRuntimeResult>`.~~

Disregard this, I know where this decision is coming from now.

src/Discord.Net.Commands/Info/CommandInfo.cs
```diff
@@ -138,8 +138,15 @@ public Task<ExecuteResult> ExecuteAsync(ICommandContext context, ParseResult par
                 switch (RunMode)
                 {
                     case RunMode.Sync: //Always sync
-                        await _action(context, args, map).ConfigureAwait(false);
-                        break;
+                        {
+                            var task = _action(context, args, map);
+                            var returnTask = task as Task<RuntimeResult>;
```
foxbot: use c#7 match block

foxbot (changes requested):

src/Discord.Net.Commands/Results/RuntimeResult.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text;
+
+namespace Discord.Commands
+{
+    [DebuggerDisplay(@"{DebuggerDisplay,nq}")]
+    public class RuntimeResult : IResult
+    {
+        private RuntimeResult(CommandError? error, string reason)
```
Joe4evr: Should be `protected` for allowing subclassed RuntimeResults.

FiniteReality: :upside_down_face: I'm sure I managed to subclass this when I tested it originally, good catch though.

src/Discord.Net.Commands/Info/CommandInfo.cs
```diff
@@ -176,12 +175,22 @@ public Task<ExecuteResult> ExecuteAsync(ICommandContext context, ParseResult par
...
 
-        private async Task ExecuteAsyncInternal(ICommandContext context, object[] args, IServiceProvider services)
+        private async Task<IResult> ExecuteAsyncInternal(ICommandContext context, object[] args, IServiceProvider services)
         {
             await Module.Service._cmdLogger.DebugAsync($"Executing {GetLogText(context)}").ConfigureAwait(false);
             try
             {
-                await _action(context, args, services).ConfigureAwait(false);
+                var task = _action(context, args, services);
+                if (task is Task<IResult> returnTask)
```
Joe4evr: I might be crazy, but this doesn't seem like it actually works. A quick test with [this code](https://gist.github.com/Joe4evr/1ba4e5abd7d7f122244cd7d681a45529) seems to print `Nope` every time.
It's possible to do `if (task is Task<RuntimeResult> resultTask)` instead, and not need the second check.

Joe4evr (changes requested):

src/Discord.Net.Commands/Info/CommandInfo.cs
FiniteReality: It does work, but probably not for users using command builders. Whoops.

The reason why I needed to return Task<IResult> is because of ModuleClassBuilder's Callback code: it returns a Task<IResult>, and unfortuantely due to contravariance not working the way I wanted it to, it would never trip the condition.

Issue #480, opened 2017-01-15
Add NullableTypeReader to ParameterBuilder


SubZero0:
To be able to create commands like:

```
[Command("command")]
public async Task Command(Nullable<int> number = null)
{
     if(number == null)
          //DoSomething
     //DoSomething
}
```

FiniteReality:
We already support default values for parameters, so you should be setting that as your default value:
```cs
[Command("command")]
public async Task CommandAsync(int number = 5)
{
    await ReplyAsync($"Hello, I was given {number}.");
}
```

If you want two different things to happen when you provide an int or not, you'd be better using two different overloads:
```cs
[Command("command")]
public async Task CommandDefaultAsync()
{
    await ReplyAsync("Hello! You didn't give me any values!");
}
[Command("command")]
public async Task CommandParameterAsync(int number)
{
    await ReplyAsync($"Hello! You gave me {number}.");
}
```

khionu:
Also, you can add support for your own TypeReaders. It's understandable to want to avoid a set number like `int.MinValue` or `int.MaxValue` as a default, so you can add a Nullable<int> TypeReader.

To learn more: https://discord.foxbot.me/docs/guides/commands.html#type-readers

SubZero0:
@FiniteReality Hm, that's one way but I think they are ugly solutions...

@khionu Didn't know that I could add TypeReaders, will do that.
But the suggestion to add default support for Nullable for already supported Value Types still stands. It would be a nice addition.

khionu:
Perhaps, but I'm not sure I would do that, myself. That said, not my decision.

SubZero0:
If anyone is interested with Nullable, here is how I did it:

```
public class NullableTypeReader<T> : TypeReader
        where T : struct
{
    public delegate bool TryParse<W>(string str, out T value);
    public TryParse<T> tryParseFunc;
    public NullableTypeReader(TryParse<T> parseFunc)
    {
        tryParseFunc = parseFunc;
    }

    public override Task<TypeReaderResult> Read(ICommandContext context, string input)
    {
        T value;
        if (tryParseFunc(input, out value))
            return Task.FromResult(TypeReaderResult.FromSuccess(new Nullable<T>(value)));
        return Task.FromResult(TypeReaderResult.FromSuccess(new Nullable<T>()));
    }
}
```

And when creating your CommandService:
`commands.AddTypeReader<Nullable<int>>(new NullableTypeReader<int>(int.TryParse));`

[...]

PR #484, opened 2017-01-15
Add NullableTypeReaders for ValueTypeReaders

SubZero0:
From #480 

Someone can say that this "doesn't have any utility" but for me that's wrong. It probably doesn't have any utility for _you_ but it has for someone else (like me). And we should let people choose what they want to use.

And the library idea is to make everything easier to who's using it. The default CommandBuilder now does an awesome job, can't deny! But I think it could be better and this is a simple and small way of doing it.

### DIFF
--- a/src/Discord.Net.Commands/Readers/NullableTypeReader.cs
+++ b/src/Discord.Net.Commands/Readers/NullableTypeReader.cs
@@ -1,0 +1,33 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace Discord.Commands
+{
+    internal static class NullableTypeReader
+    {
+        public static TypeReader Create(Type type)
+        {
+            type = typeof(NullableTypeReader<>).MakeGenericType(type);
+            return Activator.CreateInstance(type) as TypeReader;
+        }
+    }
+
+    internal class NullableTypeReader<T> : TypeReader
+        where T : struct
+    {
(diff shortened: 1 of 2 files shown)


src/Discord.Net.Commands/Readers/NullableTypeReader.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            _tryParse = PrimitiveParsers.Get<T>();
+        }
+
+        public override Task<TypeReaderResult> Read(ICommandContext context, string input)
+        {
+            T value;
+            if (_tryParse(input, out value))
+                return Task.FromResult(TypeReaderResult.FromSuccess(new Nullable<T>(value)));
+            return Task.FromResult(TypeReaderResult.FromSuccess(new Nullable<T>()));
```
khionu: This is bad behavior. Failure to parse should not be a valid success condition.

khionu (commented):
As per discussion in the C# guild, this functionality cannot be implemented in a way that adheres to any command syntax, without breaking Text-based Command Paradigm. As noted in the corresponding issue, if a person wishes to implement this themselves, they certainly are free to add a TypeReader, but it's my recommendation that this functionality not be added to the library.

SubZero0:
After talking with Khionu, I think this won't be useful. So I'm closing it.
And I messed up my fork, so I'll need to erase it :/

PR #485, opened 2017-01-16
Return CommandInfo at ExecuteResult

SubZero0:
It would help a lot if the ExecuteResult had the Command it was trying to execute, not just the reason why it failed or if it was successful.

One of the possibilites would be creating a custom error handler at your own CommandHandler without needing to redo what's already inside CommandService.ExecuterAsync or CommandInfo.ExecuteAsync.

So you at least know what command the ExecuteAsync was trying to execute (so you can get the parameters, summary, alias, etc of it).


SubZero0:
Discussed at Discord.Net channel and it would be needed to change CommandService.ExecuteAsync and a lot of the Result classes to add a CommandInfo. Not worth the changes.

If someone is interested, override CommandService.ExecuteAsync, create a new 'Result' with the Command  and get it directly in the ExecuteAsync.

PR #488, opened 2017-01-17
Replace ConcurrentBag<T> with HashSet<T>

### LINKED ISSUE #487: CommandService.RemoveModuleInternal doesn't remove the correct module
After a little investigation, ConcurrentBag.TryTake does *not* remove any specific module, and instead removes a random module (probably the most recent one to be added). Seems I messed this up when implementing it.

FiniteReality:
Fixes #487 in theory (I haven't been able to test properly, dotnet kept complaining about missing targets for some weird reason).

### DIFF
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -18,7 +18,7 @@ namespace Discord.Commands
         private readonly ConcurrentDictionary<Type, ConcurrentDictionary<Type, TypeReader>> _typeReaders;
         private readonly ConcurrentDictionary<Type, TypeReader> _defaultTypeReaders;
         private readonly ImmutableList<Tuple<Type, Type>> _entityTypeReaders; //TODO: Candidate for C#7 Tuple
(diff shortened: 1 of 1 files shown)


src/Discord.Net.Commands/CommandService.cs
```diff
@@ -18,7 +18,7 @@ public class CommandService
         private readonly ConcurrentDictionary<Type, ConcurrentDictionary<Type, TypeReader>> _typeReaders;
         private readonly ConcurrentDictionary<Type, TypeReader> _defaultTypeReaders;
         private readonly ImmutableList<Tuple<Type, Type>> _entityTypeReaders; //TODO: Candidate for C#7 Tuple
-        private readonly ConcurrentBag<ModuleInfo> _moduleDefs;
+        private readonly HashSet<ModuleInfo> _moduleDefs;
```
Joe4evr, line 21: You should also initialize that as a `Hashset<ModuleInfo>` in the ctor.

FiniteReality: I knew I missed *something* out - I wasn't able to test the code properly due to msbuild complaining about missing targets

PR #492, opened 2017-01-20
Make the WS4Net provider target NETStandard

Joe4evr:
I found out there's a beta version of WS4Net that works against NETStandard 1.3 now. This (*finally*) allows me to use WS4Net as a provider on Win7 and also target .NET Core which solves all the ref-def issues that come with trying to use Discord.Net *and* target full framework (because this lib demanded as such).


Joe4evr:
Wait, can I undo that last commit? For some reason I thought I edited the version number in the csproj and then forgot in the project.json, but I didn't edit it in the first place. 🤦‍ Too tired.

Auralytical:
Looks like a stable version of WS4Net 0.15 is never coming out, so I'm closing this. The original issue should be fixed with .NET Standard 2.0.

PR #495, opened 2017-01-23
Change logging example code to use the right Client

Flamanis:
Oh and I also added the ConnectAsync call and Task.Delay

### DIFF
--- a/docs/guides/samples/logging.cs
+++ b/docs/guides/samples/logging.cs
@@ -3,7 +3,6 @@ using Discord.Rest;
 
 public class Program
 {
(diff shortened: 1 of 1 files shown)


khionu:
...... The RestClient is perfectly usable, too...

Flamanis:
I'd say the majority of users would be using it with a SocketClient, and they may get confused. Tbh this was only because it bugged me when someone pointed it out :P

Flamanis:
I also can't read. I'll close this as fox said he was gonna rewrite it at some point.

Issue #498, opened 2017-01-26
Provide a way to indicate the CommandContext has been set.


Joe4evr:
TIL that the Context only gets set *after* module constructors have run. This means that some of the code I had from the beginning, like this excerpt from `MpGameModuleBase`, *never worked*.

```cs
protected MpGameModuleBase(TService gameService)
{
    gameService.OpenToJoin.TryGetValue(Context.Channel.Id, out _open); //BANG
}
```

I already have an idea on a way that would encapsulate this relatively cleanly, PR shortly.

Joe4evr:
Fixed by #502

PR #499, opened 2017-01-26
Add indicator that a module's CommandContext has been set.

Joe4evr:
As mentioned in #498.


Auralytical:
As said in Discord, I'd prefer if we split this up into something like BeforeExecute and AfterExecute. We could also expose the executed commandinfo here.

How Context works is supposed to be invisible to an extent, having events based on when it's set seems awkward to me. We may want other properties here too (like commandinfo?)

Joe4evr:
That's fair, as long as there's a method to set properties like this between the execution of the constructor and the command itself (just like ASP.NET Controllers).

PR #501, opened 2017-01-29
Update to Docs' Homepage

khionu:
Intended to make the Home more than a placeholder, and more homey


Joe4evr:
Would a link to [my annotated quickstart example](https://gist.github.com/Joe4evr/3b54607bf659896179be0d243c770500) be a good addition to this?

PR #502, opened 2017-01-30
Add BeforeExecute/AfterExecute methods to ModuleBase

Joe4evr:
After the feedback of the previous attempt, I decided to try again. I know that Volt would like to expose some other useful things later, but not having a `BeforeExecute()` method at all is a blocker for MpGame.


foxbot (approved):
LGTM

PR #505, opened 2017-02-01
IAudioClient additions

Joe4evr:
I started looking a little bit into audio just to have an idea of how it works so I can help others, and it occurred to me that I can't get the instance that fired an event in a subscriber, nor the Guild that the IAudioClient is connected to.

Basically, I'd like the following to work:
```cs
private Task Connected(IAudioClient client)
{
    return Log($"Connected to voice in {client.Guild.Name}");
}

private Task Disconnected(IAudioClient client, Exception ex)
{
    //need to unsubscribe events otherwise the instances won't get GC'd
    client.Connected -= Connected;
    client.Disconected -= Disconnected;

    return Log($"Disconnected from voice in {client.Guild.Name}");
}
```

### DIFF
--- a/src/Discord.Net.Core/Audio/IAudioClient.cs
+++ b/src/Discord.Net.Core/Audio/IAudioClient.cs
@@ -5,14 +5,16 @@ namespace Discord.Audio
 {
     public interface IAudioClient : IDisposable
     {
-        event Func<Task> Connected;
-        event Func<Exception, Task> Disconnected;
+        event Func<IAudioClient, Task> Connected;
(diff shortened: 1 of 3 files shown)


Auralytical:
If we add this to IAudioClient, should we add it to SocketClient too? Seems weird only having it on one and not the other.

Joe4evr:
What SocketClient? DiscordSocketClient doesn't connect one instance per guild like the AudioClient does.

Auralytical:
This needs to be updated to support the new events too

Issue #514, opened 2017-02-06
Wrong ExecuteAsync result with preconditions


SubZero0:
ExecuteAsync is returning Unknown command with commands inside groups when a precondition fails  (instead of returning the PreconditionResult).

Examples:
![image](https://cloud.githubusercontent.com/assets/3085008/22632362/a9b43fc0-ebff-11e6-8233-c3972dba7ee7.png)
![image](https://cloud.githubusercontent.com/assets/3085008/22632250/cc779a4e-ebfe-11e6-849b-f349864a4088.png)

tristanmcpherson:
I believe the issue is with this [line](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Commands/CommandService.cs#L256).

In testing yours it seems that `commands.Count == 2` so it continues on and finds no more commands instead of returning that result.

Auralytical:
This behavior was by design. If there are multiple commands, whose failed precondition do you return? It was only ever safe to return PreconditionResult if there was only one command.

As such, #515 is not a valid fix.

Maybe the result classes should be reworked to be more friendly with multiple matches. I'll look into it.

foxbot:
@RogueException given the above example, only one command should be matched when searching. There isn't an ambiguity between `test` and `test two` when determining which one had an invalid precondition if the input was `test two`

Auralytical:
Oh huh, you're right. It shouldnt even reach the preconditions with 2 commands. Must be an issue in the search function.

SubZero0:
@RogueException Not just with search. ExecuteAsync will still throw the wrong error if everything fails.
Example:

```
[Command("test"))]
[RequireContext(Context.Group), Priority(1)]
...

[Command("test")]
[RequireContext(Context.DM), Priority(0)] //lowest priority value = highest priority
...
```
Send `test` at a **Guild** channel.

ExecuteAsync will evaluate the Search result as:
`Context.DM test` first
`Context.Group test` second

**Current ExecuteAsync logic:**
Try `Context.DM test`, fails precondition, skip; try `Context.Group test`, fails at preconditions, skip; return unknown command

That's wrong, it exists and it should return the precondition error, since the priority is for DM Context, it should return the PreconditionResult from `Context.DM test`

That's why I created the second option variable at #516, that's the highest priority with a successful parse and failed precondition(s).
[...]

PR #515, opened 2017-02-06
Fix issue with precondition checking

tristanmcpherson:
Previously `commands.Count` would never change, so by the current logic
it was never checking the last command.

Fixes: https://github.com/RogueException/Discord.Net/issues/514


Auralytical:
See my answer in #514

PR #516, opened 2017-02-06
Fix issue retrieving the most probable command from ExecuteAsync

### LINKED ISSUE #514: Wrong ExecuteAsync result with preconditions

SubZero0:
The main culprit is the `continue;` in the precondition check in the ExecuteAsync. It will ignore the parse, even with the possibility of it being the right command, just without the preconditions.
After changing the logic of the `else continue`, CommandInfo.ParseAsync will fail because the preconditions failed, even being the right command, so it was removed from the ParseAsync, so it'll at least try to parse it.

A second option was added, it'll will be the highest priority command with a successful parse.

The loop won't stop before reaching the end, so if it finds a perfect match that's not the second option, it'll return the perfect one, if not, the second option (that had the successful parse, highest priority, but missing preconditions).

Fixes #514

### DIFF
--- a/src/Discord.Net.Commands/CommandMatch.cs
+++ b/src/Discord.Net.Commands/CommandMatch.cs
@@ -16,8 +16,8 @@ namespace Discord.Commands
 
         public Task<PreconditionResult> CheckPreconditionsAsync(ICommandContext context, IDependencyMap map = null)
             => Command.CheckPreconditionsAsync(context, map);
-        public Task<ParseResult> ParseAsync(ICommandContext context, SearchResult searchResult, PreconditionResult? preconditionResult = null)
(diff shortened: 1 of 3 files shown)


tristanmcpherson:
Not quite sure this is the proper fix, check out my PR. There's a bug in the logic that still remains that fixes this issue.

SubZero0:
Rewrote the first comment so it's easier to understand what was happening and what I did to fix it.

@IlmrynX your PR won't fix. It'll only send the second last possibility (the second lowest priority) just because it failed the precondition test without any guarantee that it's the right command (and ignores the last item). It's actually opening more holes to the ExecuteAsync logic.

PR #503, opened 2017-01-31
Update doc samples

Joe4evr:
Fox told me my annotated quickstart example is a good replacement for the first-steps sample in the docs. I decided to take a quick look at all the other samples while I was at it.

PR #520, opened 2017-02-10
Allow IDependencyMap injection for public properties

james7132:
This should allow dependency injection into publicly writable properties. Will definitely help avoid boilerplate constructor code in modules.

### DIFF
--- a/docs/guides/commands.md
+++ b/docs/guides/commands.md
@@ -168,7 +168,7 @@ IDependencyMap as an argument and extract services yourself.
 
 ### Module Properties
 
-Modules with public settable properties will be have them injected after
+Modules with public settable properties will have them injected after module
 construction.
 
 ### Module Groups
--- a/docs/guides/commands.md
+++ b/docs/guides/commands.md
@@ -5,16 +5,16 @@
 
 ## Setup
 
-To use Commands, you must create a [Commands Service] and a 
+To use Commands, you must create a [Commands Service] and a
 Command Handler.
 
-Included below is a very bare-bones Command Handler. You can extend 
-your Command Handler as much as you like, however the below is the 
+Included below is a very bare-bones Command Handler. You can extend
+your Command Handler as much as you like, however the below is the
 bare minimum.
 
-The CommandService optionally will accept a [CommandServiceConfig], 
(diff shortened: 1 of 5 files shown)


FiniteReality:
I think that if we're going to provide dependency injection for properties, we should provide an attribute to ignore certain properties. For example:

```cs
[Inject(false)]
public MyCustomType MyProperty { get; set }
```

I'm not sure what the best name would be though; maybe something like `IgnoreInjectAttribute`?

james7132:
PTAL.

FiniteReality:
LGTM

docs/guides/samples/dependency_module.cs
```diff
@@ -20,12 +21,21 @@ public ModuleA(DatabaseService database)
...
+
+    // Public properties annotated with [DontInject] will not
+    [DontInject]
+    public NotificationService { get; set; }
+
     public ModuleB(CommandService commands, IDependencyMap map)
     {
-        _commands = commands;
+        Commands = commands;
         _notification = map.Get<NotificationService>();
```
foxbot: This line should be removed entirely, it's against our official-unofficial design guidelines to use `map.Get<T>` anymore.

foxbot (approved):

docs/guides/commands.md
```diff
@@ -63,123 +63,128 @@ By now, your module should look like this:
...
-that are placed in the constructor must be injected into an 
-@Discord.Commands.IDependencyMap. Alternatively, you may accept an 
+Modules are constructed using Dependency Injection. Any parameters
+that are placed in the constructor must be injected into an
+@Discord.Commands.IDependencyMap. Alternatively, you may accept an
 IDependencyMap as an argument and extract services yourself.
 
+### Module Properties
+
+Modules with public settable properties will be have them injected after
```
tristanmcpherson: `will be have` - probably a typo

docs/guides/samples/dependency_module.cs
Joe4evr: *coughcough* #503

james7132:
Is there anything else that needs to be changed on this PR for it to be merged?

PR #523, opened 2017-02-13
Fixed couple non-default arguments in REST messages

Emzi0767:
This fixes `RequestOptions` not having a default value on `.ModifyAsync`, `.PinAsync`, and `.UnpinAsync`.


Auralytical:
Oh I missed these, good catch

PR #526, opened 2017-02-15
Add Roles property to SocketGuildUser

foxbot:
there's a few different ways to select a user's roles from their role IDs, and this one seems to be the most efficient.

doesn't seem like there's any reason this shouldn't be included.


Joe4evr:
I feel like `Roles` has been on and off the `SocketGuildUser` object many times now.

Joe4evr:
I will defer to Volt's judgement on whether or not to add this. ¯\\\_(ツ)\_/¯

Auralytical:
Not intentional, can remove `RoleIds` and replace with this.

However, you should expose it as an `IReadOnlyCollection` using my `.ToReadOnlyCollection(x => _roleIds)` or something like that which maps the count to a different collection.

PR #530, opened 2017-02-19
Format.Sanitize escapes mention special chars now. Fixes 529.

raphaelr:


khionu:
URL Embeds are also escaped via being surrounded by angular brackets, not to mention this would render the text as `@152037375890030592` which is no good. The Format class is meant to handle Markdown Formatting, specifically.

This is neither a working fix or a valid usecase for the method. Please test your PRs.

PR #532, opened 2017-02-20
Improve parameter precondition type safety

FiniteReality:
Also removes some terrible code which was left over when I first implemented parameter preconditions. I don't know why that was there.

With this commit, parameter preconditions should be much safer as they use generic methods instead of janky casting of objects.


Auralytical:
Iirc we decided not to do this in Discord?

FiniteReality:
I think we decided against the generic CheckPermisssions, but I think there
are a bunch of useful improvements elsewhere in this PR. I'll remove the
generics stuff for now.

Auralytical:
LGTM

PR #533, opened 2017-02-21
Fix [Priority] behaving inverse to spec.

Joe4evr:
Because after command matches are sorted the `ExecuteAsync()` method goes through them backwards (for a tiny perf gain), command priorities behave inverse to what was originally spec'd for them. This fixes it without changing to a forwards `for` loop.


Joe4evr:
Closing, as #534 was merged instead.

PR #534, opened 2017-02-21
Fix order of iteration in ExecuteAsync

FiniteReality:
:hash: BlameFinite


Joe4evr:
In case it wasn't obvious from the debate we had over this, this and #533 are mutually exclusive.

PR #535, opened 2017-02-21
Add type constraints for Add/RemoveModuleAsync

FiniteReality:
I keep forgetting to do this, let's get it out of the way while I still remember.

### DIFF
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -73,7 +73,7 @@ namespace Discord.Commands
                 _moduleLock.Release();
             }
         }
-        public async Task<ModuleInfo> AddModuleAsync<T>() where T : IModuleBase
+        public async Task<ModuleInfo> AddModuleAsync<T>() where T : IModuleBase<ICommandContext>
         {
             await _moduleLock.WaitAsync().ConfigureAwait(false);
             try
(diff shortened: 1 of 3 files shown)


src/Discord.Net.Commands/CommandService.cs
```diff
@@ -73,7 +73,7 @@ public CommandService(CommandServiceConfig config)
                 _moduleLock.Release();
             }
         }
-        public async Task<ModuleInfo> AddModuleAsync<T>()
+        public async Task<ModuleInfo> AddModuleAsync<T>() where T : ModuleBase
```
foxbot: This should be `IModuleBase`

src/Discord.Net.Commands/CommandService.cs
```diff
@@ -143,7 +143,7 @@ private ModuleInfo LoadModuleInternal(ModuleInfo module)
                 _moduleLock.Release();
             }
         }
-        public async Task<bool> RemoveModuleAsync<T>()
+        public async Task<bool> RemoveModuleAsync<T>() where T : ModuleBase
```
foxbot: Also should be `IModuleBase`

foxbot (dismissed):

Joe4evr:
I have the solution to this conundrum of the generic type constraints.

1) Add the following interface:
```cs
public interface IModuleBase<out T>
    where T : ICommandContext
{
    // This property isn't really required, but it'd be good to have regardless
    T Context { get; }
}
```
2) Change the `ModuleBase<T>` declaration so it implements both the existing and the new one:
```cs
public class ModuleBase<T> : IModuleBase, IModuleBase<T>
    where T : class, ICommandContext // Existing constraint is equal (actually tighter), so no change
```
3) Use the new generic interface as type parameter constraint:
```cs
public async Task<ModuleInfo> AddModuleAsync<T>() where T : IModuleBase<ICommandContext>
```
Compiler happy, user happy.

FiniteReality:
Done, PTAL

FiniteReality:
Closing for now since we're discussing a different approach.

PR #540, opened 2017-02-26
GuildUserExtensions removed in favour of atomic role add/remove endpoints

Emzi0767:
Am I doing this right?


Emzi0767:
I made some changes as per suggestions from @foxbot:

* `ChangeRolesAsync` -> `ModifyRolesAsync`
* `ModifyRolesAsync` has a new overload that takes a single `IEnumerable<IRole>`, which replaces all roles on a user

james7132:
Not sure what the benefit of this is. The implementation should be uniform across all GuildUsers.

khionu:
This is not only boilerplate code, it fails to provide for RPC entities. There is no good gained by having this code split across 3 classes, when it does exactly the same thing. This PR has no merit.

In addition, this would not compile for all portions of the Library, as the RpcGuildUser no longer fully implements the interface. When testing PRs, compile all segments of the library.

khionu:
It's still boilerplate code, even if it compiles. There is nothing gained from this aside from getting rid of extension methods. Extension methods, or Boilerplate code? One is design weirdness, one is actually worse when compiled to IL.

Auralytical:
I would probably get rid of ModifyRolesAsync entirely.

khionu:
There are use cases for adding and removing in bulk. Given the API supports bulk role modification, I think we should too. That said, adding and removing individual roles might be best to use the endpoints that allow it.

I can open an issue for this.

Auralytical:
Why do we need adding+removing in a single operation though? The issue that was made for has already been resolved by Discord.

khionu:
Because the API supports it through the ModifyGuildUser endpoint. We already have the code for it, and it is part of the API, so why shouldn't we continue supporting it? We don't, however, support the AddRole and RemoveRole endpoints, that deal with singular roles, last I knew.

Giving Bulk Add and Bulk Remove through two separate methods that access the same endpoint seems redundant. While the Extension Methods for this are not really nice design choices, they are the best design choices, for this, from my perspective.

A goal was to mirror the API as best as possible, though I haven't seen us mention that in a while. Keeping ModifyRoles goes with that goal.

Auralytical:
I was talking mainly about `ModifyRolesAsync(add, remove)` which is not needed at all. But the others arent needed for exactly what you said: its behavior exists elsewhere.

khionu:
This isn't really the appropriate place for the discussion. Maybe we should open an issue, or we could move it to the Dev Guild?

Emzi0767:
So to address concerns:

* [This was requested by Volt](http://i.imgur.com/8G01Add.png), although I admit, I may have misunderstood what Volt meant.
* 2nd and 3rd commits are a result of further discussion with Fox.
* I can still make further changes, based on feedback.

PR #541, opened 2017-02-27
Added GetPrecondition<TPrecondition>() to ModuleInfo and CommandInfo

MarkusG:
An easier way of getting Preconditions for modules and commands without having to use LINQ on `x.Preconditions` yourself.

### DIFF
--- a/src/Discord.Net.Commands/Info/CommandInfo.cs
+++ b/src/Discord.Net.Commands/Info/CommandInfo.cs
@@ -63,6 +63,9 @@ namespace Discord.Commands
             _action = builder.Callback;
         }
 
+        public TPrecondition GetPrecondition<TPrecondition>() where TPrecondition : PreconditionAttribute =>
+            Preconditions.FirstOrDefault(x => x is TPrecondition) as TPrecondition;
+
         public async Task<PreconditionResult> CheckPreconditionsAsync(ICommandContext context, IDependencyMap map = null)
         {
             if (map == null)
--- a/src/Discord.Net.Commands/Info/CommandInfo.cs
+++ b/src/Discord.Net.Commands/Info/CommandInfo.cs
@@ -63,8 +63,8 @@ namespace Discord.Commands
             _action = builder.Callback;
         }
 
-        public TPrecondition GetPrecondition<TPrecondition>() where TPrecondition : PreconditionAttribute =>
-            Preconditions.FirstOrDefault(x => x.GetType() == typeof(TPrecondition)) as TPrecondition;
+        public IEnumerable<TPrecondition> GetPrecondition<TPrecondition>() where TPrecondition : PreconditionAttribute =>
+            Preconditions.Where(x => x.GetType() == typeof(TPrecondition)).Select(x => x as TPrecondition);
 
         public async Task<PreconditionResult> CheckPreconditionsAsync(ICommandContext context, IDependencyMap map = null)
         {
--- a/src/Discord.Net.Commands/Info/CommandInfo.cs
+++ b/src/Discord.Net.Commands/Info/CommandInfo.cs
@@ -64,7 +64,7 @@ namespace Discord.Commands
         }
 
         public TPrecondition GetPrecondition<TPrecondition>() where TPrecondition : PreconditionAttribute =>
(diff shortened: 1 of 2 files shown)


khionu (approved):
LGTM

khionu:
@FiniteReality

FiniteReality:
Personally I feel that these should be extension methods as they don't add
that much new functionality.

Also, I'm not sure how your logic handles multiple preconditions inheriting
from the same base precondition, e.g.:

```cs
class MyBasePrecondition : Precondition attribute { }
class MySubPrecondition1 : MyBasePrecondition { }
```

If both of these preconditions were specified on a module or command, I
think that trying to retrieve one may instead retrieve the other, which I
think is confusing behaviour. You could filter the preconditions based on
the initial condition, and then do further filtering to be more specific if
multiple preconditions are found.

khionu:
I actually have to agree with the point made, regarding inheritance. That is undesirable behavior, if it could pull only the first of 3. The `MyBasePrecondition` would grab either, so the type matching should be a bit more explicit.

That said, it's worth considering whether it would be worth fixing it, given Reflection would be necessary

MarkusG:
Resolved the issue described by @FiniteReality.

src/Discord.Net.Commands/Info/CommandInfo.cs
```diff
@@ -63,6 +63,9 @@ internal CommandInfo(CommandBuilder builder, ModuleInfo module, CommandService s
             _action = builder.Callback;
         }
 
+        public TPrecondition GetPrecondition<TPrecondition>() where TPrecondition : PreconditionAttribute =>
```
Auralytical: This should return an `IEnumerable` of preconditions instead of using `FirstOrDefault`. We can't assume the user wants only the first result.

Auralytical (changes requested):

src/Discord.Net.Commands/Info/CommandInfo.cs
MarkusG: A command/module can't have duplicate attributes. In what case would it need to return an `IEnumerable`?

Joe4evr: You can if you set `AllowMultiple = true`, but whether that's actually useful is a completely different thing.

Also, for the query itself, wouldn't it be better to use `x => x is TPrecondition`?

MarkusG: Yeah, I don't know of any case where someone would have duplicate preconditions on a command. As for `x => x is TPrecondition`, see earlier in the PR conversation. That would cause undesirable behavior with preconditions that derive from other preconditions.

FiniteReality: What about `RequireUserPermission` or `RequireBotPermission`?

MarkusG: I suppose that's true. Will commit momentarily.

Joe4evr: Ah, I see. But if you *had* a base Precondition, then it would make sense to get all of the applied preconditions when you do `GetPreconditions<MyBasePrecondition>()` instead of none of them.

src/Discord.Net.Commands/Info/CommandInfo.cs
```diff
@@ -63,6 +63,9 @@ internal CommandInfo(CommandBuilder builder, ModuleInfo module, CommandService s
             _action = builder.Callback;
         }
 
+        public IEnumerable<TPrecondition> GetPrecondition<TPrecondition>() where TPrecondition : PreconditionAttribute =>
+            Preconditions.Where(x => x.GetType() == typeof(TPrecondition)).Select(x => x as TPrecondition);
```
Joe4evr: 1) Should probably pluralize the name. 2) You can use `.Cast<TPrecondition>()`.

Auralytical:
Can't this whole thing be reduced to a single `Preconditions.OfType<T>()` call?

MarkusG:
That does appear to be the case now.

MarkusG:
I should've closed this a while ago. Whoops.

PR #546, opened 2017-03-01
Fix detection of IDependencyMap impl

khionu:
~~Not pretty, but it works.~~ Now it's pretty, thanks @FiniteReality

### DIFF
--- a/src/Discord.Net.Commands/Dependencies/DependencyMap.cs
+++ b/src/Discord.Net.Commands/Dependencies/DependencyMap.cs
@@ -38,8 +38,10 @@ namespace Discord.Commands
         public void AddFactory<T>(Func<T> factory) where T : class
         {
             var t = typeof(T);
-            if (typeof(T) == typeof(IDependencyMap) || typeof(T) == typeof(CommandService))
-                throw new InvalidOperationException("The dependency map cannot contain services directly added as IDependencyMap or CommandService. Only Implimentations and Derivatives are permitted");
+            if (typeof(T) == typeof(IDependencyMap))
+                throw new InvalidOperationException("IDependencyMap is used internally and cannot be added as a dependency");
+            if (typeof(T) == typeof(CommandService))
+                throw new InvalidOperationException("CommandService is used internally and cannot be added as a dependency");
             if (map.ContainsKey(t))
                 throw new InvalidOperationException($"The dependency map already contains \"{t.FullName}\"");
             map.Add(t, factory);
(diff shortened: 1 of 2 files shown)


src/Discord.Net.Commands/Dependencies/DependencyMap.cs
```diff
@@ -38,6 +38,8 @@ public bool TryAddTransient<T>() where T : class, new()
         public void AddFactory<T>(Func<T> factory) where T : class
         {
             var t = typeof(T);
+            if (typeof(T) == typeof(IDependencyMap) || typeof(T) == typeof(CommandService))
+                throw new InvalidOperationException("The dependency map cannot contain services directly added as IDependencyMap or CommandService. Only Implimentations and Derivatives are permitted");
```
foxbot: Implementations

src/Discord.Net.Commands/Dependencies/DependencyMap.cs
```diff
@@ -48,6 +50,8 @@ public bool TryAddTransient<T>() where T : class, new()
             var t = typeof(T);
             if (map.ContainsKey(t))
                 return false;
+            if (typeof(T) == typeof(IDependencyMap) || typeof(T) == typeof(CommandService))
+                throw new InvalidOperationException("The dependency map cannot contain services directly added as IDependencyMap or CommandService. Only Implimentations and Derivatives are permitted");
```
foxbot: Implementations

src/Discord.Net.Commands/Utilities/ReflectionUtils.cs
```diff
@@ -51,14 +51,16 @@ internal static T CreateObject<T>(TypeInfo typeInfo, CommandService service, IDe
             };
         }
 
+        private static readonly TypeInfo _dependencyTypeInfo = typeof(IDependencyMap).GetTypeInfo();
```
foxbot, line 54: What is this even used for?

src/Discord.Net.Commands/Dependencies/DependencyMap.cs
foxbot: "Only objects that derive from IDependencyMap are supported" would be better truthfully. Also, neither implementations nor derivatives are proper nouns, no need to capitalize them.

src/Discord.Net.Commands/Dependencies/DependencyMap.cs
Auralytical: I'd clean this up by looking for each type individually and splitting this into two exception messages.

src/Discord.Net.Commands/Dependencies/DependencyMap.cs
Auralytical: I'd probably just keep it simple and explain why. `IDependencyMap is used internally and cannot be added as a dependency`.

Auralytical (changes requested):

src/Discord.Net.Commands/Dependencies/DependencyMap.cs
khionu: Same for CommandService?

khionu: IE, `s/IDependencyMap/CommandService/`, but same message?

PR #549, opened 2017-03-03
Added support for reaction 'me' field

Mushroom:


khionu (approved):
LGTM, but this ReactionMetadata should be expanded later on, so, as with InviteMetadata, there is a summary Reaction and more detailed ReactionMetadata, which will contain the user collection.

PR #560, opened 2017-03-08
UserTypeReader ArgumentNullException in DM

### LINKED ISSUE #559: Error with the UserTypeReader when not in a Guild channel.
In this file: https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Commands/Readers/UserTypeReader.cs there are checks by nickname and with var guildUser, which only exist inside a guild

If you have the SocketUser param in one of your commands, this error will always come up if the command is not used in a guild: http://prntscr.com/ehbf4b

Flamanis:
Fixes #559

When you attempted to use the UserTypeReader in a DM channel, the TypeReader would attempt to use the Linq extension method Where on a null collection. The collection was null because it was a collection of Guild users, which in a DM would be null as it's checked before attempting to instantiate the collection.

### DIFF
--- a/src/Discord.Net.Commands/Readers/UserTypeReader.cs
+++ b/src/Discord.Net.Commands/Readers/UserTypeReader.cs
@@ -60,12 +60,14 @@ namespace Discord.Commands
             {
                 foreach (var channelUser in channelUsers.Where(x => string.Equals(input, x.Username, StringComparison.OrdinalIgnoreCase)))
                     AddResult(results, channelUser as T, channelUser.Username == input ? 0.65f : 0.55f);
-
+                
+                if(context.Guild != null)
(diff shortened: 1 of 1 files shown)


foxbot (approved):
LGTM, I assume this is tested?

Flamanis:
It is now tested, however I modified it from the original format as I noticed that when attempting to look up a User by the format `Name#Discrim` the second `FirstOrDefault` call was actually not being run on the correct collection, it was supposed to be run on `guildUsers`. I talked with Weebu about just adding another null check for that, and he recommended just initializing `guildUsers` as an empty collection and removing the null checks. Only other change I had to do after that was use the null conditional operator on line 61 to avoid a nullref.

PR #565, opened 2017-03-12
Add installing with nuget using VSC to guides

Bond-009:


foxbot (changes requested):
I'd personally leave out the picture of the NuGet.config and instead link to the section about 'adding feeds outside of visual studio'.

Leaving out the image for the csproj and instead just placing it directly in samples and including a codeblock would also be preferred.

foxbot (approved):

PR #573, opened 2017-03-15
Automatic animated avatar detection

james7132:
Right now avatars are statically fetched depending on the parameter passed into GetAvatarUrl. This PR does the following:

* Adds a AvatarFormat.Auto option and makes it the default for all GetAvatarUrl implementations.
* When used, it determines the type of avatar using the "a_" prefix used to differentiate animated avatars to fetch gifs, otherwise it defaults to PNGs.

I have been getting complaints on how all avatars returned by this call have horrible artifacts due to the GIF format. This should resolve them.

### DIFF
--- a/src/Discord.Net.Core/CDN.cs
+++ b/src/Discord.Net.Core/CDN.cs
@@ -8,11 +8,9 @@
         {
             if (avatarId == null)
                 return null;
-            var baseUrl = $"{DiscordConfig.CDNUrl}avatars/{userId}/{avatarId}";
             if (format == AvatarFormat.Auto)
-                return baseUrl + (avatarId.StartsWith("a_") ? "gif" : "png") + $"?size={size}";
-            else
-                return baseUrl + format.ToString().ToLower() + $"?size={size}";
+                format = avatarId.StartsWith("a_") ? AvatarFormat.Gif : AvatarFormat.Png;
+            return $"{DiscordConfig.CDNUrl}avatars/{userId}/{avatarId}.{format.ToString().ToLower()}?size={size}";
         }
         public static string GetGuildIconUrl(ulong guildId, string iconId)
             => iconId != null ? $"{DiscordConfig.CDNUrl}icons/{guildId}/{iconId}.jpg" : null;
--- a/src/Discord.Net.Core/CDN.cs
+++ b/src/Discord.Net.Core/CDN.cs
@@ -4,14 +4,15 @@
     {
         public static string GetApplicationIconUrl(ulong appId, string iconId)
             => iconId != null ? $"{DiscordConfig.CDNUrl}app-icons/{appId}/{iconId}.jpg" : null;
-        public static string GetUserAvatarUrl(ulong userId, string avatarId, ushort size, AvatarFormat format) {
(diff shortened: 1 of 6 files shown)


src/Discord.Net.Core/CDN.cs
```diff
@@ -4,8 +4,15 @@ public static class CDN
     {
         public static string GetApplicationIconUrl(ulong appId, string iconId)
             => iconId != null ? $"{DiscordConfig.CDNUrl}app-icons/{appId}/{iconId}.jpg" : null;
-        public static string GetUserAvatarUrl(ulong userId, string avatarId, ushort size, AvatarFormat format)
-            => avatarId != null ? $"{DiscordConfig.CDNUrl}avatars/{userId}/{avatarId}.{format.ToString().ToLower()}?size={size}" : null;
+        public static string GetUserAvatarUrl(ulong userId, string avatarId, ushort size, AvatarFormat format) {
+            if (avatarId == null)
```
foxbot: Is there ever a case where we pass an empty avatarId into this function?

src/Discord.Net.Core/CDN.cs
```diff
@@ -4,8 +4,15 @@ public static class CDN
     {
         public static string GetApplicationIconUrl(ulong appId, string iconId)
             => iconId != null ? $"{DiscordConfig.CDNUrl}app-icons/{appId}/{iconId}.jpg" : null;
-        public static string GetUserAvatarUrl(ulong userId, string avatarId, ushort size, AvatarFormat format)
-            => avatarId != null ? $"{DiscordConfig.CDNUrl}avatars/{userId}/{avatarId}.{format.ToString().ToLower()}?size={size}" : null;
+        public static string GetUserAvatarUrl(ulong userId, string avatarId, ushort size, AvatarFormat format) {
```
foxbot: Bracket should go on the next line down

src/Discord.Net.Core/CDN.cs
```diff
@@ -4,8 +4,15 @@ public static class CDN
...
+            if (avatarId == null)
+                return null;
+            var base = $"{DiscordConfig.CDNUrl}avatars/{userId}/{avatarId}";
```
foxbot: `base` is a reserved keyword, this needs a different name

foxbot (changes requested):

src/Discord.Net.Core/CDN.cs
james7132: It was part of the original function, and I preserved it. To my knowledge, it should never be null.

src/Discord.Net.Core/CDN.cs
james7132: Done

src/Discord.Net.Core/CDN.cs
james7132: Done.

james7132:
PTAL.

foxbot (approved):

src/Discord.Net.Core/CDN.cs
Sentinent: A user can have a null avatarId if they still have their default avatar. We might be able to handle this better by calculating the default avatar instead of just returning null.

PR #581, opened 2017-03-21
Overloaded AddModuleAsync with Type

Bond-009:
IDK

### DIFF
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -154,10 +154,7 @@ namespace Discord.Commands
                 _moduleLock.Release();
             }
         }
-        public Task<bool> RemoveModuleAsync<T>()
-        {
-            return RemoveModuleAsync(typeof(T));
-        }
+        public Task<bool> RemoveModuleAsync<T>() => RemoveModuleAsync(typeof(T));
         public async Task<bool> RemoveModuleAsync(Type type)
         {
             await _moduleLock.WaitAsync().ConfigureAwait(false);
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -83,20 +83,21 @@ namespace Discord.Commands
                 _moduleLock.Release();
             }
         }
-        public async Task<ModuleInfo> AddModuleAsync<T>()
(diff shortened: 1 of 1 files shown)


MarkusG:
How is this different from `AddModule<T>()`?

Bond-009:
You can do things like this:
```csharp
foreach (KeyValuePair<System.Type, bool> pair in dic)
{
    if (pair.Value)
    {
        await CommandService.AddModuleAsync(pair.Key);
    }
}
```

FiniteReality:
:-1: Code duplication which could have been avoided by redirecting one method to the other. Also, there is no check that the type is actually a valid module before attempting to build it.

@MarkusGordathian `AddModule(type)` is useful for when you don't have access to use its generic twin, e.g. if you're loading modules from a source other than an Assembly.

Bond-009:
@FiniteReality better?

src/Discord.Net.Commands/CommandService.cs
```diff
@@ -83,20 +83,21 @@ public CommandService(CommandServiceConfig config)
                 _moduleLock.Release();
             }
         }
-        public async Task<ModuleInfo> AddModuleAsync<T>()
+        public async Task<ModuleInfo> AddModuleAsync<T>() => await AddModuleAsync(typeof(T));
```
FiniteReality: Remove `async` and `await` here, there's no need to use an async context as we're just redirecting to the other method.

Bond-009: Done

FiniteReality:
:+1: LGTM, with this I can add a nice check to AddModuleAsync to prevent non-module types from attempting to be added before we try to build them.

Auralytical (changes requested):
This change should be done to RemoveModuleAsync too for consistency.

src/Discord.Net.Commands/CommandService.cs
```diff
@@ -153,13 +154,17 @@ private ModuleInfo LoadModuleInternal(ModuleInfo module)
                 _moduleLock.Release();
             }
         }
-        public async Task<bool> RemoveModuleAsync<T>()
+        public Task<bool> RemoveModuleAsync<T>()
```
FiniteReality: Use an expression-bodied method here for consistency

PR #590, opened 2017-03-26
Fires Ready event after Resumed dispatch message is done

Cisien:
The Ready event isn't fired when the client recovers from being disconnected from Discord, this solves the gap between Connected and Ready for this case.

### DIFF
--- a/src/Discord.Net.WebSocket/DiscordSocketClient.cs
+++ b/src/Discord.Net.WebSocket/DiscordSocketClient.cs
@@ -519,6 +519,7 @@ namespace Discord.WebSocket
                                             await GuildAvailableAsync(guild).ConfigureAwait(false);
                                     }
 
+                                    await _readyEvent.InvokeAsync().ConfigureAwait(false);
                                     await _gatewayLogger.InfoAsync("Resumed previous session").ConfigureAwait(false);
                                 }
                                 break;


khionu:
The Ready event is fired when Discord sends us the Ready Dispatch. This is not meant to be more than that.

khionu:
Huh.... so then this might be appropriate, if we want to make it a multiple occurrence

Mushroom:
I disagree with this - I believe the ready event should behave as is, ie, fires after receiving a READY packet/full (re)connection. I would suggest a separate event for resumes. 

If not that, then at least an indication of whether the event was fired by a resume, or an full (re)connect.

For example, in my client, I reload certain data on the ready event - if it was after a resume, then the data would not need to be reloaded, as the client has successfully resumed.

khionu:
Hmm..... do we have a way, atm, to tell the difference between when the client has Resumed vs done a full reconnect? We could pass a bool "IsResume", or maybe pass an int, counting resumes? Anything, really, to help distinguish.

Cisien:
A separate event would work just as well for my case.
When disconnected I un-register event handlers, and expect to register them again in ready. I think the Resumed dispatch message would be all that's needed to know if it was a resume vs full reconnect

Mushroom:
Right now there is no way of telling if the client has resumed, as far as I can tell, other than watching the availability of the guilds.

I would say firing "Ready" on a resume is not a good idea, as that doesn't really indicate what really happened. I propose a separate "Resume" event.

khionu:
I disagree with a separate event. We already have far too many, and this is too closely related

Cisien:
Closed after further discussion in the discord channel. We determined a separate event wasn't needed to cover my use case

PR #603, opened 2017-04-09
Expose the 'fields' collection on EmbedBuilder

foxbot:
After some discussion I decided that there was really no reason to keep this private, and it didn't really go along with the rest of the design of the EmbedBuilder.

This is NOT a breaking change.

Exposing this property should not have any negative effects.

(ignore docs commit god bless git)


Still34:
Yes, please. I was looking into it the other day.

Auralytical:
Add a null check in build, skipping the loop if fields is null and then LGTM

PR #617, opened 2017-04-20
Made the default type readers publicly accessible 

danny-may:
Added method `public IDictionary<Type, TypeReader> CommandService.GetTypeReaders(Type type)` which gets all type readers for the given type, default readers included. This is intended to allow commands to use the built in type readers, rather than being restricted to the custom ones that can be added.

I also changed the naming of an existing method from `GetTypeReaders` to `GetCustomTypeReaders` and a private readonly variable from `_typeReaders` to `_customTypeReaders` to better fit what I understand their purpose to be with relation to the new function I added.


FiniteReality:
:-1: This is not how type readers are intended to be used.

PR #619, opened 2017-04-23
Restructure/Redesign Emotes

foxbot:
**This is a breaking change.**

This enhancement replaces the previous clunky emote pattern with a more versatile one.

As of this change, there is now one abstraction and three implementations, as follows:
```
IEmote
| - Emote
|   | - GuildEmote
|
| - Emoji
```

In this pattern, the IEmote is the most basic representation of an emote, encapsulating both custom emote objects, and emoji objects.

The Emote is a general emote (emotes being Discord's custom emojis, that can be uploaded on a per-guild basis), which is _detached_ from a guild. It is extended into the GuildEmote object, which also includes information about this emote while _attached_ to a guild, such as the roles allowed to use it, whether it is managed, etc.

The Emoji object is a basic structure encapsulating a unicode emoji. This object exists solely to be used in reactions where an emoji reaction is present, eliminating the need for the more sloppy string overload.


FiniteReality:
This change is great. It should make it more obvious that users need to pass *unicode* for their emotes instead of things like `ok_hand`. I'd suggest making the constructor for Emoji internal and using a static `FromUnicode` method to make this more obvious, but it's not necessary to implement.

PR #626, opened 2017-04-25
Use implicit package references

Auralytical:
As per https://github.com/dotnet/standard/issues/71, implicit NETStandard.Library references should be preferred over individual packages.


Auralytical:
It sounds like package management is being improved in NetStandard.Library 1.6.2 and 2.0 and will no longer have the massive dependency list it currently does.

PR #628, opened 2017-04-27
Clean up command result logic

FiniteReality:
Actual implementation of the discussion in #632


src/Discord.Net.Commands/Results/TypeReaderResult.cs
```diff
@@ -40,14 +40,13 @@ private TypeReaderResult(IReadOnlyCollection<TypeReaderValue> values, CommandErr
 
         public static TypeReaderResult FromSuccess(object value)
```
Auralytical, line 41: Is this constructor ever called?

FiniteReality: I don't think it is, I'll remove it

EDIT: Actually, it's used in 7 places where we don't care about the score

PR #633, opened 2017-04-29
Add IChannel.Nsfw, RequireNsfw precondition

foxbot:
This change adds a NSFW property to Channels, to reflect Discord's new guidelines regarding NSFW channels.

This change also adds a RequireNsfw precondition, to allow bots with NSFW commands to easily adjust to Discord's new rules regarding NSFW content outside of an NSFW channel.


src/Discord.Net.Core/Entities/Channels/IChannel.cs
```diff
@@ -8,6 +8,9 @@ public interface IChannel : ISnowflakeEntity
         /// <summary> Gets the name of this channel. </summary>
         string Name { get; }
 
+        /// <summary> Checks if the channel is NSFW. </summary>
+        bool Nsfw { get; }
```
Auralytical: IsNsfw would be better here

src/Discord.Net.Core/Utils/NsfwUtils.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Discord
+{
+    public static class NsfwUtils
```
Auralytical: Do we need a public class for this? Why not just add to ChannelHelper?

src/Discord.Net.Rest/Entities/Channels/ChannelHelper.cs
```diff
@@ -289,5 +289,10 @@ private static IUser GetAuthor(BaseDiscordClient client, IGuild guild, UserModel
                 author = RestUser.Create(client, guild, model, webhookId);
             return author;
         }
+
+        public static bool IsNsfw(IChannel channel) =>
+            IsNsfw(channel.Name);
+        public static bool IsNsfw(string channelName) =>
```
Auralytical, line 295: I think it's `channelName == "nsfw" || channelName.StartsWith("nsfw-")`

Issue #113, opened 2016-07-09
Ensure users are being removed from global cache


Auralytical:
This was never actually tested as far as I know - users should be removed from global cache when they are no longer in any visible guild, nor have a current dm channel.

RogueException:
With the removal of shared datastore support, I wonder if this system can be made much more solid using WeakReferences...

RogueException:
I recently patched one of these memory leaks as part the DownloadUsersAsync fix. We should check if this is still a problem.

khionu:
This can be tested as part of https://github.com/RogueException/Discord.Net/issues/272

PR #641, opened 2017-05-03
Fix for empty user objects after GUILD_MEMBER_REMOVE

### LINKED ISSUE #554: The user's username and discriminator in an unban event are not correct.
So I was using the unban event to log unbans, but it seems that the unbanned user's username is empty and the discriminator returns "0".

After some tests, if I interpolate the whole user, it returns the correct username and discriminator.

`$"{user.Username}#{user.Discriminator}";`
image

`$"{user}";`
image

LassieME:
Should fix #554 as well as improve guildmember counts.
Might be related to #113 and #272 though that is not tested.


Auralytical:
Am I missing something? Why was the GlobalUser.AddRef raised out of the user creation block?

PR #659, opened 2017-05-18
Add SocketRole.Members property

AntiTcb:


foxbot (approved):

FiniteReality (approved):

src/Discord.Net.WebSocket/Entities/Roles/SocketRole.cs
```diff
@@ -22,6 +25,8 @@ public class SocketRole : SocketEntity<ulong>, IRole
         public DateTimeOffset CreatedAt => SnowflakeUtils.FromSnowflake(Id);
         public bool IsEveryone => Id == Guild.Id;
         public string Mention => MentionUtils.MentionRole(Id);
+        public IReadOnlyCollection<SocketGuildUser> Members 
+            => Guild.Users.Where(x => x.Roles.Any(r => r.Id == Id)).ToImmutableArray();
```
foxbot: Replace with `.ToReadOnlyCollection`

foxbot (changes requested):

src/Discord.Net.WebSocket/Entities/Roles/SocketRole.cs
```diff
@@ -25,8 +24,8 @@ public class SocketRole : SocketEntity<ulong>, IRole
         public DateTimeOffset CreatedAt => SnowflakeUtils.FromSnowflake(Id);
         public bool IsEveryone => Id == Guild.Id;
         public string Mention => MentionUtils.MentionRole(Id);
-        public IReadOnlyCollection<SocketGuildUser> Members 
-            => Guild.Users.Where(x => x.Roles.Any(r => r.Id == Id)).ToImmutableArray();
+        public IEnumerable<SocketGuildUser> Members
```
khionu: Mutable collections should not be exposed. Why did you change it like this?

AntiTcb: Asked in the dev guild about it, post recommendation to change to `ToReadOnlyCollection()`. There are no backing field collections for this property that I can use for the fast `.Count` access, and any function to generate a count for this would be along the lines of
```
private int GetMemberCount()
        {
            int result = 0;
            for (int i = 0; i < Guild.Users.Count; i++)
            {
                var user = Guild.Users.ElementAt(i);
                if (user.Roles.Any(r => r.Id == Id))
                    result++;
            }
            return result;
        }
```
I asked for alternative ideas, fox spoke up stating there's no reason it *has* to be a readonlycollection. I'd like if it were for consistency's sake though.

foxbot: IEnumerable isn't a mutable collection..?

Joe4evr: Not to mention that you can't mutate the source collection from a LINQ query, anyway.

PR #661, opened 2017-05-20
Typo in the Nsfw Precondition Attribute

RealBlazeIt:


RealBlazeIt:
Oh, well it just sounds weird since I pronounce it as "Not Safe For Work" in my head, instead of saying the acronym. My bad

PR #670, opened 2017-05-25
Adds RequireAny, RequireAll PreconditionMode to Commands

LassieME:
Stole @AntiTcb idea from the dev guild since I had some time to kill, enables commands where only one precondition must be satisfied, aka: 
`[Command("echo", PreconditionsMode = PreconditionsMode.RequireAny), RequireOwner, RequireUserPermission(GuildPermission.Administrator)]
public async Task EchoCommand([Remainder] string input)` instead of having to have multiple commands pointing to the same method.

### DIFF
--- a/src/Discord.Net.Commands/Attributes/CommandAttribute.cs
+++ b/src/Discord.Net.Commands/Attributes/CommandAttribute.cs
@@ -7,6 +7,7 @@ namespace Discord.Commands
     {
         public string Text { get; }
         public RunMode RunMode { get; set; } = RunMode.Default;
+        public PreconditionsMode PreconditionsMode { get; set; } = PreconditionsMode.Default;
 
         public CommandAttribute()
         {
--- a/src/Discord.Net.Commands/Builders/CommandBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/CommandBuilder.cs
@@ -20,6 +20,7 @@ namespace Discord.Commands.Builders
         public string Remarks { get; set; }
         public string PrimaryAlias { get; set; }
         public RunMode RunMode { get; set; }
+        public PreconditionsMode PreconditionsMode { get; set; }
         public int Priority { get; set; }
 
         public IReadOnlyList<PreconditionAttribute> Preconditions => _preconditions;
@@ -65,6 +66,11 @@ namespace Discord.Commands.Builders
         public CommandBuilder WithRunMode(RunMode runMode)
         {
             RunMode = runMode;
+            return this;
+        }
+        public CommandBuilder WithPreconditionsMode(PreconditionsMode preconditionsMode)
+        {
+            PreconditionsMode = preconditionsMode;
             return this;
         }
         public CommandBuilder WithPriority(int priority)
--- a/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
@@ -128,26 +128,33 @@ namespace Discord.Commands
             
             foreach (var attribute in attributes)
             {
-                // TODO: C#7 type switch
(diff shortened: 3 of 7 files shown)


Joe4evr:
Is it really that hard for people to make a custom precondition to OR two (or more) preconditions together?

Not to mention that this breaks cases like "marker" preconditions that always return `FromSuccess()` but have other data that would be inspected for at runtime (such as my `[Hidden]`, or whatever [Willy Wonka here](https://cdn.discordapp.com/attachments/81384956881809408/316261661985996801/unknown.png) wants to do). At that point, this proposed enhancement becomes completely superfluous. Also, addons should then take this option in consideration _everywhere_.

Maybe that's just me, but I see too little value and more potential for large mistakes in this than just guiding users to [make a simple class of their own](https://gist.github.com/Joe4evr/275fdadc4848ba9755ae0e8d1e022920) that does the right thing from the start.

LassieME:
So your issue with this practically is that people are inheriting PreconditionAttribute for non-preconditions? To me that sounds like we need to have some way to expose some kind of non-precondition attributes in the CommandInfo. Also, to be clear, this PR is not a breaking change.

Joe4evr:
> To me that sounds like we need to have some way to expose some kind of non-precondition attributes in the CommandInfo.

Guess what? Many versions ago, you had access to the complete `MethodInfo` and `TypeInfo` objects. Then @FiniteReality added Command Builders and removed it. This is where #BlameFinite came from.

And it'll break for addons that rely on preconditions being ANDed and a user decided to set the default mode to `RequireAny` before the addon updates to use `RequireAll` explicitly.

FiniteReality:
I do have a PR open for adding custom attributes​ to commands: it might
need rebasing, but it should just work when it's merged.
As for command builders, since there is more than one way to build them
which doesn't necessarily require a methodinfo type, I decided that it was
better to not expose values which we don't necessarily have. I feel it is
similar in a way to how our interfaces are separated, versus how they were
in 0.9.
Additionally, I believe that the perfect solution would not require users
to know about the "internals" of how we wrap commands.

LassieME:
I assume the PR Finite is talking about is #458 , whats currently holding that back from being pulled?

@Joe4evr would you feel better if there was no way to set RequireAny as default, and you had to do it explicitly on every command you wanted it on?

Joe4evr:
It'd help, I guess. But I still don't think this is the best course of action. ¯\\\_(ツ)\_/¯

khionu:
I think the inverse might be more helpful: Preconditions marked as optional. This way, they can count as a pass, but they don't count as a fail, if they precondition isn't met. This allows preconditions that should be broken up to be done so.

What kind of behavior this encourages is a valid concern, so whether the PR is accepted or not is a matter of functionality vs designing around the user.

People who are determined will find a way to abuse anything added. I'm personally for addressing the concern, though not in the manner this PR does.

Joe4evr:
Now there's an idea. Actually, now that I think about it, rather than a boolean `IsOptional` flag, I think a *much* better option is introducing a settable number to define "groups" of preconditions that only need one to return success (except the default (0?) requires *all* preconditions met). To illustrate, take this (somewhat contrived) example:
```cs
[Command("echo"), RequireOwner(Group = 1)]
[RequireUserPermission(GuildPermission.Administrator, Group = 1)]
[CustomPreconditionFoo(Group = 2), CustomPreconditionBar(Group = 2)]
[CustomPreconditionBaz, CustomPreconditionQux] //default group
public async Task EchoCommand([Remainder] string input)
```
The resolution here would look like the expression
```
(owner || admin) && (foo || bar) && baz && qux
```
This would allow a lot more flexibility than a single flag *and* would be simpler to add to existing commands without too much hassle.

LassieME:
Joe's PR does everything this does + works on modules as well. Closing.

PR #672, opened 2017-05-25
Add grouping of preconditions to allow for flexible precondition logic.

Joe4evr:
Following the discussion in #670, I came up with this ingenious idea and I was asked to make the PR.

### DIFF
--- a/src/Discord.Net.Commands/Attributes/PreconditionAttribute.cs
+++ b/src/Discord.Net.Commands/Attributes/PreconditionAttribute.cs
@@ -7,12 +7,9 @@ namespace Discord.Commands
     public abstract class PreconditionAttribute : Attribute
     {
         /// <summary>
-        /// Specify a group that this precondition belongs to.
-        /// Preconditions of the same group require only one
-        /// of the preconditions to pass in order to be
-        /// successful (A || B). Specifying <see cref="Group"/> = 0
-        /// or not at all will require *all* preconditions
-        /// to pass, just like normal (A && B).
+        /// Specify a group that this precondition belongs to. Preconditions of the same group require only one
+        /// of the preconditions to pass in order to be successful (A || B). Specifying <see cref="Group"/> = 0
+        /// or not at all will require *all* preconditions to pass, just like normal (A && B).
         /// </summary>
         public int Group { get; set; } = 0;
 
--- a/src/Discord.Net.Commands/Attributes/PreconditionAttribute.cs
+++ b/src/Discord.Net.Commands/Attributes/PreconditionAttribute.cs
@@ -9,7 +9,7 @@ namespace Discord.Commands
         /// <summary>
         /// Specify a group that this precondition belongs to. Preconditions of the same group require only one
         /// of the preconditions to pass in order to be successful (A || B). Specifying <see cref="Group"/> = 0
-        /// or not at all will require *all* preconditions to pass, just like normal (A && B).
+        /// or not at all will require *all* preconditions to pass, just like normal (A &amp;&amp; B).
         /// </summary>
         public int Group { get; set; } = 0;
 
--- a/src/Discord.Net.Commands/Attributes/PreconditionAttribute.cs
+++ b/src/Discord.Net.Commands/Attributes/PreconditionAttribute.cs
@@ -8,10 +8,10 @@ namespace Discord.Commands
     {
         /// <summary>
         /// Specify a group that this precondition belongs to. Preconditions of the same group require only one
-        /// of the preconditions to pass in order to be successful (A || B). Specifying <see cref="Group"/> = 0
+        /// of the preconditions to pass in order to be successful (A || B). Specifying <see cref="Group"/> = <see cref="null"/>
         /// or not at all will require *all* preconditions to pass, just like normal (A &amp;&amp; B).
         /// </summary>
-        public int Group { get; set; } = 0;
+        public string Group { get; set; } = null;
 
         public abstract Task<PreconditionResult> CheckPermissions(ICommandContext context, CommandInfo command, IServiceProvider services);
     }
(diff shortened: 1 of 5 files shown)


foxbot (approved):

src/Discord.Net.Commands/Info/CommandInfo.cs
```diff
@@ -68,18 +68,48 @@ internal CommandInfo(CommandBuilder builder, ModuleInfo module, CommandService s
...
+                    foreach (PreconditionAttribute precondition in preconditionGroup)
+                        results.Add(await precondition.CheckPermissions(context, this, services).ConfigureAwait(false));
+
+                    if (!results.Any(p => p.IsSuccess))
+                        return results[0];
+                }
             }
 
-            foreach (PreconditionAttribute precondition in Preconditions)
+            foreach (IGrouping<int, PreconditionAttribute> preconditionGroup in Module.Preconditions.GroupBy(p => p.Group))
```
khionu: Checking the Modules Preconditions twice

khionu (changes requested):
Need to fix

src/Discord.Net.Commands/Info/CommandInfo.cs
Joe4evr: AH! Good catch.

khionu (approved):

Joe4evr:
When a non-zero group fails, should we emit a custom error message? Right now, I'm thinking
`$"(Module|Command) precondition group {preconditionGroup.Key} failed: {String.Join("\n", results.Select(r => r.ErrorReason))}."`

Anyone has a better suggestion, let me know.

src/Discord.Net.Commands/Attributes/PreconditionAttribute.cs
```diff
@@ -6,6 +6,13 @@ namespace Discord.Commands
     [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class, AllowMultiple = true, Inherited = true)]
     public abstract class PreconditionAttribute : Attribute
     {
+        /// <summary>
+        /// Specify a group that this precondition belongs to. Preconditions of the same group require only one
+        /// of the preconditions to pass in order to be successful (A || B). Specifying <see cref="Group"/> = 0
+        /// or not at all will require *all* preconditions to pass, just like normal (A && B).
```
FiniteReality: Minor nitpick: I'm not sure how `&&` will work with docfx considering this is badly formed XML; I'd suggest replacing with `&amp;&amp;` to ensure that it works best.

foxbot: Good catch^^ won't actually even compile

foxbot: Needs to be replaced with `&amp;&amp;`

foxbot (changes requested):

src/Discord.Net.Commands/Attributes/PreconditionAttribute.cs
Joe4evr: Ah, that might explain the way GitHub displays that. ***And*** that was what prevented VS from rendering the tooltip.

src/Discord.Net.Commands/Info/CommandInfo.cs
```diff
@@ -68,18 +68,48 @@ internal CommandInfo(CommandBuilder builder, ModuleInfo module, CommandService s
         {
             services = services ?? EmptyServiceProvider.Instance;
 
-            foreach (PreconditionAttribute precondition in Module.Preconditions)
+            foreach (IGrouping<int, PreconditionAttribute> preconditionGroup in Module.Preconditions.GroupBy(p => p.Group))
```
FiniteReality: I think refactoring the body of these foreaches into a local method would be best for maintainability and deduplicating the code

src/Discord.Net.Commands/Info/CommandInfo.cs
```diff
@@ -68,18 +68,48 @@ internal CommandInfo(CommandBuilder builder, ModuleInfo module, CommandService s
...
             {
-                var result = await precondition.CheckPermissions(context, this, services).ConfigureAwait(false);
-                if (!result.IsSuccess)
-                    return result;
+                if (preconditionGroup.Key == 0)
```
FiniteReality: I dislike the special-casing of group 0; I feel it may mislead users.

FiniteReality (changes requested):
After our discussion in the dev guild, I think using strings instead of ints will be safer and less misleading. (Especially with the special-casing of group 0!) If backwards-compatibility is important enough, you can have any non-null value represent a group, while the `null` group defaults to the old behaviour.

src/Discord.Net.Commands/Info/CommandInfo.cs
Joe4evr: ¯\\\_(ツ)_/¯ Going back through the chatlog, the last thing Volt said about this is `then talking about types re:perf is meaningless`. And I haven't seen her take a *definite* stance on the cosmetics yet, so I'll leave it like this for now.

src/Discord.Net.Commands/Results/PreconditionGroupResult.cs
```diff
@@ -0,0 +1,35 @@
...
+    [DebuggerDisplay(@"{DebuggerDisplay,nq}")]
+    public struct PreconditionGroupResult : IResult
+    {
+        public CommandError? Error { get; }
+
+        public string ErrorReason { get; }
+
+        public bool IsSuccess => !Error.HasValue;
+
+        public IEnumerable<PreconditionResult> Preconditions { get; }
```
FiniteReality: IMO naming this something like `PreconditionResults` instead would be more semantic. Additionally, it would probably make more sense as an IReadOnlyCollection rather than an IEnumerable.

foxbot (approved):

FiniteReality (approved):

Issue #675, opened 2017-05-27
Log a warning when loading a non-public module class


Joe4evr:
Filing mainly as a reminder.

We should log a warning [here](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs#L254) when a module class is not declared public. The check is as simple as `if (typeInfo.IsNotPublic)`, but because `ModuleClassBuilder` doesn't have access to a logger already *and* it's entirely `static`, implementation may require an overhaul of the class.

Confruggy:
huh, the methods of ModuleClassBuilder have the CommandService as arg which has a logger.

Joe4evr:
Oh does it? Didn't notice that, my bad.

FiniteReality:
I'll cover this in #678 - it wouldn't be hard to check in AddModulesAsync for any types that have possibly been skipped due to not being public.

Joe4evr:
👌

PR #676, opened 2017-05-27
Log warning on non-public modules

Joe4evr:
As per #675. Will take suggestions for the log message if anyone has a better one.


Joe4evr:
@FiniteReality pointed out that this is useless because `Assembly.ExportedTypes` already filters to public types. ¯\\\_(ツ)_/¯

PR #678, opened 2017-05-28
Rewrite command and overload builders using C#7 features

FiniteReality:
Passes simple test cases and some more complicated ones, I would appreciate more testers though.

EDIT: I forgot to mention that this is *definitely* a breaking change.

### DIFF
--- a/src/Discord.Net.Commands/Builders/CommandBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/CommandBuilder.cs
@@ -8,22 +8,17 @@ namespace Discord.Commands.Builders
 {
     public class CommandBuilder
     {
-        private readonly List<PreconditionAttribute> _preconditions;
-        private readonly List<ParameterBuilder> _parameters;
+        private readonly List<OverloadBuilder> _overloads;
         private readonly List<string> _aliases;
 
         public ModuleBuilder Module { get; }
-        internal Func<ICommandContext, object[], IServiceProvider, Task> Callback { get; set; }
 
         public string Name { get; set; }
         public string Summary { get; set; }
         public string Remarks { get; set; }
         public string PrimaryAlias { get; set; }
-        public RunMode RunMode { get; set; }
-        public int Priority { get; set; }
 
-        public IReadOnlyList<PreconditionAttribute> Preconditions => _preconditions;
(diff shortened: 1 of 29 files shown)


src/Discord.Net.Commands/Builders/CommandBuilder.cs
```diff
@@ -8,43 +8,41 @@ namespace Discord.Commands.Builders
...
             Module = module;
 
-            _preconditions = new List<PreconditionAttribute>();
-            _parameters = new List<ParameterBuilder>();
+            _overloads = new List<OverloadBuilder>();
             _aliases = new List<string>();
         }
         //User-defined
-        internal CommandBuilder(ModuleBuilder module, string primaryAlias, Func<ICommandContext, object[], IServiceProvider, Task> callback)
+        internal CommandBuilder(ModuleBuilder module, string primaryAlias, Action<OverloadBuilder> defaultOverloadBuilder)
```
Joe4evr, line 33: Why go from a `Func<..., Task>` to an `Action<...>`? Just thinking out loud.

src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
```diff
@@ -196,23 +229,29 @@ private static void BuildParameter(ParameterBuilder builder, System.Reflection.P
             foreach (var attribute in attributes)
             {
                 // TODO: C#7 type switch
```
Auralytical: You left a TODO in

src/Discord.Net.Commands/Builders/CommandBuilder.cs
FiniteReality: The overload builder contains the callback, so it'd be something like this:

```
module.AddCommand("name", overload => {
    overload.WithCallback(/*callback here*/);
});
```

src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
FiniteReality: Whoops. At least it was a comment and not missing code :stuck_out_tongue:

FiniteReality:
After a discussion, we've decided it would be better to find another way to do this.

PR #683, opened 2017-05-30
Added/clarified some comments in structure.cs

Joe4evr:


docs/guides/getting_started/samples/intro/structure.cs
```diff
@@ -65,6 +65,13 @@ private static Task Logger(LogMessage message)
         }
         Console.WriteLine($"{DateTime.Now,-19} [{message.Severity,8}] {message.Source}: {message.Message}");
         Console.ForegroundColor = cc;
+        
+        // If you get an error saying 'CompletedTask' doesn't exist,
```
FiniteReality, line 69: IMO you should make this comment smaller. Something like
```cs
// If the following causes an error, target .NET Framework 4.6 or higher
```
~~Users who are specifically targeting .NET 4.5 will likely know about Task.Delay(0)~~ EDIT: Seems this was requested, nevermind.

docs/guides/getting_started/samples/intro/structure.cs
```diff
@@ -92,16 +99,16 @@ private static Task Logger(LogMessage message)
...
-        await _commands.AddModulesAsync(Assembly.GetEntryAssembly());
-        // Or add Modules manually if you prefer to be a little more explicit:
-        await _commands.AddModuleAsync<SomeModule>();
-
         // When all your required services are in the collection, build the container.
         // Tip: There's an overload taking in a 'validateScopes' bool to make sure
         // you haven't made any mistakes in your dependency graph.
         _services = _map.BuildServiceProvider();
 
+        // Either search the program and add all Module classes that can be found:
```
FiniteReality: I'd suggest placing a comment here specifying that modules have to be public if they are to be automatically loaded.

FiniteReality (changes requested):

FiniteReality (approved):

PR #689, opened 2017-06-05
Update commands with C#7 features

### LINKED ISSUE #264: Prioritise typed overloads in commands
Right now our command system has an issue with multiple overloads:

``` cs
[Command("a")]
public async Task OverloadedCommand(IUserMessage message, int argument)
[Command("a")]
public async Task OverloadCommand(IUserMessage message, string argument)
```

Since these overloads do not get prioritised based on what arguments were provided, the overload with the string parameter may get invoked instead of the overload with the integer parameter, but never the other way around.

One solution for this, as foxbot suggested, would be to add a 'Priority' attribute, to prioritise commands manually. Another solution would be to use the data provided by the type readers to choose the best overload to invoke.

### LINKED ISSUE #514: Wrong ExecuteAsync result with preconditions

### LINKED ISSUE #675: Log a warning when loading a non-public module class

FiniteReality:
This PR contains most of the contents of #678, but without the overload builders themselves.

I'm not sure about argument order, especially in places like [here](https://github.com/FiniteReality/Discord.Net/blob/ed51d3e84677469420ddc85c3b2e2d2fc3f89c6d/src/Discord.Net.Commands/Info/CommandInfo.cs#L88) and [here](https://github.com/FiniteReality/Discord.Net/blob/ed51d3e84677469420ddc85c3b2e2d2fc3f89c6d/src/Discord.Net.Commands/CommandParser.cs#L17), (CommandInfo line 88 and CommandParser line 17 respectively) but I have tested and all of these changes work successfully.

This is a breaking change for users who create custom TypeReaders or use builders, but should not affect anybody who does not use these custom features.

EDIT: Fixes #675, Fixes #264 and Fixes #514

### DIFF
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -309,7 +309,8 @@ namespace Discord.Commands
                 var argValuesScore = parseResult.ArgValues.Sum(x => x.Values.OrderByDescending(y => y.Score).FirstOrDefault().Score) / match.Command.Parameters.Count;
                 var paramValuesScore = parseResult.ParamValues.Sum(x => x.Values.OrderByDescending(y => y.Score).FirstOrDefault().Score) / match.Command.Parameters.Count;
 
-                return match.Command.Priority + argValuesScore + paramValuesScore;
+                var totalArgsScore = (argValuesScore + paramValuesScore) / 2;
+                return match.Command.Priority + totalArgsScore * 0.99f;
             }
 
             //Order the parse results by their score so that we choose the most likely result to execute
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -281,7 +281,7 @@ namespace Discord.Commands
 
             //If we get this far, at least one precondition was successful.
 
-            var parseResults = new Dictionary<CommandMatch, ParseResult>();
+            var parseResultsDict = new Dictionary<CommandMatch, ParseResult>();
             foreach (var pair in successfulPreconditions)
             {
                 var parseResult = await pair.Key.ParseAsync(context, searchResult, pair.Value, services).ConfigureAwait(false);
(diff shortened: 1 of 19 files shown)


src/Discord.Net.Commands/CommandService.cs
```diff
@@ -259,46 +259,78 @@ public Task<IResult> ExecuteAsync(ICommandContext context, int argPos, IServiceP
...
+            }
+
+            // Calculates the 'score' of a command given a parse result
+            float CalculateScore(CommandMatch match, ParseResult parseResult)
+            {
+                //TODO: is this calculation correct?
+                var argValuesScore = parseResult.ArgValues.Sum(x => x.Values.OrderByDescending(y => y.Score).FirstOrDefault().Score) / match.Command.Parameters.Count;
+                var paramValuesScore = parseResult.ParamValues.Sum(x => x.Values.OrderByDescending(y => y.Score).FirstOrDefault().Score) / match.Command.Parameters.Count;
+
+                return match.Command.Priority + argValuesScore + paramValuesScore;
```
Auralytical: [Priority = 1] Command(IUser user) <- Mention (1.0) = 2.0
[Priority = 2] Command(string arg) <- String (0.0) = 2.0
Which function is executed?

Auralytical: Possible Solution:
Merge argValuesScore + paramValuesScore in argValueScores (0.0 - 1.0)
Priority + argValuesScore * 0.99

PR #706, opened 2017-06-17
Add IUser#SendMessageAsync extension

foxbot:
Allows users to use `await user.SendMessageAsync()` instead of `await (await user.GetOrCreateDMChannelAsync()).SendMessageAsync()`


FiniteReality (commented):
Most of our internal code uses `ConfigureAwait(false)`, doesn't it?

PR #707, opened 2017-06-17
Add EmbedType enum

### LINKED ISSUE #667: Embed types should be an enum?
Currently a string.

This might be a breaking change, though I don't know that anyone even uses this value.

Valid types that I've seen:
- Rich
- Link

foxbot:
**This is a breaking change!**

Resolves #667 

Use a verbose EmbedType enum for an IEmbed's type field over a string.


FiniteReality (approved):

khionu:
Do we already have a JsonConverter to convert an Enum to String? Default, that will serialize to an integer. Also need to make sure that it serializes to lowercase.

PR #708, opened 2017-06-17
Add support for audit log reasons

foxbot:
**This is a breaking change!**
(But only if anyone actually uses RequestOptions)

A reason can now be set on kicks/bans, using the `reason` field that has been prepended to the RequestOptions parameter in their respective methods (making this essentially a non-breaking change).

A reason can also be added to the audit log for these actions via the `AuditLogReason` field in RequestOptions.

Examples:

```cs
[Command("ban")]
public async Task Ban(SocketUser target, [Remainder] string reason = null)
{
    await Context.Guild.AddBanAsync(target, 0, reason);
}
```
```cs
[Command("nick")]
public async Task Nick([Remainder] string nick)
{
    await Context.CurrentUser.ModifyAsync(x => x.Nickname = nick, RequestOptions.WithReason($"'nick' command as requested by {Context.User}"));
}
```

### DIFF
--- a/src/Discord.Net.Rest/DiscordRestApiClient.cs
+++ b/src/Discord.Net.Rest/DiscordRestApiClient.cs
@@ -803,7 +803,8 @@ namespace Discord.API
             options = RequestOptions.CreateOrClone(options);
 
             var ids = new BucketIds(guildId: guildId);
-            await SendAsync("PUT", () => $"guilds/{guildId}/bans/{userId}?delete-message-days={args.DeleteMessageDays}&reason={args.Reason}", ids, options: options).ConfigureAwait(false);
+            string reason = string.IsNullOrWhiteSpace(args.Reason) ? "" : $"&reason={args.Reason}";
+            await SendAsync("PUT", () => $"guilds/{guildId}/bans/{userId}?delete-message-days={args.DeleteMessageDays}{reason}", ids, options: options).ConfigureAwait(false);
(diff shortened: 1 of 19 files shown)


src/Discord.Net.Rest/DiscordRestApiClient.cs
```diff
@@ -803,7 +803,7 @@ internal virtual void Dispose(bool disposing)
             options = RequestOptions.CreateOrClone(options);
 
             var ids = new BucketIds(guildId: guildId);
-            await SendAsync("PUT", () => $"guilds/{guildId}/bans/{userId}?delete-message-days={args.DeleteMessageDays}", ids, options: options).ConfigureAwait(false);
+            await SendAsync("PUT", () => $"guilds/{guildId}/bans/{userId}?delete-message-days={args.DeleteMessageDays}&reason={args.Reason}", ids, options: options).ConfigureAwait(false);
```
FiniteReality: If we're already doing this at the header level, is there any point in adding the reason to the query string?

foxbot: @FiniteReality the audit log reason and ban/kick reason are actually decoupled in these two cases, AFAIK

FiniteReality: :thinking: Really? That's a bit weird.

PR #710, opened 2017-06-17
Change wording of permission preconditions

### LINKED ISSUE #704: RequireBotPermission and RequireUserPermission have the same error
The attributes for RequireBotPermission and RequireUserPermission have the same error message saying "Command requires permission" this should be changed to "Bot requires permission" and "User requires permission"

And yes i know that there are some issues already about reworking the command system.

FiniteReality:
Also fix an issue where RequireBotPermission may throw if used in a non-guild channel which required guild permissions.

EDIT: Fixes #704

### DIFF
--- a/src/Discord.Net.Commands/Attributes/Preconditions/RequireBotPermissionAttribute.cs
+++ b/src/Discord.Net.Commands/Attributes/Preconditions/RequireBotPermissionAttribute.cs
@@ -44,14 +44,16 @@ namespace Discord.Commands
 
         public override async Task<PreconditionResult> CheckPermissions(ICommandContext context, CommandInfo command, IServiceProvider services)
         {
-            var guildUser = await context.Guild.GetCurrentUserAsync();
+            IGuildUser guildUser = null;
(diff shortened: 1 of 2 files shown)


AntiTcb:
Shouldn't these also check permission overwrites for the permission as well?

FiniteReality:
They already do, IIRC.

AntiTcb:
So, throws occur if ChannelPermission.SendMessages is false due to not being able to send the error message to the context channel. Should this case be handled by trying to find a channel you can send a message to to report the precondition fail? Sending a DM to the caller also runs the risk of throwing if they don't accept DMs.

FiniteReality:
Informing the user of errors is the bot developer's problem; it's how they deal with the result we return.

foxbot (approved):

Joe4evr:
Should mention that this Fixes #704 ~~(should auto-close the issue when this gets merged in)~~.
(Actually, that may only be the case if it's part of the commit title. ¯\\\_(ツ)_/¯)

PR #711, opened 2017-06-18
Add various property validation in EmbedBuilder

moiph:


src/Discord.Net.Rest/Entities/Messages/EmbedBuilder.cs
```diff
@@ -8,14 +8,36 @@ public class EmbedBuilder
     {
         private readonly Embed _embed;
 
+        public const int MaxFieldCount = 25;
```
moiph, line 11: Any preference on using DiscordConfig for these consts?

moiph:
If preferred I can update all of the getters/setters to be consistent ( get => field  vs get { return field } ).  Usually I prefer to keep that type of cleanup isolated from functional changes.

Edit:  Making that change because I ended up touching all the URL properties.

foxbot (changes requested):
This PR doesn't account for the 2000/4000 total character limit placed on embeds (for users or bot accounts respectively). For the sake of simplicity I'm fine with just using the 4000 character bot limit, though to provide full coverage, maybe that should go as a precondition at the request level?

src/Discord.Net.Rest/Entities/Messages/EmbedBuilder.cs
```diff
@@ -8,14 +8,36 @@ public class EmbedBuilder
...
+        {
+            get => _embed.Description;
+            set
+            {
+                if (value?.Length > MaxDescriptionLength) throw new ArgumentException($"Description length must be less than or equal to {MaxDescriptionLength}.");
+                _embed.Description = value;
+            }
+        }
+
         public string Url { get { return _embed.Url; } set { _embed.Url = value; } }
```
AntiTcb: I can't help but feel like these would be better as `System.Uri`s rather than strings.

src/Discord.Net.Rest/Entities/Messages/EmbedBuilder.cs
```diff
@@ -149,8 +178,29 @@ public class EmbedFieldBuilder
...
-        public object Value { get { return _field.Value; } set { _field.Value = value.ToString(); } }
+        public const int MaxFieldNameLength = 256;
+        public const int MaxFieldValueLength = 1024;
+
+        public string Name
+        {
+            get => _field.Name;
+            set
+            {
+                if (value?.Length > MaxFieldNameLength) throw new ArgumentException($"Field name length must be less than or equal to {MaxFieldNameLength}.");
```
AntiTcb: Should probably also check field names/values for empty strings or nulls since those also throw back 400s.

moiph: Good call, agreed

src/Discord.Net.Rest/Entities/Messages/EmbedBuilder.cs
FiniteReality: I concur; using the Uri type means that we get validation of the input as an artefact of its implementation, and IMO the Uri class is nicer to work with.

moiph: Agreed I'm a fan of that change. So much for a single file edit :)

moiph:
Regarding the overall character limit -- it seems 4000 is not accurate.  So I'm intentionally excluding that for now until we can get a clearer idea of what those limits are.

FiniteReality (approved):

FiniteReality (changes requested):
The ArgumentException instances should be passed the property name too, I missed this initially.

FiniteReality (approved):

foxbot (approved):

src/Discord.Net.Core/Entities/Messages/Embed.cs
```diff
@@ -56,6 +57,8 @@ internal Embed(string type)
             Fields = fields;
         }
 
+        public int Length => Title?.Length + Author?.Name?.Length + Description?.Length + Footer?.Text?.Length + Fields.Sum(f => f.Name.Length + f.Value.ToString().Length) ?? 0;
```
moiph, line 60: This assumes only text properties count toward the limit; if URLs count then obviously this needs an update :)

PR #712, opened 2017-06-19
Update events.cs, clarify enabling message caching.

AntiTcb:
Gladly taking suggestions for better comments.


docs/guides/concepts/samples/events.cs
```diff
@@ -8,7 +8,11 @@ public class Program
 	
 	public async Task MainAsync()
 	{
-		_client = new DiscordSocketClient();
+		// When working with events that have Cacheable<IMessage, ulong> parameters,
+		// you must enable the message cache in your config settings if you plan to
+		// use the cached message entity. 
+		_config = new DiscordSocketConfig { MessageCacheSize = 100 };
```
foxbot: _config isn't defined anywhere

foxbot (changes requested):

PR #716, opened 2017-06-23
Added EmbedBuilderExtensions for Uri parsing

Confruggy:
Because some people suggested I should PR this. Any thoughts?

### DIFF
--- a/src/Discord.Net.Rest/Extensions/EmbedBuilderExtensions.cs
+++ b/src/Discord.Net.Rest/Extensions/EmbedBuilderExtensions.cs
@@ -1,0 +1,25 @@
++using System;
++
++namespace Discord
++{
++    public static class EmbedBuilderExtensions
++    {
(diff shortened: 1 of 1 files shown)


khionu:
Two issues:
1) This conceals a failed parsing. That is not ok
2) [This CTOR](https://msdn.microsoft.com/en-us/library/z6c2z492(v=vs.110).aspx) makes it extremely easy, as is
```cs
embedBuilder.WithUrl(new Uri("https://google.com"));
```

I don't see a good reason to do this. That said, `HttpClient` methods accept a string or Uri, and I think it would be reasonable to do the same, as a proper change.

FiniteReality:
Did you like copy a diff or something? You've got double plusses, and git only ignores one which makes a syntax error.

Confruggy:
Honestly I have no Idea how the double plusses got there. I'm not that good with git tho.

Auralytical:
Made unnecessary by https://github.com/RogueException/Discord.Net/pull/724

PR #718, opened 2017-06-24
Add 'tweet' embed type

FiniteReality:
Paulo in Discord:
```
Error converting value "tweet" to type 'Discord.EmbedType'. Path '[47].embeds[0].type', line 1, position 37232. System.ArgumentException: Requested value 'tweet' was not found.
```

### DIFF
--- a/src/Discord.Net.Core/Entities/Messages/EmbedType.cs
+++ b/src/Discord.Net.Core/Entities/Messages/EmbedType.cs
@@ -7,6 +7,7 @@
         Video,
         Image,
         Gifv,
-        Article
+        Article,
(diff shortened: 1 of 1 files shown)


src/Discord.Net.Core/Entities/Messages/EmbedType.cs
```diff
@@ -7,6 +7,7 @@ public enum EmbedType
         Video,
         Image,
         Gifv,
-        Article
+        Article,
+        Tweet
```
moiph, line 11: I like trailing commas it makes diffs 1 line shorter when adding new lines

PR #719, opened 2017-06-25
Audit Log support

FiniteReality:
All but the invite update action is implemented, to my knowledge. Feedback on the design and current implementation would be appreciated, plus more testing outside of the simple validation I did using Discord API's audit logs.

So, all I really need to do is:
- [ ] Implement InviteUpdateAuditLogData
- [ ] Rewrite some data classes to use read-only objects
- [ ] Replace most occurences of ChannelId with IChannel somehow
- [ ] Finish the few properties of GuildUpdateAuditLogData which are missing


ArcanoxDragon:
I know this feature was complete a number of months ago, but would it be possible to support custom Audit Log entries? I'm not sure if this is even supported by Discord, but from what I can tell in the commits in this PR, the Discord.Net Audit Log support is read-only.

It would be useful to be able to push custom Audit Log entries from bots if such behavior is supported by Discord. My bot allows users in certain roles (moderators) to promote guests to a certain higher rank (allowing them write access in other channels), and I would like to audit these actions if possible.

asm512:
image
This is about the most customisable you can get, there's a parameter which allows for reasons : / , entire custom entries would make for a bit of a mess I think, not sure which you meant

edit @foxbot  sorry I clicked on this before you commented

ArcanoxDragon:
So I guess my only option is to add a reason when the bot updates the role of the user in question. I was hoping for some sort of customized `<so-and-so> promoted user <whomever> to User`, but I think the reason on the "updated roles" action should suffice.

asm512:
Uhm no , not exactly, adding a role has doesn't have a  ```reason``` overload,  only ```ban``` and ```kick``` have this option , if you look at the commit desc. on #708 you'll see

peaches6:
I'd suggest, since Audit Logs I'm assuming aren't even visible to everyone , I'd make a channel for only the bot, where you can have it send promotion messages and have it send ```<so-and-so> promoted <so-and-so> to <role>``` to that respective channel

ArcanoxDragon:
Adding a role has a reason on the version that I'm running...I just tested it and it works (reason showed up in the audit action from the bot). I don't need everybody to see it; I just want to know which users are promoting through the bot since the actual audit action only shows the bot doing it.

asm512:
Oh OK sorry my bad , from my glance at the commit desc. I gathered that this was only available for ban & kick, if, as you say it has a reason that's your problem solved right?

ArcanoxDragon:
Yeah. Not _quite_ in the way I hoped, but it'll do.

Issue #721, opened 2017-06-26
Multi use commands return the wrong error if you provide wrong arguments


aStonePenguin:
If you have a command with multiple types of arguments it always returns CommandError.UnknownCommand when you provide the wrong arguments.

Code:
![error](http://i.imgur.com/PF6FCzs.png)

Joe4evr:
Well..... Isn't that kind of expected? If neither argument could be parsed, the CommandService doesn't know which overload was intended to be used, so `UnknownCommand` is the best Error status available.

I *think* @FiniteReality has a PR in the works that makes overloads better in this regard (not sure if something like `CommandError.OverloadResolutionFailed` is already set to be added, but if it isn't, here's the idea).

khionu:
`CommandError.UnknownCommand` isn't referenced anywhere in the Command Parser, so it's not returning that during the parsing. Can you use a breakpoint to see what commands are loaded into your Command Service? Just to make sure.

foxbot:
https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Commands/CommandService.cs#L301

This was already discussed and identified as a 'bug' in the way we handle command overloads.

FiniteReality:
@Joe4evr In this case, we'd return the overload which was most likely to be the one the user wanted, which would be decided using `PriorityAttribute`.

PR #722, opened 2017-06-27
Better "solution" for #721

aStonePenguin:
It's unanimously encouraged to completely ignore `CommandError.UnknownCommand` while in this instance that does not necessarily apply. It doesn't really make a lot of sense to reuse the same enum anyway.


FiniteReality:
This is not a good solution: overloads are stored as commands, so it makes no sense to return a different result. Additionally, this encourages bad behaviour as the result returned does not necessarily suggest that the overload was unknown. Something that would actually be a "better solution" would be to wait for #689.

PR #724, opened 2017-06-28
Revert change of all Url types on IEmbed to string

foxbot:
This change reverts the impact on usability in #711 where types of URL on an IEmbed were changed to `System.Uri`. 

The functionality of validating URLs is still left intact, by manually validating URLs on `EmbedBuilder`


FiniteReality:
For 1.0, leaving types as string makes sense. Maybe switching to Uri would make more sense as a 1.1 thing?

FiniteReality (approved):

PR #734, opened 2017-06-30
Fix potential nullref in embedBuilder value setter

moiph:


Auralytical:
Collapse into a string extension method instead of repeating the same logic

src/Discord.Net.Core/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace Discord
+{
+    public static class StringExtensions
```
Auralytical: Make internal

PR #735, opened 2017-07-01
Add back the case for ParameterPreconditions

### LINKED ISSUE #732: ParameterPreconditions aren't working as of 1.0.0
I noticed this while upgrading my project from 1.0.0-rc2 (Nuget Prerelease) to 1.0.0 "Final" (also Nuget). It seems parameter preconditions simply never run. 

Here is a simple repro, which I used in conjunction with the [Command Service example in the documentation](https://discord.foxbot.me/docs/guides/commands/commands.html).

```csharp
public class TestModule : ModuleBase
{
    [Command("test")]
    public async Task DoTest([TestPrecon]string input)
    {
        //this should never happen
        await ReplyAsync("Test Succeeded");
    }
}

public class TestPrecon : ParameterPreconditionAttribute
{
    public override async Task<PreconditionResult> CheckPermissions(ICommandContext context, ParameterInfo parameter, object value, IServiceProvider services)
    {
        //do some async work
        var ownerId = (await services.GetService<DiscordSocketClient>().GetApplicationInfoAsync()).Owner.Id;

        //always return an error
        return PreconditionResult.FromError("Failure.");
    }
}
```

With this code, "Test Succeeded" is always sent, regardless of the PreconditionResult being an error result. None of the code within `CheckPermissions` is executed.

Joe4evr:
Fixes #732

### DIFF
--- a/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
@@ -251,6 +251,9 @@ namespace Discord.Commands
                         builder.IsMultiple = true;
                         paramType = paramType.GetElementType();
                         break;
+                    case ParameterPreconditionAttribute precon:
+                        builder.AddPrecondition(precon);
+                        break;
                     case RemainderAttribute _:
                         if (position != count - 1)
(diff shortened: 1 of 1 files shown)


FiniteReality (approved):
Looks like I fucked up with git and missed this in the rebase. :upside_down_face:

PR #738, opened 2017-07-04
Add build prefix in CI rather than in project files

### LINKED ISSUE #728: CI builds are missing their suffix
See [here](https://ci.appveyor.com/project/RogueException/discord-net/build/build-785-swqvielo) for an example of a failing build.

Volt says our targets logic may be wrong, and since I'm not a msbuild expert opening this issue sounded like the best idea so that somebody can fix it.

FiniteReality:
Fixes #728 by altering the build script so that it adds the build prefix rather than in the project config.

This likely means more flexibility as it is easier to define build prefixes in CI than it is in the targets and nuspec file.

### DIFF
--- a/Discord.Net.targets
+++ b/Discord.Net.targets
@@ -1,7 +1,6 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
     <PropertyGroup>
         <VersionPrefix>1.0.2</VersionPrefix>
-        <VersionSuffix></VersionSuffix>
         <Authors>RogueException</Authors>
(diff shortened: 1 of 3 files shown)


Discord.Net.targets
```diff
@@ -1,21 +1,18 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
     <PropertyGroup>
-        <VersionPrefix>1.0.1</VersionPrefix>
-        <VersionSuffix></VersionSuffix>
+        <VersionPrefix>1.0.2</VersionPrefix>
```
Auralytical, line 3: Nope.

PR #744, opened 2017-07-10
Added method to flatten paged content into one async collection

ObsidianMinor:
Now you can use LINQ with your async collections. Have fun.


ObsidianMinor:
This could wait until C# 8 so I can replace the PagedCollectionEnumerator<T> with just a yield return

Still34:
Wouldn't the release take another year or so?

ObsidianMinor:
Discord.Net 2.x could also take another year. So it depends on if both C# 8 and the next major version are released near each other I guess.

Still34:
I don't think 2.0 will take that long to release, as 2.0 includes some breaking changes that need to be pushed ASAP. 3.0 would make more sense for integrating C#8 changes.

PR #745, opened 2017-07-10
Added relationship features

ObsidianMinor:
We're working on 1.1, right?

New methods
 * IDiscordClient.GetRelationshipsAsync(RequestOptions options = null);
 * IUser.AddFriendAsync(RequestOptions options = null);
 * IUser.BlockUserAsync(RequestOptions options = null);
 * IUser.RemoveRelationshipAsync(RequestOptions options = null);

New events
 * DiscordSocketClient.RelationshipAdded
 * DiscordSocketClient.RelationshipRemoved


Joe4evr:
> New events
> * DiscordSocketClient.RelationshipAdded
> * DiscordSocketClient.RelationshipRemoved

Shouldn't these be added to `DiscordShardedClient` as well?

src/Discord.Net.Rest/ClientHelper.cs
```diff
@@ -152,5 +152,11 @@ internal static class ClientHelper
             var models = await client.ApiClient.GetVoiceRegionsAsync(options).ConfigureAwait(false);
             return models.Select(x => RestVoiceRegion.Create(client, x)).FirstOrDefault(x => x.Id == id);
         }
+
+        public static async Task<IReadOnlyCollection<RestRelationship>> GetRelationshipsAsync(BaseDiscordClient client)
+        {
+            var models = await client.ApiClient.GetRelationshipsAsync().ConfigureAwait(false);
+            return models.Select(r => RestRelationship.Create(client, r)).ToImmutableArray();
```
AntiTcb: AFAIK, `ToReadOnlyCollection()` is preferred when you have a collection, instead of `ToImmutableArray()`.

AntiTcb:
- I take it that using any of these endpoints as a non-user/to a non-user throws some manner of 400 HTTP error. These should probably be intercepted before the API is hit with the request, to stay consistent with what is done with bulk deleting/embed validation.
- Is there a major reason `AddFriendAsync`/`BlockUserAsync`/`RemoveRelationshipAsync` aren't implemented as ClientHelper methods and instead use the ApiClient directly?

ObsidianMinor:
@AntiTcb 
 * Yes, however I don't know how bulk deletes intercept so you'll have to point that out
 * Not really, I never thought about it because it was a one-line method

AntiTcb:
>you'll have to point that out

https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Rest/DiscordRestApiClient.cs#L539
It's just done as a precondition within the ApiClient.

ObsidianMinor:
Added in https://github.com/RogueException/Discord.Net/pull/745/commits/985268f7778af18172e2475dd2ec351d7b7290fc

Rapptz:
I hope you guys are aware these endpoints are blacklisted and will just get your user account unverified.

PR #747, opened 2017-07-10
Added CommandService.CommandExecuted

foxbot:
This change allows developers to more easily access command results, without needing to search the text attached to the CommandService's Log event.

Commands with an async RunMode can now return a RuntimeResult.


FiniteReality (approved):
It's a shame we can't consolidate the logic in CommandInfo.ExecuteAsync. I'll have to take a look into this in the future.

khionu (approved):

Issue #741, opened 2017-07-04
Ensure multiple CHANNEL_CREATEs in a DM does not break the cache


foxbot:
See: https://github.com/hammerandchisel/discord-api-docs/issues/184#issuecomment-312960055

Auralytical:
Fixed in https://github.com/RogueException/Discord.Net/commit/f9970891742775253b453a2b8532dbb7b2f5c7d1

PR #751, opened 2017-07-13
Try to pull DM channels from cache on CHANNEL_CREATE

foxbot:
This resoles #741

Since Discord now dispatches a CHANNEL_CREATE for every message sent to
a bot, we check to see if a channel has already been added to the state
before creating a new one.

### DIFF
--- a/src/Discord.Net.WebSocket/DiscordSocketClient.cs
+++ b/src/Discord.Net.WebSocket/DiscordSocketClient.cs
@@ -1631,7 +1631,10 @@ namespace Discord.WebSocket
 
         internal ISocketPrivateChannel AddPrivateChannel(API.Channel model, ClientState state)
         {
-            var channel = SocketChannel.CreatePrivate(this, state, model);
+            ISocketPrivateChannel channel;
+            if ((channel = state.GetChannel(model.Id) as ISocketPrivateChannel) != null)
+                return channel;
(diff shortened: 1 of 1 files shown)


src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -1631,7 +1631,10 @@ internal SocketGuild RemoveGuild(ulong id)
 
         internal ISocketPrivateChannel AddPrivateChannel(API.Channel model, ClientState state)
         {
-            var channel = SocketChannel.CreatePrivate(this, state, model);
+            ISocketPrivateChannel channel;
+            if ((channel = state.GetChannel(model.Id) as ISocketPrivateChannel) != null)
```
Joe4evr: 🤔 Literally doing a longer form of `if (state.GetChannel(model.Id) is ISocketPrivateChannel channel)`

foxbot: it hadn't come to mind when I first wrote this, since I was really more concerned with checking whether the method would return null, not if the cast was valid

Auralytical:
Moved this logic up a level ( https://github.com/RogueException/Discord.Net/commit/f9970891742775253b453a2b8532dbb7b2f5c7d1 ) so it avoids raising the ChannelCreated event too

PR #752, opened 2017-07-14
Update client.cs

AntiTcb:
Let's not have the client be a local variable, hm?


docs/guides/getting_started/samples/intro/complete.cs
```diff
@@ -7,12 +7,14 @@ namespace MyBot
 {
 	public class Program
 	{
+		private DiscordSocketClient client;
```
foxbot: Why is this one `client` but the first one was `_client`

AntiTcb: Good question.

PR #755, opened 2017-07-15
Fix everyone mention.

### LINKED ISSUE #754: guild.EveryoneRole.Mention returns incorrect string
Calling Mention on the everyone role returns `<@&SERVERID>` instead of `@everyone`.
While the everyone role's ID is the same as the server's, and <@&ID> is otherwise how you mention a role, it's not how you mention the everyone role. Mentioning the everyone role is literally `@everyone`.

<@&SERVERID> will produce an inexplicable non-existent phantom `@@everyone` role that doesn't highlight anyone.

AntiTcb:
Resolves #754.


foxbot (approved):

ObsidianMinor:
Damn that was fast

AntiTcb:
More than glad to change implementation if IEveryoneRole is the agreed upon approach.

PR #757, opened 2017-07-16
Disable ratelimits on reactions

foxbot:
This pull request generates a random channel ID to be passed into the ratelimit bucket when adding a reaction.

This should avoid any preemptive ratelimiting on the AddReactionAsync endpoint.

This change is necessary because right now it takes our library seven seconds to generate seven reactions, whereas competing libraries (Discord4J, DiscordNim) can do it in just under two seconds.

### DIFF
--- a/src/Discord.Net.WebSocket/DiscordSocketClient.cs
+++ b/src/Discord.Net.WebSocket/DiscordSocketClient.cs
@@ -1631,7 +1631,10 @@ namespace Discord.WebSocket
 
         internal ISocketPrivateChannel AddPrivateChannel(API.Channel model, ClientState state)
         {
-            var channel = SocketChannel.CreatePrivate(this, state, model);
+            ISocketPrivateChannel channel;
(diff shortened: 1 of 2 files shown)


src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -1631,7 +1631,9 @@ internal SocketGuild RemoveGuild(ulong id)
 
         internal ISocketPrivateChannel AddPrivateChannel(API.Channel model, ClientState state)
         {
-            var channel = SocketChannel.CreatePrivate(this, state, model);
+            if (state.GetChannel(model.Id) is ISocketPrivateChannel channel)
```
foxbot, line 1634: Please disregard this change I forgot to change branches.

LassieME:
Any reason for this to not be an option in either the config or the RequestOptions or both? I feel like that was the consensus when it was discussed in the dev guild months ago.

PR #758, opened 2017-07-17
NullOrEmpty -> NullOrWhiteSpace

SapiensAnatis:
Recently, it seems that checks have been added to this EmbedBuilder file to better inform users what's wrong with an invalid embed they might be sending, ahead of a BadRequest error that Discord will return.

However, I believe that Discord will also throw if you try and add an embed field where the value (and probably the title as well, though this is only an assumption) is just whitespace. If anyone could repro this fact in a more controlled environment that'd be great.

To make these 'pre-exceptions' more consistent, I changed NullOrEmpty to NullOrWhiteSpace, and updated the exception message to reflect it.

### DIFF
--- a/src/Discord.Net.Rest/Entities/Messages/EmbedBuilder.cs
+++ b/src/Discord.Net.Rest/Entities/Messages/EmbedBuilder.cs
@@ -249,7 +249,7 @@ namespace Discord
             get => _field.Name;
             set
             {
-                if (string.IsNullOrEmpty(value)) throw new ArgumentException($"Field name must not be null or empty.", nameof(Name));
+                if (string.IsNullOrWhiteSpace(value)) throw new ArgumentException($"Field name must not be null, empty or entirely whitespace.", nameof(Name));
                 if (value.Length > MaxFieldNameLength) throw new ArgumentException($"Field name length must be less than or equal to {MaxFieldNameLength}.", nameof(Name));
                 _field.Name = value;
             }


Joe4evr:
I'm pretty sure it's possible to have fields with a value of the Non-Breaking Space `U+00A0` or the Zero-Width Space `U+200B`, which are the work-arounds for users wanting an "empty" field.

SapiensAnatis:
I am getting a BadRequest with the following snippet (the BadRequest occurs later when I try and send, but still):

```csharp
string CodePoint = "00A0";
int Code = int.Parse(CodePoint, System.Globalization.NumberStyles.HexNumber);
string CharString = char.ConvertFromUtf32(Code);
ResultEmbed.AddField("test", CharString);
```

However, `U+200B` does work if I set that as CodePoint. Additionally, a string with that character alone returns False on IsNullOrWhiteSpace.

Furthermore, even though `U+00A0` fails, it (conveniently) returns True on IsNullOrWhiteSpace. The check seems sound to me. Maybe Discord is using something very similar to this on their end 👀

Joe4evr:
That has to be **the most** convoluted way of writing a Unicode codepoint I've ever seen. You know you could've just written `string nbsp = "\u00A0";`, right? 👀

Or *at least* know that Hex-literals exist: `char.ConvertFromUtf32(0x00A0)`

SapiensAnatis:
Oh well, live and learn. All the same - doing it your way gives me a BadRequest error. Not that I think you were insinuating anything else would happen - I just saw this as a chance to test it in a more sterile manner.

```csharp
[Command("badembed")]
public async Task badembed()
{
    string BadCharacter = "\u00A0";
    int x = BadCharacter[0];
    var msg = await ReplyAsync($"{x.ToString("X4")}: {string.IsNullOrWhiteSpace(BadCharacter)}");
    await msg.ModifyAsync(m => m.Embed = new EmbedBuilder().AddField(BadCharacter, "Hello!").Build());
}
```
That throws a BadRequest for me. The bot's message, before it throws, was: `00A0: True`

PR #759, opened 2017-07-17
Allow users to opt out of preemtive rate limiting

foxbot:
This change adds a property, `BypassBuckets` to RequestOptions.

When this property is set to true, the global ratelimit, and route-specific ratelimit buckets will be bypassed entirely. Users can now send requests at their intended speed, instead of being crippled by the broken ratelimit headers.

Example

### DIFF
--- a/src/Discord.Net.Core/RequestOptions.cs
+++ b/src/Discord.Net.Core/RequestOptions.cs
@@ -15,6 +15,11 @@ namespace Discord
         public RetryMode? RetryMode { get; set; }
         public bool HeaderOnly { get; internal set; }
         /// <summary>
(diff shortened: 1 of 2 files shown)


Nensec:
What is stopping this from being merged since januari?

jianmingyong:
Man I NEED this. (Sorry for unrelevant comment but I need a faster reaction speed)

oppahansi:
like Nensec already asked

> What is stopping this from being merged since januari?

FiniteReality:
It's a very dangerous change, and I disagree with it entirely.

HelpfulStranger999:
Maybe it would be better to allow an opt out for reactions only since that's the only common use case for this

PR #761, opened 2017-07-17
Update README.md

ghost:
Come on guys, you're not even telling them which channel to look for.
What if they don't know :(.


Flamanis:
#roasted

PR #768, opened 2017-07-26
Fix CreateGuildAsync not sending icon stream.

AntiTcb:
Fix CreateGuildAsync not doing anything with the input stream for the guild icon.

Also fixes an issue with potential stream types that throw a `NotSupportedException` when checking its properties. [Apparently, they exist.](https://github.com/dotnet/corefx/blob/master/src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseStream.cs)


ghost:
+1.
Nice work Anti.

PR #771, opened 2017-07-27
Reworked IChannel.IsNsfw to support the new API flag

foxbot:
IChannel.IsNsfw will now return false when being used on any channel
that is not an ITextChannel. When being used on an ITextChannel, this
will now account for the API flag, and fall back to the channel name.

(this is gross design, thanks discord)

### DIFF
--- a/src/Discord.Net.Rest/Entities/Channels/RestChannel.cs
+++ b/src/Discord.Net.Rest/Entities/Channels/RestChannel.cs
@@ -46,7 +46,7 @@ namespace Discord.Rest
 
         //IChannel
         string IChannel.Name => null;
-        bool IChannel.IsNsfw => ChannelHelper.IsNsfw(this);
+        bool IChannel.IsNsfw => false;
 
         Task<IUser> IChannel.GetUserAsync(ulong id, CacheMode mode, RequestOptions options)
             => Task.FromResult<IUser>(null); //Overriden
--- a/src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs
+++ b/src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs
@@ -15,6 +15,8 @@ namespace Discord.Rest
 
         public string Mention => MentionUtils.MentionChannel(Id);
 
+        internal bool Nsfw { get; private set; }
+
         internal RestTextChannel(BaseDiscordClient discord, IGuild guild, ulong id)
             : base(discord, guild, id)
         {
@@ -30,6 +32,7 @@ namespace Discord.Rest
             base.Update(model);
 
             Topic = model.Topic.Value;
+            Nsfw = model.Nsfw.GetValueOrDefault();
         }
 
         public async Task ModifyAsync(Action<TextChannelProperties> func, RequestOptions options = null)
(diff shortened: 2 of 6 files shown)


src/Discord.Net.Rest/Entities/Channels/RestChannel.cs
```diff
@@ -46,7 +46,7 @@ internal static IRestPrivateChannel CreatePrivate(BaseDiscordClient discord, Mod
 
         //IChannel
         string IChannel.Name => null;
-        bool IChannel.IsNsfw => ChannelHelper.IsNsfw(this);
+        bool IChannel.IsNsfw => false;
```
FiniteReality, line 49: Why are we explicitly implementing IsNsfw? Users will just have to cast to IChannel to get access to this

src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs
```diff
@@ -15,6 +15,8 @@ public class RestTextChannel : RestGuildChannel, IRestMessageChannel, ITextChann
 
         public string Mention => MentionUtils.MentionChannel(Id);
 
+        internal bool Nsfw { get; private set; }
```
FiniteReality, line 18: Why internal? Additionally, IChannel.IsNsfw has a different behaviour to this; surely it may become confusing?

src/Discord.Net.WebSocket/Entities/Channels/SocketChannel.cs
```diff
@@ -41,7 +41,7 @@ internal static ISocketPrivateChannel CreatePrivate(DiscordSocketClient discord,
 
         //IChannel
         string IChannel.Name => null;
-        bool IChannel.IsNsfw => ChannelHelper.IsNsfw(this);
+        bool IChannel.IsNsfw => false;
```
FiniteReality, line 44: Again, it doesn't really make sense that we're only exposing IsNsfw on IChannel.

src/Discord.Net.WebSocket/Entities/Channels/SocketTextChannel.cs
```diff
@@ -16,6 +16,7 @@ public class SocketTextChannel : SocketGuildChannel, ITextChannel, ISocketMessag
         private readonly MessageCache _messages;
 
         public string Topic { get; private set; }
+        internal bool Nsfw { get; private set; }
```
FiniteReality, line 19: Again, this is implemented differently to IsNsfw, and is internal.

src/Discord.Net.Rest/Entities/Channels/RestChannel.cs
foxbot: This PR was originally developed as a patch on 1.0.x -- currently looking to merge this to fix #800. Exposing this property directly is currently being tracked on #794, and can be looked at later

src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs
foxbot: See above

FiniteReality (approved):
That reasoning makes sense. After both PRs are merged, somebody should make sure that the behaviour is consistent and that we are not duplicating code needlessly.

PR #773, opened 2017-07-27
Add BaseSocketClient object.

AntiTcb:
**Breaking changes:**

~~- Removed leftover `DiscordShardedClient.UserPresenceUpdated` event.~~
(Will be done on next major release)

**Additions:**
- Add BaseSocketClient object.
- Add ShardConnected, ShardDisconnected, ShardReady, and ShardLatencyUpdated events to DiscordShardedClient.
***Minor:***
-- Added optional [RequestOptions](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Core/RequestOptions.cs) argument that was missing from various methods.
-- Added docstrings for all events.

There is no currently existing bridge between [DiscordSocketClient](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.WebSocket/DiscordSocketClient.cs) and [DiscordShardedClient](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.WebSocket/DiscordShardedClient.cs), leaving users to duplicate all websocket client functionality between the two if they wish to abstract code that should be shared to a BCL, or for addon developers that wish to support both client types out of the box. 

While it makes logical sense that this wasn't implemented previously as DiscordShardedClient isn't technically a client itself, but a wrapper for multiple DiscordSocketClients, that comes at an unfortunate expense of usability for accessing events or methods only supported via websockets.

### DIFF
--- a/src/Discord.Net.Rpc/DiscordRpcClient.cs
+++ b/src/Discord.Net.Rpc/DiscordRpcClient.cs
@@ -67,6 +67,7 @@ namespace Discord.Rpc
 
         public Task StartAsync() => _connection.StartAsync();
         public Task StopAsync() => _connection.StopAsync();
+                        
 
         private async Task OnConnectingAsync()
         {
--- a/src/Discord.Net.Rpc/DiscordRpcClient.cs
+++ b/src/Discord.Net.Rpc/DiscordRpcClient.cs
@@ -22,7 +22,7 @@ namespace Discord.Rpc
 
         public ConnectionState ConnectionState { get; private set; }
         public IReadOnlyCollection<string> Scopes { get; private set; }
-        public DateTimeOffset TokenExpiresAt { get; private set; }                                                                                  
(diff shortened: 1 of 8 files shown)


src/Discord.Net.Rpc/DiscordRpcClient.cs
```diff
@@ -67,7 +67,7 @@ internal override void Dispose(bool disposing)
 
         public Task StartAsync() => _connection.StartAsync();
         public Task StopAsync() => _connection.StopAsync();
-
+
```
Joe4evr: Wat.

(Same goes for the other line in this file.)

Joe4evr:
Remove all that superfluous whitespace (`Ctrl+E, S` in VS to toggle its visibility).

AntiTcb:
Sounds good to me.

PR #790, opened 2017-08-12
Condense redundant AddField overloads in EmbedBuilder

MarkusG:
It seems redundant to have a `(string, string, bool)` overload when `EmbedFieldBuilder` takes an object anyway. I also don't know why the `(string, object)` overload has no parameter to specify inline where the `(string, string, bool)` one does.

### DIFF
--- a/src/Discord.Net.Rest/Entities/Messages/EmbedBuilder.cs
+++ b/src/Discord.Net.Rest/Entities/Messages/EmbedBuilder.cs
@@ -171,10 +171,10 @@ namespace Discord
             return this;
         }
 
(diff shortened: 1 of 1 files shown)


Nikey646:
Doesn't this make the [`AddInlineField`](https://github.com/MarkusGordathian/Discord.Net/blob/c01da2596d5cdfe3590c8a07d589b937439fc91e/src/Discord.Net.Rest/Entities/Messages/EmbedBuilder.cs#L183-L191) method redundant as well?

MarkusG:
Good point. I've removed `AddInlineField()` as well.

khionu (approved):

PR #801, opened 2017-08-27
Remove EmbedBuilder -> Embed implicit conversion

Joe4evr:
Literally every time someone in the support channel wants to know how to modify a message's embed, they ask why the compiler tells them `Cannot implicitly convert type 'Discord.EmbedBuilder' to 'Discord.Optional<Discord.Embed>'`. Reason being that the C# compiler doesn't do double implicit conversion.

Therefor I believe we should force users to always `.Build()` their embeds explicitly, so people can stop hitting this issue of seeming inconsistency.

### DIFF
--- a/src/Discord.Net.Rest/Entities/Messages/EmbedBuilder.cs
+++ b/src/Discord.Net.Rest/Entities/Messages/EmbedBuilder.cs
@@ -234,7 +234,6 @@ namespace Discord
 
             return _embed;
         }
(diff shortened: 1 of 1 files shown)


Auralytical (approved):
LGTM - I've always been a fan of doing conversions like this explicitly.

FiniteReality (approved):
I've never liked this implicit conversion in the first place; it's about time we decided to remove it.

PR #805, opened 2017-09-02
Fix typo in command setup example

Still34:


Still34:
Fix proposed in #826.

Issue #519, opened 2017-02-10
Abstract DiscordShardedClient's group connection lock


RheaAyase:
Description: Implement a system that could be used to `lock` both, connecting and reconnecting of the client.

Use case: Server cluster running one (or a few) shard(s) per hardware(or virtual) machine (or container)

=)

Auralytical:
We already have a system like this in place for DiscordShardedClient, just need to abstract it to support scenarios where shards are in different processes.

RheaAyase:
This should be in non-sharded client in my opinion, because the use case does not involve running multiple shards in one application. Single shard client should have synchronous event `i just started (re)connecting` in which one could halt and wait for whatever lock there needs be.

It is quite important event without which no large bot can function properly due to them "connecting" ratelimits.

PR #806, opened 2017-09-03
Added a `Connecting` event.

RheaAyase:
The purpose of this event is described in #519
_(Please excuse my IDE cutting off the trailing whitespace)_

### DIFF
--- a/src/Discord.Net.WebSocket/ConnectionManager.cs
+++ b/src/Discord.Net.WebSocket/ConnectionManager.cs
@@ -8,6 +8,8 @@ namespace Discord
 {
     internal class ConnectionManager
     {
+        public event Func<Task> Connecting { add { _connectingEvent.Add(value); } remove { _connectingEvent.Remove(value); } }
+        private readonly AsyncEvent<Func<Task>> _connectingEvent = new AsyncEvent<Func<Task>>();
         public event Func<Task> Connected { add { _connectedEvent.Add(value); } remove { _connectedEvent.Remove(value); } }
         private readonly AsyncEvent<Func<Task>> _connectedEvent = new AsyncEvent<Func<Task>>();
         public event Func<Exception, bool, Task> Disconnected { add { _disconnectedEvent.Add(value); } remove { _disconnectedEvent.Remove(value); } }
@@ -123,6 +125,8 @@ namespace Discord
 
             try
             {
+                await _connectingEvent.InvokeAsync().ConfigureAwait(false);
+
                 var readyPromise = new TaskCompletionSource<bool>();
                 _readyPromise = readyPromise;
 
--- a/src/Discord.Net.WebSocket/ConnectionManager.cs
+++ b/src/Discord.Net.WebSocket/ConnectionManager.cs
@@ -16,8 +16,8 @@ namespace Discord
         private readonly SemaphoreSlim _stateLock;
         private readonly Logger _logger;
         private readonly int _connectionTimeout;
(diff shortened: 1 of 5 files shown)


src/Discord.Net.WebSocket/ConnectionManager.cs
```diff
@@ -26,14 +28,14 @@ internal class ConnectionManager
...
-            Func<Task> onConnecting, Func<Exception, Task> onDisconnecting, Action<Func<Exception, Task>> clientDisconnectHandler)
+        internal ConnectionManager(SemaphoreSlim stateLock, Logger logger, int connectionTimeout,
+            Func<Task> connect, Func<Exception, Task> disconnect, Action<Func<Exception, Task>> clientDisconnectHandler)
         {
             _stateLock = stateLock;
             _logger = logger;
             _connectionTimeout = connectionTimeout;
-            _onConnecting = onConnecting;
-            _onDisconnecting = onDisconnecting;
+            this._connect = connect;
```
khionu: To be consistent with the conventions used in this library, `this` is not used, in favor of prefixing private fields with `_`. Ergo, the use of `this` is redundant.

khionu (changes requested):

src/Discord.Net.WebSocket/ConnectionManager.cs
RheaAyase: Wow really, I wouldn't have guessed...

(_I see that your superior treatment of others haven't changed in a year. You should really try to be more friendly. I already commented about my editor being configured for completely different code formatting, you should have assumed that I just didn't notice that it used `this` notation instead of assuming that i'm a blind idiot.)_

Flamanis: Statement of fact is not being mean.

RheaAyase: I'm merely passing on what I learned during my time managing and leading leading development teams at work for the benefit of the "manager" in this situation. Take it or leave it.

> Statement of fact is not being mean.

If left on its own in this situation, it is. Statement of a fact should be followed by your "story" that talks about the "why" and if it's this obvious that the person already knows your facts, either say that you're not trying to discourage them or use other technique to "soften it," or just don't state your fact in this manner in the first place - they already know it and you're wasting keyboard strokes.

FiniteReality:
This code is being rewritten anyway, so there's no point in it being changed.

RheaAyase:
Sure there is. Your rewrite takes months or years, this is a quick fix for years old issue that you've not fixed in 1.0 even tho it was present since 0.9x - how am I supposed to trust that 2.0 or 3.0 will fix it?

Flamanis:
You don't. If you want your own features, write your own version and use it.

RheaAyase:
> You don't. If you want your own features, write your own version and use it.

Every major bot using this library is using modified version of it. Seems to me, that there is something wrong with issue triage here.

Also the question isn't whether you implement it, the question is why not click the merge button when it's already there and doesn't break or change anything. It's an empty event, a property, no major code change or feature.

FiniteReality:
> no major code change

You're adding new API surface area which we have to maintain and provide support for. You're also adding behaviour which could have been implemented much easier. (according to Volt this should've been a one-line change)

"Every major bot using this library" is only two bots, AFAIK. Everybody else (a lot of people) use the NuGet version of Discord.Net and hardly ever complain - the only major complaint we've got is that 1.x means they have to rewrite their code to follow sane C# programming standards.

PR #808, opened 2017-09-08
Changed Debug and Verbose output colors

Erlite:
DarkGray is basically black on Linux (tested on Debian 8)


FiniteReality:
The actual colours displayed can be changed by the client, anyway.

Erlite:
Yeah but that will avoid some people asking why they can't see debug/verbose logging.

Issue #810, opened 2017-09-09
Channel Categories are not supported at the API-level


LiruMouse:
In Discord.Net 1.0.1 and on the latest dev(479361bbea4f7baebbb55ceb211ce9f02368b91e), Trying to interact with the bot will fail for any server with categories. Interaction in PMs and in servers without categories works fine.
(Oddly this isn't an issue for my bots still running on 0.9)

Here's a piece of my log from typing and sending a message in one such server:
```[23:47:14.2318847] [Gateway] Received Dispatch (TYPING_START)
[23:47:14.2318847] [Gateway] Unsynced Guild (TYPING_START Guild=63296437932273664).
[23:47:17.8300030] [Gateway] Received Dispatch (MESSAGE_CREATE)
[23:47:17.8305033] [Gateway] Unsynced Guild (MESSAGE_CREATE Guild=63296437932273664).
```

So I decided to do some debugging, and put a breakpoint in `client.MessageReceived`, I got a hit for DM, and for a server without categories, but not for messaging in any channel in a category server.

misapprehending:
You'll also get
`[Error] Gateway -> Error handling Dispatch (GUILD_AVAILABLE)`

FiniteReality:
This should be partially fixed in ec03883; a full fix is coming soon.

foxbot:
This has been resolved as of ec03883e26fca5f67cc00b8cd2ba8cc5b8158b73.

A patch release for this issue is live on NuGet under 1.0.2 -- the latest 2.0 alphas are also no longer affected.

PR #811, opened 2017-09-09
Create unspecified channel object for unknown channel types

AntiTcb:
Partial resolution of #810. Yay Discord for introducing breaking changes out of left field.

### DIFF
--- a/src/Discord.Net.WebSocket/Entities/Channels/SocketGuildChannel.cs
+++ b/src/Discord.Net.WebSocket/Entities/Channels/SocketGuildChannel.cs
@@ -10,7 +10,7 @@ using Model = Discord.API.Channel;
 namespace Discord.WebSocket
 {
     [DebuggerDisplay(@"{DebuggerDisplay,nq}")]
-    public abstract class SocketGuildChannel : SocketChannel, IGuildChannel
+    public class SocketGuildChannel : SocketChannel, IGuildChannel
     {
         private ImmutableArray<Overwrite> _overwrites;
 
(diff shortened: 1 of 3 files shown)


src/Discord.Net.WebSocket/Entities/Channels/SocketGuildChannel.cs
```diff
@@ -35,7 +35,8 @@ internal static SocketGuildChannel Create(SocketGuild guild, ClientState state,
                 case ChannelType.Voice:
                     return SocketVoiceChannel.Create(guild, state, model);
                 default:
-                    throw new InvalidOperationException("Unknown guild channel type");
+                    // TODO: Proper implementation for channel categories
+                    return new SocketGuildChannel(guild.Discord, model.Id, guild);
```
Chris-Johnston, line 39: Should we call SocketGuildChannel.Update before returning? That's what SocketTextChannel and SocketVoiceChannel both do.

```csharp
default:
// TODO: Proper implementation for channel categories
SocketGuildChannel ret = new SocketGuildChannel(guild.Discord, model.Id, guild);
ret.Update(state, model);
return ret;
```

In my example case, IGuildChannels returned by `IGuild#GetChannelsAsync()` at least have their names. [Screenshot before/after.](http://i.imgur.com/kKnSltG.png)

Same could probably be applied to `RestGuildChannel`.

AntiTcb: Doing so would make the assumption that whatever unknown channel type existed follows the Channel model pattern, to some degree, which, in reality probably could be safely assumed, but I'd err on the side of skepticism here.

Chris-Johnston: Aren't we already making that assumption by returning a `SocketGuildChannel` in the first place for this patch?

AntiTcb: `SocketGuildChannel.Create()` is only ever used internally in a context where you'd only ever expect guild channels.

Having said that, I should probably also change [RestChannel](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Rest/Entities/Channels/RestChannel.cs) to cover those bases as well.

PR #819, opened 2017-09-23
System Channel Id ("Welcome new members channel") support, fix #815

### LINKED ISSUE #815: No method for retrieving the "Welcome New Members" channel
Recently Discord has added a feature under Server Settings>Overview where one can set a channel as the channel that welcome messages will be sent to when a user joins. However, this change doesn't seem to be reflected in the .NET API yet as I can't find a method for obtaining the text channel set for welcoming new users. Will this be implemented, or has it already been implemented?

Chris-Johnston:
Implementation of the "new member messages channel" guild option. Tried to closely follow the similar implementation of AFKChannel.

I was able to test the functionality of `SocketChannel#SystemChannel` property and `IGuild.ModifyAsync`, and both appear to be working as expected.

Per @khionu 's comment on #815, and some confirmation in the debugger, `system_channel_id` appears to be the correct name for this value in the API. I kept this name in the implementation to stay consistent with the API.

Feedback and review is appreciated. My discord handle is ChrisJ#8703.

### DIFF
--- a/src/Discord.Net.Core/Entities/Guilds/GuildProperties.cs
+++ b/src/Discord.Net.Core/Entities/Guilds/GuildProperties.cs
@@ -60,6 +60,14 @@
         /// </summary>
         public Optional<ulong?> AfkChannelId { get; set; }
         /// <summary>
+        /// The ITextChannel where System messages should be sent.
+        /// </summary>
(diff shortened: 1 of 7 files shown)


FiniteReality (approved):
Looks good to me, I don't see any obvious issues.

PR #825, opened 2017-09-26
Improve the Command Service documentation

Still34:
### The PR will,
-	Fix minor grammatical errors.
-	Capitalize terms such as Commands, Modules and such, as the context is specific to the lib.
-	Wrap methods and properties in code blocks.

The current Command Service documentation suffers from several minor issues here and there; the PR aims to improve upon the Command Service section of the docs and add clarifications to ambiguous parts of the docs. 

Below are several issues that remain to be fixed or need to be clarified.

---

1. >This article is out of date and has not been rewritten yet.
Information is not guaranteed to be accurate.

The docs doesn't necessarily seem "out of date" as the warning claims. The basics seem pretty relevant to the latest version of the lib. This warning turns off newcomers, as they might think that the official guide is unusable because of it being "outdated."

---

2. >“To manually load a module, invoke [CommandService.AddModuleAsync], by passing in the generic type of your module and optionally, a dependency map.”

The latter part of the sentence seems unclear. Where should the user pass the dependency map to? 

---

3. >“First, you need to create an @System.IServiceProvider You may create your own IServiceProvider if you wish.” 

Any mention of @System.IServiceProvider is currently broken on the docs.

---

4. >“Submodules are Modules that reside within another one. Typically, submodules are used to create nested groups (although not required to create nested groups).”

Clarification on the part after "although?"

---

5. >“Finally, pass the map into the `LoadAssembly` method. Your modules will automatically be loaded with this dependency map.”

Where is this `LoadAssembly` method?

---

6. >Preconditions can be applied to Modules, Groups, or Commands.

The docs should mention `ParameterPreconditionAttribute`'s existence.

---

7. ~~>To use Commands, you must create a [Commands Service]...~~

~~The `[Commands Service]` xref tag is broken.~~

Resolved in f897798.

### DIFF
--- a/Discord.Net.sln
+++ b/Discord.Net.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26228.4
+VisualStudioVersion = 15.0.26730.12
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Discord.Net.Core", "src\Discord.Net.Core\Discord.Net.Core.csproj", "{91E9E7BD-75C9-4E98-84AA-2C271922E5C2}"
 EndProject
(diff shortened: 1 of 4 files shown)


Still34:
In the sample code for Command Handler `var context = new CommandContext(client, message);`, shouldn't we encourage the use of `SocketCommandContext` instead?

PR #832, opened 2017-09-30
Added StreamType.Youtube

Yucked:
Needed this since Youtube is another big streaming platform.

### DIFF
--- a/src/Discord.Net.Core/Entities/Users/StreamType.cs
+++ b/src/Discord.Net.Core/Entities/Users/StreamType.cs
@@ -3,6 +3,7 @@
     public enum StreamType
     {
         NotStreaming = 0,
-        Twitch = 1
+        Twitch = 1,
+        Youtube = 2
     }
 }


FiniteReality:
Are you sure that Youtube is actually a supported stream type? The docs say that it's not, and other people seem to agree with that.

SubZero0:
Discord Docs says it's twitch only and there's no PR about Youtube.
https://github.com/discordapp/discord-api-docs/blob/master/docs/topics/Gateway.md#game-types

"The streaming type currently only supports Twitch. Only https://twitch.tv/ urls will work."

MinnDevelopment:
Considering the type is called "Streaming" in the official documentation you can assume that all other platforms will use the same game type just with different urls.

Yucked:
Yea I know they only support Twitch but I didn't wanna use Twitch enum for when I'm streaming live on Youtube. On a 2nd note, I just realized when I'm streaming it thinks I'm streaming on Twitch but probably directs to my Youtube channel :thinking:

Emzi0767:
Documentation [clearly states](https://discordapp.com/developers/docs/topics/gateway#game-object) that only Twitch is supported, furthermore it does not define game type 2 at all. So where did this come from? Which part of the documentation is this based on?

PR #834, opened 2017-09-30
Update voice OP codes

Still34:


MinnDevelopment:
Where are you using these?

Still34:
Several people have complained about `Unknown Opcode: 8` in the past, figured the lib should have the updated list.

MinnDevelopment:
Some of these codes only apply to v3 of the voice gateway so if you change them you should also make sure you use the correct version of that.

Emzi0767:
This commit would actually break voice, unless you switcg VGW to v3, and implement the remaining logic for that.

Still34:
I actually did not realize that we weren't using the latest Gateway for voice. My bad, although I might look into implementing them. For now, I will close the PR and examine this myself first.

PR #835, opened 2017-10-01
Implemented emoji endpoints

Emzi0767:
This PR implements the emoji endpoints, as [documented here](https://discordapp.com/developers/docs/resources/emoji). This enables bots to manipulate guild emoji.

in action

There is one caveat, however. Bots are only able to get, modify, and delete emoji they've created. They can, however, list all emoji, but it requires manage emoji permission.

Everything is tested, and it works.

### DIFF
--- a/src/Discord.Net.Core/Entities/Guilds/IGuild.cs
+++ b/src/Discord.Net.Core/Entities/Guilds/IGuild.cs
@@ -117,5 +117,16 @@ namespace Discord
         Task DownloadUsersAsync();
         /// <summary> Removes all users from this guild if they have not logged on in a provided number of days or, if simulate is true, returns the number of users that would be removed. </summary>
         Task<int> PruneUsersAsync(int days = 30, bool simulate = false, RequestOptions options = null);
+
+        /// <summary> Gets a collection of all emotes in this guild. </summary>
+        Task<IReadOnlyCollection<GuildEmote>> ListEmotesAsync(RequestOptions options = null);
+        /// <summary> Gets a specific emote from this guild. </summary>
+        Task<GuildEmote> GetEmoteAsync(ulong id, RequestOptions options = null);
+        /// <summary> Creates a new emote in this guild. </summary>
+        Task<GuildEmote> CreateEmoteAsync(string name, Image image, Optional<IEnumerable<IRole>> roles = default(Optional<IEnumerable<IRole>>), RequestOptions options = null);
+        /// <summary> Modifies an existing emote in this guild. </summary>
+        Task<GuildEmote> ModifyEmoteAsync(GuildEmote emote, Action<EmoteProperties> func, RequestOptions options = null);
+        /// <summary> Deletes an existing emote from this guild. </summary>
+        Task DeleteEmoteAsync(GuildEmote emote, RequestOptions options = null);
     }
 }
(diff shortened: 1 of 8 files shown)


src/Discord.Net.Core/Entities/Guilds/IGuild.cs
```diff
@@ -117,5 +117,16 @@ public interface IGuild : IDeletable, ISnowflakeEntity
...
         Task<int> PruneUsersAsync(int days = 30, bool simulate = false, RequestOptions options = null);
+
+        /// <summary> Gets a collection of all emotes in this guild. </summary>
+        Task<IReadOnlyCollection<GuildEmote>> ListEmotesAsync(RequestOptions options = null);
+        /// <summary> Gets a specific emote from this guild. </summary>
+        Task<GuildEmote> GetEmoteAsync(ulong id, RequestOptions options = null);
+        /// <summary> Creates a new emote in this guild. </summary>
+        Task<GuildEmote> CreateEmoteAsync(string name, Image image, Optional<IEnumerable<IRole>> roles = default(Optional<IEnumerable<IRole>>), RequestOptions options = null);
+        /// <summary> Modifies an existing emote in this guild. </summary>
+        Task<GuildEmote> ModifyEmoteAsync(GuildEmote emote, Action<EmoteProperties> func, RequestOptions options = null);
```
foxbot, line 126: Any reason these methods were left on IGuild instead of GuildEmote?

src/Discord.Net.Core/Entities/Guilds/IGuild.cs
```diff
@@ -117,5 +117,16 @@ public interface IGuild : IDeletable, ISnowflakeEntity
         Task DownloadUsersAsync();
         /// <summary> Removes all users from this guild if they have not logged on in a provided number of days or, if simulate is true, returns the number of users that would be removed. </summary>
         Task<int> PruneUsersAsync(int days = 30, bool simulate = false, RequestOptions options = null);
+
+        /// <summary> Gets a collection of all emotes in this guild. </summary>
+        Task<IReadOnlyCollection<GuildEmote>> ListEmotesAsync(RequestOptions options = null);
```
ObsidianMinor: Doesn't this already exist in the form of `IGuild.Emotes`?

src/Discord.Net.Core/Entities/Guilds/IGuild.cs
Emzi0767: GuildEmote does not hold the Guild object required for this to work. This would require some deeper changes, and I didn't feel confident enough to make them without some guidance.

src/Discord.Net.Core/Entities/Guilds/IGuild.cs
Emzi0767: It also exists in the API. See: linked Discord API documentation.

ObsidianMinor: Is there an advantage or reason to run `ListEmotesAsync` (which should probably be `GetEmotesAsync`) vs reading the cached collection?

Emzi0767: AFAIK, not really. But it exists, so I implemented it.

foxbot: Better to just leave this off - we can always readd it if there's any API changes that make it necessary

Emzi0767: A'ight, is commenting out acceptable? Will make it easier to re-add later.

foxbot: it's only 10 lines, just remove it

PR #837, opened 2017-10-02
Adding missing WithVideoUrl method for EmbedBuilder

Ilddor:
So while working with Embeds in my bot I came across VideoEmbed being unable to build with EmbedBuilder, it seems that only get/set and "WithVideoUrl" parts are missing while everything else is already implemented. So here it is:)

### DIFF
--- a/src/Discord.Net.Rest/Entities/Messages/EmbedBuilder.cs
+++ b/src/Discord.Net.Rest/Entities/Messages/EmbedBuilder.cs
@@ -66,6 +66,15 @@ namespace Discord
                 _embed.Image = new EmbedImage(value, null, null, null);
             }
         }
+        public string VideoUrl
+        {
+            get => _embed.Video?.Url;
+            set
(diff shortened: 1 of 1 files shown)


Joe4evr:
I don't think this is supported with user-made embeds, which would be the entire reason that it's not there.

Ilddor:
But isn't it used for gifs etc? If so then other APIs seems to use it this way

Joe4evr:
GIFs are just normal images, AFAIK.

Ilddor:
Well, linking GIF with "WithImageUrl" does not work for me, message appears empty

Still34:
Clarify doesn't work? It should.

Ilddor:
Clarify?

FiniteReality:
Discord only supports rich embeds, which do not support the `video` field: https://discordapp.com/developers/docs/resources/channel#create-message

PR #839, opened 2017-10-04
Improve upon Property Summaries in CommandServiceConfig

Vollrat:
Having the `CaseSensitiveCommands` property summary asking a question whenever Intellisense is invoked seems a bit nonessential instead of properly explaining what exactly it does. It would be better if instead, it stated it's use to be more comprehensible to the reader.

### Changes
- Edits the summaries of CaseSensitiveCommands and ThrowOnError to follow a more methodical convention for boolean property summaries 
(Determines whether X ... rather than Should X be ... ?).
- Changes the summary for `DefaultRunMode` to accurately reflect the default standard and be more consistent,

This is just a small change to improve upon the current documentation, so it shouldn't conflict with anything.

### DIFF
--- a/src/Discord.Net.Commands/CommandServiceConfig.cs
+++ b/src/Discord.Net.Commands/CommandServiceConfig.cs
@@ -2,7 +2,7 @@
 {
     public class CommandServiceConfig
     {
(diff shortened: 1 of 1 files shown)


Emzi0767:
Gets or sets is the C# convention for docstrings though.

PR #848, opened 2017-10-16
Add exception parameter to AfterExecute method

janvostry:
Adds ability to observe eventual exception in AfterExecute method.

This is handy since commands with _RunMode.Async_ doesn't report error status and the eventual exception might be lost.


Joe4evr:
You're kinda expected to use the `CommandService.CommandExecuted` event to observe exceptions from Async commands.

janvostry:
@Joe4evr Yeah, that looks better.

Issue #850, opened 2017-10-19
TCP Buffer overflow with large bot accounts.


RheaAyase:
The 2.0.x version of the library is unable to handle medium-to-large bot accounts as it is too slow to process messages fast enough and causes `TCP ZeroWindow` errors almost every 100 milliseconds.

[.NET Core 2.0.0](https://copr.fedorainfracloud.org/coprs/g/dotnet-sig/dotnet/) on `Fedora 26 4.13.5-200.fc26.x86_64` and `Fedora 26 4.11.11-300.fc26.x86_64`

See the example test bot at https://github.com/RheaAyase/Discord.NET-issue-850 where you can also find the [tcp dump](https://cloud.rhea-ayase.eu/s/o3qSCkmIs46VJiT), and a screenshot demonstrating the issue. (Change the token to whatever you can use as large test bot account to reproduce the issue.)

My test large bot has about half a million users (only 450 guilds) but I've been able to reproduce it on around 20 000 users as well. The more you have, the more consistent. it becomes. In the case of half a million users, it's every single time it connects, straight away. On 40 000 users, it's within 30 seconds after connecting... etc...

Exceptions observed:
```
Exception message: WebSocket connection was closed
Exception stack:
InnerException message: The remote party closed the WebSocket connection without completing the close handshake.
InnerException stack:
```

RheaAyase:
_(Feel free to correct the issue title if this is not appropriate.)_

FiniteReality:
Might want to remove your token from that repo. Furthermore, there's no TCP dump there to see.

FiniteReality:
After a quick chat with Voltana, it seems that this will likely be fixed in 3.0. The slowdown is probably due to JSON parsing.

An intermediate solution would probably be to [increase the TCP buffer size](http://www.linux-admins.net/2010/09/linux-tcp-tuning.html) until 3.0 hits.

RheaAyase:
> Might want to remove your token from that repo.

The token is a dummy (invalidated before commit.)

> Furthermore, there's no TCP dump there to see.

Yeah I didn't make it before some work stuff, sorry about that... (Don't want to post original data containing some private IP's and stuff...) I'll slap a screenshot of it... [right here!](https://cloud.rhea-ayase.eu/s/o3qSCkmIs46VJiT)

> An intermediate solution would probably be to increase the TCP buffer size until 3.0 hits.

[...]

PR #853, opened 2017-10-20
Remove payload compression, use stream compression

FiniteReality:
This should help improve #850. I have tested this code and it appears to work correctly with 28k users across 3 guilds.

### DIFF
--- a/src/Discord.Net.WebSocket/DiscordSocketApiClient.cs
+++ b/src/Discord.Net.WebSocket/DiscordSocketApiClient.cs
@@ -29,7 +29,11 @@ namespace Discord.API
         private CancellationTokenSource _connectCancelToken;
         private string _gatewayUrl;
         private bool _isExplicitUrl;
-        
+
+        //Store our decompression streams for zlib shared state
+        private MemoryStream _compressed;
+        private DeflateStream _decompressor;
+
         internal IWebSocketClient WebSocketClient { get; }
 
         public ConnectionState ConnectionState { get; private set; }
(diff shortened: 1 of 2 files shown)


src/Discord.Net.WebSocket/DiscordSocketApiClient.cs
```diff
@@ -202,8 +230,7 @@ public Task SendGatewayAsync(GatewayOpCode opCode, object payload, RequestOption
             {
                 Token = AuthToken,
                 Properties = props,
-                LargeThreshold = largeThreshold,
-                UseCompression = useCompression,
+                LargeThreshold = largeThreshold
```
khionu, line 233: Trailing comma?

FiniteReality: I removed it since it didn't make sense. It doesn't harm functionality.

khionu: The reason it's typically included is for Diffs, so it shows as a single-line change vs two.

FiniteReality: I'm fairly sure we don't do it in other places... There's no point changing it back now because it'll show both commits.

RheaAyase:
Can confirm that this helps #850 _a little bit..._ The issue happens with a few times more user counts now.

FiniteReality:
That sounds great; glad to know that it's helping rather than hindering 😛

foxbot (approved):

PR #874, opened 2017-11-07
Fix #854 Added ViewChannel enum and property to channel permissions

### LINKED ISSUE #854: Read Permission -> View Permission
Given the same bitvalue is now used to determine whether you can see a voice channel, I propose creating a new value, named appropriately, giving it the bitvalue, then pointing the old value to the new, marking it as deprecated.

Chris-Johnston:
#854 proposed adding an enum and property to channel permissions that obsoletes `ReadMessages` in favor of `ViewChannel`.

I could see this being an issue with existing code, making `ReadMessages` obsolete will give unnecessary warnings. In addition, the name now differs from how it is defined in the API.

In addition, there is a discrepancy in the naming of ManageChannel(s) for ChannelPermissions.

https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Core/Entities/Permissions/ChannelPermissions.cs#L39

https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Core/Entities/Permissions/ChannelPermission.cs#L10

https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Core/Entities/Permissions/GuildPermission.cs#L13

Would this also be something worth changing, obsoleting one in favor of another? Or could/should this be a breaking change?

### DIFF
--- a/src/Discord.Net.Core/Entities/Permissions/ChannelPermission.cs
+++ b/src/Discord.Net.Core/Entities/Permissions/ChannelPermission.cs
@@ -11,7 +11,9 @@ namespace Discord
 
         // Text
         AddReactions        = 0x00_00_00_40,
-        ReadMessages        = 0x00_00_04_00,
+        [Obsolete("Use ViewChannel instead.")]
+        ReadMessages        = ViewChannel,
+        ViewChannel         = 0x00_00_04_00,
         SendMessages        = 0x00_00_08_00,
         SendTTSMessages     = 0x00_00_10_00,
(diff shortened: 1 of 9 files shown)


Chris-Johnston:
Build output shows how many warnings pop up as a result: https://ci.appveyor.com/project/RogueException/discord-net/build/build-847-fhpuuugu#L88

Maybe should reconsider.

Chris-Johnston:
Oh, yeah that should be pretty easy. Can add that later today.

foxbot (approved):

PR #877, opened 2017-11-10
Refactor Games, support reading Rich Presences

foxbot:
## Abstract

This pull-request refactors the archaic `Discord.Game` type into a more robust `IActivity` interface, with implementations for `Game`, `StreamingGame`, and `RichGame`.

This change supports reading rich presences, but cannot send them - Discord does not permit bots to send rich presences, they must be sent over the RPC protocol.

### Additions

An abstract `IActivity` interface has been added<sup>[1]</sup>, to allow for more generic implementations of playing statuses as Discord adds them.

`RichGame` has been added to support reading of Rich Presences<sup>[2]</sup>

### Changes

**Breaking:** The old `Game` structure has been refactored into a class that inherits from `IActivity`. Properties specific to streaming statuses have been moved to `StreamingGame`.<sup>[3]</sup>

**Breaking:** `IPresence.Game` has been renamed to `IPresence.Activity`, and now returns an `IActivity`.<sup>[4]</sup>

## Possible Caveats:

This change assumes that all rich presences will be sent with their `application_id` set<sup>[5]</sup>. This assumption is based on the official `discord-rpc` SDK, which requires<sup>[6]</sup> `application_id` to be set when initializing. Rich Presences without this field set are more than likely fake, or originate from an unofficial/inaccurate implementation of the SDK. 

[1]: https://github.com/RogueException/Discord.Net/blob/bb3c1c287cdf3aab84a5537ac767b7054b901170/src/Discord.Net.Core/Entities/Activities/IActivity.cs
[2]: https://github.com/RogueException/Discord.Net/blob/bb3c1c287cdf3aab84a5537ac767b7054b901170/src/Discord.Net.Core/Entities/Activities/RichGame.cs
[3]: https://github.com/RogueException/Discord.Net/blob/bb3c1c287cdf3aab84a5537ac767b7054b901170/src/Discord.Net.Core/Entities/Activities/Game.cs#L6
[4]: https://github.com/RogueException/Discord.Net/blob/bb3c1c287cdf3aab84a5537ac767b7054b901170/src/Discord.Net.Core/Entities/Users/IPresence.cs#L6
[5]: https://github.com/RogueException/Discord.Net/blob/bb3c1c287cdf3aab84a5537ac767b7054b901170/src/Discord.Net.WebSocket/Extensions/EntityExtensions.cs#L8
[6]: https://github.com/discordapp/discord-rpc/blob/master/src/discord-rpc.cpp#L225


src/Discord.Net.Core/Entities/Activities/GameAssets.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace Discord
+{
+    public class GameAssets
```
khionu: Given the only context that this pairing occurs is in a Rich Presence, I would suggest removing this class and having properties for each asset separately.

src/Discord.Net.Core/Entities/Activities/Game.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Diagnostics;
+
+namespace Discord
+{
+    [DebuggerDisplay(@"{DebuggerDisplay,nq}")]
+    public class Game : IActivity
+    {
+        public string Name { get; internal set; }
+
+        public Game() { }
```
Joe4evr: Do we really want this ctor public?

khionu (changes requested):
In addition to the below, there are numerous ctors that are public and should not be.

src/Discord.Net.Core/Entities/Activities/Game.cs
foxbot: Fixed.

src/Discord.Net.Core/Entities/Activities/GameParty.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Discord
+{
+    public class GameParty
+    {
+        internal GameParty() { }
+
+        public string Id { get; internal set; }
+        public int[] Size { get; internal set; }
```
khionu: Array of Int? This probably has a better way of being exposed. Assuming it's two ints, count and cap, a maybe a ValueTuple?

PR #888, opened 2017-11-20
Allow users to opt-in to proxies

### LINKED ISSUE #836: Adding proxy support
Hello,
Is there a particular reason as to why the proxies are disabled @ https://github.com/RogueException/Discord.Net/blob/1ce1c019b399b83fb7ed946ec7908d5d37b567fb/src/Discord.Net.Rest/Net/DefaultRestClient.cs#L33 ?
It would be nice for the library to work in restricted environments.
Thanks.

foxbot:
## Abstract

This pull request resolves #836.

To opt-in to proxies, the user must use a custom Rest Client provider:

```cs
var config = new DiscordRestConfig // or DiscordSocketConfig, if using a Socket client
{
    RestClientProvider = DefaultRestClientProvider.Create(useProxy: true);
}
```

## Possible Caveats

When opting in to proxies, client performance may degrade. Before this option was disabled (on older versions of CoreCLR -- this may have been before the 1.0 release, so the issue could be fixed now!), requests were prone to take >5x the time to complete, regardless of using a proxy.

### DIFF
--- a/src/Discord.Net.Rest/Net/DefaultRestClient.cs
+++ b/src/Discord.Net.Rest/Net/DefaultRestClient.cs
@@ -22,7 +22,7 @@ namespace Discord.Net.Rest
         private CancellationToken _cancelToken;
         private bool _isDisposed;
 
-        public DefaultRestClient(string baseUrl)
+        public DefaultRestClient(string baseUrl, bool useProxy = false)
         {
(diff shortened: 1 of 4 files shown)


Emzi0767:
What about WebSocket connections?

PR #906, opened 2017-12-13
Add API Analyzer Assembly

Joe4evr:
There's a number of things that people tend to fall into when using this lib. But we have the power of Roslyn, so I thought it was time to let it do some work for us and have the compiler point out common mistakes to our users.

Unit tests are included.

Next analyzer will be checking if you aren't leaking `this` from inside modules.


FiniteReality:
Awesome, these are looking good. Would it be possible to write analysers for suggesting that users write preconditions and parameter preconditions instead of using if/else chains in their code?

That is, something like this:
```cs
public async Task RollDice(int numberOfDice)
{
    if (numberOfDice < 0)
        await ReplyAsync("Must be at least more than one dice");
    else
        await ReplyAsync($"You rolled a {numberOfDice * 3}");
}
```

Which could be:
```cs
public async Task RollDice([AtLeast(1)]int numberOfDice)
{
    await ReplyAsync($"You rolled a {numberOfDice * 3}");
}
```

This is for future use, of course: right now we don't have very many built-in preconditions, which is something I hope to change when I rewrite the commands system.

Joe4evr:
That would be *really* complicated to achieve.

Not impossible, but it's a lot more likely to get false positives/false negatives on something like that.

FiniteReality:
Okay, that's fine, it doesn't surprise me that it would be really complicated to achieve.

PR #909, opened 2017-12-18
Enable Mono support with new WebSocket provider

Nihlus:
This PR enables Discord.Net to support running on Mono, by adding a new optional WebSocket provider based on the `websocket-sharp` library.

As an example, this allows the use of Pi Zero boards as hosts for Discord.Net bots, since .NET Core only supports armv7 processors - the Zero runs an armv6l. Mono does not have this issue - see https://github.com/dotnet/core/issues/667 and related for more information.

The provider has been thoroughly tested in a nontrivial Discord.Net bot.

Of note is the usage of the `websocket-sharp-customheaders` package instead of the official package. The official package does not support custom headers, and as such, this fork is used instead.

Depending on your version of Mono, an additional project-specific workaround may be needed. The `System.Net.Http.dll` file must be deleted from the output directory, and an assembly redirect must be added to the configuration.

```xml
<configuration>
    <runtime>
        <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
            <dependentAssembly>
                <assemblyIdentity name="System.Net.Http" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
                <bindingRedirect oldVersion="0.0.0.0-4.2.0.0" newVersion="4.0.0.0" />
            </dependentAssembly>
        </assemblyBinding>
    </runtime>
</configuration>
```
```xml
<Target Name="DeleteNetHttpMono" AfterTargets="AfterBuild">
  <Delete Files="$(OutputPath)\System.Net.Http.dll" />
</Target>
```

### DIFF
--- a/Discord.Net.sln
+++ b/Discord.Net.sln
@@ -1,4 +1,4 @@
-Microsoft Visual Studio Solution File, Format Version 12.00
+﻿Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
 VisualStudioVersion = 15.0.26730.12
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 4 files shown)


Nihlus:
An alternate websocket provider is already included in the library as an optional package (WS4Net). This PR does not explicitly add first-class support for Mono, but it does allow its use by adding this new provider. The primary use case is that it allows d.net to run on a much broader spectrum of hardware where .NET Core does not run, but Mono does.

Nihlus:
I can see the argument. It would be a lot more convenient to have this built along with Discord.Net itself, though.

Nihlus:
I've created a standalone project for this provider, and an associated Nuget package.

https://github.com/Nihlus/Discord.Addons.WebsocketSharp
https://www.nuget.org/packages/Discord.Addons.WebSocketSharp/1.0.1

Issue #602, opened 2017-04-09
Suggestion: Ignore all params after the required ones.


RealBlazeIt:
I constantly see in my chat users using commands with no params, who add some either way, that don't effect the command itself, but still prevent it from being run.

For example, if a use was to type `!invite link` the command would fail responding that they were incorrectly using the command, and provide the correct usage, being: `!invite`.

I know this can be fixed by adding the param `[Remainder] string ignore = null` to every command, but I feel it would be insanely redundant.

Another example would be something like: `!chill 30 stop spamming guys`, a user putting some text after the required params, but in consequence failing the command.

I feel it's unnecessary to fail the execution of a command simply because they added things after a command, and it would be nice if there could be a bool added in the `CommandServiceConfig` allowing you to ignore everything after the required params of a command.

FiniteReality:
I finally have a bit of free time to put some input here.

I totally agree that ignoring extra params would be useful, and I don't even think it would be that hard to implement. (Just break out of the loop [here](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Commands/CommandParser.cs#L111) rather than return an error)

foxbot:
Note that this type of behavior should be a config option - not everyone will want the command to parse successfully if it contains extra parameters.

foxbot:
This was resolved as of 5f46aef3a761b24dcd8ef5ecb4b8811d0f374059

PR #915, opened 2017-12-25
Ability to ignore unused parameters instead of failing the command.

ghost:
This will implement the suggestion in #602. This has been tested and works as expected.

### DIFF
--- a/src/Discord.Net.Commands/CommandParser.cs
+++ b/src/Discord.Net.Commands/CommandParser.cs
@@ -14,7 +14,7 @@ namespace Discord.Commands
             QuotedParameter
         }
         
-        public static async Task<ParseResult> ParseArgsAsync(CommandInfo command, ICommandContext context, CommandService commandService, IServiceProvider services, string input, int startPos)
+        public static async Task<ParseResult> ParseArgsAsync(CommandInfo command, ICommandContext context, bool ignoreExtraArgs, IServiceProvider services, string input, int startPos)
         {
             ParameterInfo curParam = null;
             StringBuilder argBuilder = new StringBuilder(input.Length);
@@ -110,10 +110,10 @@ namespace Discord.Commands
                 {
                     if (curParam == null)
                     {
-                        if (commandService._failOnTooManyArgs)
-                            return ParseResult.FromError(CommandError.BadArgCount, "The input text has too many parameters.");
-                        else
+                        if (ignoreExtraArgs)
                             break;
+                        else
+                            return ParseResult.FromError(CommandError.BadArgCount, "The input text has too many parameters.");
                     }
 
                     var typeReaderResult = await curParam.ParseAsync(context, argString, services).ConfigureAwait(false);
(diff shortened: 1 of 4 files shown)


MarkusG:
If this gets merged, I suggest renaming to `FailOnExtraArgs` to be less verbose.

Additionally, it might be wiser to make an attribute that can be applied to modules and commands in addition to a configuration across the entire service. This would allow for more precise control of when the command should fail, while still allowing users to ignore extra arguments for all of their commands if they please.

src/Discord.Net.Commands/CommandParser.cs
```diff
@@ -109,7 +109,12 @@ private enum ParserPart
                 if (argString != null)
                 {
                     if (curParam == null)
-                        return ParseResult.FromError(CommandError.BadArgCount, "The input text has too many parameters.");
+                    {
+                        if (commandService._failOnTooManyArgs)
```
FiniteReality: Personally, I'm more of a fan of passing a bool arg `failOnExtraArgs` instead of the whole CommandService.

src/Discord.Net.Commands/CommandServiceConfig.cs
```diff
@@ -15,5 +15,8 @@ public class CommandServiceConfig
 
         /// <summary> Determines whether RunMode.Sync commands should push exceptions up to the caller. </summary>
         public bool ThrowOnError { get; set; } = true;
+
+        /// <summary> Determines whether the command execution should fail if too many parameters are provided. </summary>
+        public bool FailOnTooManyArgs { get; set; } = true;
```
FiniteReality: I think a better name for this property would be `IgnoreExtraArguments`, and its behaviour should be inversed to match the opposite behaviour.

src/Discord.Net.Commands/Info/CommandInfo.cs
```diff
@@ -18,8 +18,9 @@ public class CommandInfo
         private static readonly System.Reflection.MethodInfo _convertParamsMethod = typeof(CommandInfo).GetTypeInfo().GetDeclaredMethod(nameof(ConvertParamsList));
         private static readonly ConcurrentDictionary<Type, Func<IEnumerable<object>, object>> _arrayConverters = new ConcurrentDictionary<Type, Func<IEnumerable<object>, object>>();
 
+        private readonly CommandService _commandService;
         private readonly Func<ICommandContext, object[], IServiceProvider, CommandInfo, Task> _action;
-
+
```
FiniteReality: Please remove extra whitespace

FiniteReality (changes requested):
This is pretty good: it fits in with the style of the lib fairly well and is pretty much exactly how I'd implement it. Good job!

FiniteReality:
@MarkusGordathian Right now, per-module "configuration" of the command service like that is pretty hard to do. It's part of a larger issue which I'll tackle when I eventually get to rewriting it all.

ghost:
@FiniteReality I have added all the requested changes, except I named it `IgnoreExtraArgs` instead of `IgnoreExtraArguments` to keep consistency with `BadArgCount`.

FiniteReality (approved):

foxbot (approved):

PR #928, opened 2018-01-08
Apply consistency to attributes in the commands extension

### LINKED ISSUE #527: Can we have a consistent policy on `[AttributeUsage]` usage?
I would like for all the attributes that the lib provides to have consistent behavior (preferably that most attributes get `Inherited = true`). If that's not desirable, then at the very least specify the `AllowMultiple` and `Inherited` properties explicitly everywhere, so that it can be picked up by Metadata as Source in VS, and make some behavior [a lot less surprising](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

foxbot:
This resolves #527.

Not sure if I missed any, putting this up for review.

### DIFF
--- a/src/Discord.Net.Commands/Attributes/AliasAttribute.cs
+++ b/src/Discord.Net.Commands/Attributes/AliasAttribute.cs
@@ -3,7 +3,7 @@ using System;
 namespace Discord.Commands
 {
     /// <summary> Provides aliases for a command. </summary>
-    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method)]
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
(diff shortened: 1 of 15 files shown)


AntiTcb:
[RequireContextAttribute](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Commands/Attributes/Preconditions/RequireContextAttribute.cs#L18), [RequireNsfwAttribute](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Commands/Attributes/Preconditions/RequireNsfwAttribute.cs#L9), and [RequireOwnerAttribute](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Commands/Attributes/Preconditions/RequireOwnerAttribute.cs#L12) should allow multiples; has use for precondition groups.

FiniteReality (approved):

Issue #912, opened 2017-12-23
No equality operators for IUser (and possibly others)


TheBrambleShark:
This makes it impossible to use LINQ and other built-in methods for ensuring that objects are equal to each other. Will submit a pull request with this.

    IReadOnlyCollection<IUser> usersReacted = await msg.GetReactionUsersAsync(emote.Name); // Returns list of 25 users (subset of everyone in guild)
    IReadOnlyCollection<IUser> usersInGuild = sguser.Guild.Users; // Returns list of 50 users
    IEnumerable<IUser> usersToRemove = usersInGuild.Except(usersReacted); // Should return everyone in the guild except those who reacted

`usersToRemove` ends up being the same as `usersInGuild`. I'll try and find a less-painful solution than doing things manually.

khionu:
1) You can use the IDs, instead. That's how I've done my Linq in bots, and I imagine how most people do.
2) This is almost certainly how it would be done internally. If you wanted to PR it, you'd be free to (the acceptance of it to the lib would be up to the maintainers, though)

TheBrambleShark:
I can wrangle the return type for `GetReactionUsersAsync` to be a `RestUser`, but the only common type between `RestUser` and `SocketGuildUser` (The return type of `sguser.Guild.Users`) is `IUser` where equality is broken. Thus, the only kind of thing I can think of doing is writing my own method that steps through both collections and only takes the unique values based on ID, but it would be far easier to just rely on the built-in methods.

Joe4evr:
> Thus, the only kind of thing I can think of doing is writing my own method that steps through both collections and only takes the unique values based on ID

`.Except()` (and just about every other LINQ method) has an overload that takes in an `IEqualityComparer<T>` at which point you can just do [something like this](https://github.com/Joe4evr/Discord.Addons/blob/master/src/Discord.Addons.Core/Comparers.cs). No need to write your own methods.

khionu:
Like I noted, you can always use the IDs
```cs
if (restUserObject.Id == socketUserObject.Id)
    return;
```

FiniteReality:
Keep in mind that we already tried something like this in #181 and it ended up being postponed almost indefinitely. However, this is a good place to discuss it properly, so cc @RogueException @foxbot

TheBrambleShark:
There seem to be two solutions for this. Here's the code I'm using as suggested by @khionu:

[...]

PR #929, opened 2018-01-08
Comparers

Joe4evr:
As suggested in #912.

Also, if there are other suggestions for the containing class name, make them before this gets merged.

### DIFF
--- a/src/Discord.Net.Core/Utils/Comparers.cs
+++ b/src/Discord.Net.Core/Utils/Comparers.cs
@@ -1,0 +1,43 @@
+using System;
+using System.Collections.Generic;
+
+namespace Discord.Addons.Core
+{
+    public static class DiscordComparers
+    {
+        // TODO: simplify with '??=' slated for C# 8.0
+        public static IEqualityComparer<IUser> UserComparer => _userComparer ?? (_userComparer = new EntityEqualityComparer<IUser, ulong>());
+        public static IEqualityComparer<IGuild> GuildComparer => _guildComparer ?? (_guildComparer = new EntityEqualityComparer<IGuild, ulong>());
+        public static IEqualityComparer<IChannel> ChannelComparer => _channelComparer ?? (_channelComparer = new EntityEqualityComparer<IChannel, ulong>());
+        public static IEqualityComparer<IRole> RoleComparer => _roleComparer ?? (_roleComparer = new EntityEqualityComparer<IRole, ulong>());
+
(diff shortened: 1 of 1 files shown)


src/Discord.Net.Core/Utils/Comparers.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Discord
+{
+    public static class DiscordComparers
+    {
+        // TODO: simplify with '??=' slated for C# 8.0
+        public static IEqualityComparer<IUser> UserComparer => _userComparer ?? (_userComparer = new EntityEqualityComparer<IUser, ulong>());
+        public static IEqualityComparer<IGuild> GuildComparer => _guildComparer ?? (_guildComparer = new EntityEqualityComparer<IGuild, ulong>());
+        public static IEqualityComparer<IChannel> ChannelComparer => _channelComparer ?? (_channelComparer = new EntityEqualityComparer<IChannel, ulong>());
+        public static IEqualityComparer<IRole> RoleComparer => _roleComparer ?? (_roleComparer = new EntityEqualityComparer<IRole, ulong>());
+
```
AntiTcb: IMessage comparer as well?

Joe4evr: ~~Good call.~~

~~EDIT: Wait, wasn't there something that it's not guaranteed message IDs will remain globally unique?~~

Decided to make a specialized implementation just to be safe.

src/Discord.Net.Core/Utils/Comparers.cs
```diff
@@ -0,0 +1,67 @@
...
+                bool xNull = x == null;
+                bool yNull = y == null;
+
+                if (xNull && yNull)
+                    return true;
+
+                if (xNull ^ yNull)
+                    return false;
+
+                return x.Channel.Id.Equals(y.Channel.Id) && x.Id.Equals(y.Id);
```
foxbot: I'm not sure we really need to check channel IDs as well as message IDs, snowflakes in messages should be unique globally.

FiniteReality: Yeah, we definitely don't. All snowflakes generated by Discord are guaranteed to be unique, unless they are explicitly duplicated (e.g. default general channel, everyone role)

Joe4evr: I vaguely recall a Discord dev saying that message IDs being globally unique can't be guaranteed indefinitely.

PR #934, opened 2018-01-15
Add callback method for when a module class has been added

Joe4evr:
This is potentially breaking depending on how users have their code laid out. Previously, the following would just work:
```cs
_serviceCollection.AddSingleton(new MyService());
await _commands.AddModuleAsync<MyModule>();
// where 'MyModule' has 'MyService' as a constructor argument
```
With this PR right now, this code would throw an exception because the `IServiceProvider` isn't built yet, so the dependency can't get resolved. I could make the parameter required, but I wanted to get a review of the rest first.

The expected use-case is for something like adding a custom TypeReader for an associated module. Especially for `MpGame` where I want to provide the type reader as an implementation detail.

### DIFF
--- a/src/Discord.Net.Commands/Builders/ModuleBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/ModuleBuilder.cs
@@ -19,6 +19,7 @@ namespace Discord.Commands.Builders
         public string Name { get; set; }
         public string Summary { get; set; }
         public string Remarks { get; set; }
+        public string Group { get; set; }
 
         public IReadOnlyList<CommandBuilder> Commands => _commands;
         public IReadOnlyList<ModuleBuilder> Modules => _submodules;
--- a/src/Discord.Net.Commands/Builders/ModuleBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/ModuleBuilder.cs
@@ -122,8 +122,6 @@ namespace Discord.Commands.Builders
 
             if (TypeInfo != null)
             {
-                //keep this for safety?
-                //services = services ?? EmptyServiceProvider.Instance;
                 try
                 {
                     var moduleInstance = ReflectionUtils.CreateObject<IModuleBase>(TypeInfo, service, service._serviceProvider);
--- a/src/Discord.Net.Commands/Builders/ModuleBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/ModuleBuilder.cs
@@ -114,7 +114,7 @@ namespace Discord.Commands.Builders
             return this;
         }
 
-        private ModuleInfo BuildImpl(CommandService service, ModuleInfo parent = null)
+        private ModuleInfo BuildImpl(CommandService service, IServiceProvider services, ModuleInfo parent = null)
         {
             //Default name to first alias
             if (Name == null)
@@ -124,7 +124,7 @@ namespace Discord.Commands.Builders
             {
                 try
                 {
-                    var moduleInstance = ReflectionUtils.CreateObject<IModuleBase>(TypeInfo, service, service._serviceProvider);
+                    var moduleInstance = ReflectionUtils.CreateObject<IModuleBase>(TypeInfo, service, services);
                     moduleInstance.OnModuleBuilding(service);
                 }
                 catch (Exception)
@@ -134,11 +134,11 @@ namespace Discord.Commands.Builders
                 }
             }
 
-            return new ModuleInfo(this, service, parent);
+            return new ModuleInfo(this, service, services, parent);
         }
 
-        public ModuleInfo Build(CommandService service) => BuildImpl(service);
+        public ModuleInfo Build(CommandService service, IServiceProvider services) => BuildImpl(service, services);
 
-        internal ModuleInfo Build(CommandService service, ModuleInfo parent) => BuildImpl(service, parent);
+        internal ModuleInfo Build(CommandService service, IServiceProvider services, ModuleInfo parent) => BuildImpl(service, services, parent);
     }
 }
--- a/src/Discord.Net.Commands/Builders/ModuleBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/ModuleBuilder.cs
@@ -26,7 +26,7 @@ namespace Discord.Commands.Builders
         public IReadOnlyList<Attribute> Attributes => _attributes;
         public IReadOnlyList<string> Aliases => _aliases;
 
-        internal Optional<TypeInfo> TypeInfo { get; set; }
+        internal TypeInfo TypeInfo { get; set; }
 
         //Automatic
         internal ModuleBuilder(CommandService service, ModuleBuilder parent)
@@ -119,6 +119,22 @@ namespace Discord.Commands.Builders
             //Default name to first alias
             if (Name == null)
                 Name = _aliases[0];
+
+            if (TypeInfo != null)
+            {
+                //keep this for safety?
+                //services = services ?? EmptyServiceProvider.Instance;
+                try
+                {
+                    var moduleInstance = ReflectionUtils.CreateObject<IModuleBase>(TypeInfo, service, service._serviceProvider);
+                    moduleInstance.OnModuleBuilding(service);
+                }
+                catch (Exception)
+                {
+                    //unsure of what to do here
+                    throw;
+                }
+            }
 
             return new ModuleInfo(this, service, parent);
         }
--- a/src/Discord.Net.Commands/Builders/ModuleBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/ModuleBuilder.cs
@@ -1,5 +1,6 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
+using System.Reflection;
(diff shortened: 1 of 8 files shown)


FiniteReality:
I'm not too sure I like the way this has been implemented, it seems a bit weird... Is there any way to detect when a method has been overriden in a class? If so, we could probably add an 'OnModuleBuilding' which has the CommandService as a param, and the ModuleBuilder that is being built.

EDIT: Wait, we don't need to check for a method being overriden at all. Herp derp.

Joe4evr:
So I found that the lib allows storing more than one TypeReader for a given type, but it only ever uses the `FirstOrDefault()` reader. In `MpGame`, the `PlayerTypeReader` is dependent on the Game and Service types to actually find the correct Player instance to return, so in the following scenario, it would only work for whichever module was registered first. (Maybe a bit of a reach, but I see no reason this *shouldn't* work right.)
```cs
public class CardGamePlayer : Player { ... }

public class PokerModule : MpGameModuleBase<PokerService, PokerGame, CardGamePlayer> { ... }

public class BlackjackModule : MpGameModuleBase<BlackjackService, BlackjackGame, CardGamePlayer> { ... }
```
Probably the best solution is to store the collection of TypeReaders for a given parameter and loop over them until one succeeds. But it would need a sensible error message for if *all* have failed.

src/Discord.Net.Commands/Builders/ModuleBuilder.cs
```diff
@@ -25,6 +26,8 @@ public class ModuleBuilder
         public IReadOnlyList<Attribute> Attributes => _attributes;
         public IReadOnlyList<string> Aliases => _aliases;
 
+        internal Optional<TypeInfo> TypeInfo { get; set; }
```
FiniteReality: Isn't TypeInfo a class? Surely we could compare against null rather than bringing in an extra layer of indirection?

src/Discord.Net.Commands/CommandService.cs
```diff
@@ -139,15 +142,31 @@ public CommandService(CommandServiceConfig config)
                 _moduleLock.Release();
             }
         }
-        private ModuleInfo LoadModuleInternal(ModuleInfo module)
+        private ModuleInfo LoadModuleInternal(ModuleInfo module, IServiceProvider services)
```
FiniteReality: Considering that the signature is now OnModuleBuilding, wouldn't it be better to call it in ModuleBuilder.Build instead?

FiniteReality (changes requested):

src/Discord.Net.Commands/CommandService.cs
Joe4evr: That's a *ton* more stuff to pass the `IServiceProvider` to.
```diff
#  ModuleBuilder.cs
- private ModuleInfo BuildImpl(CommandService service, ModuleInfo parent = null);
+ private ModuleInfo BuildImpl(CommandService service, IServiceProvider services, ModuleInfo parent = null);

- public ModuleInfo Build(CommandService service);
+ public ModuleInfo Build(CommandService service, IServiceProvider services);
 
- internal ModuleInfo Build(CommandService service, ModuleInfo parent);
+ internal ModuleInfo Build(CommandService service, IServiceProvider services, ModuleInfo parent);

# ModuleClassBuilder.cs
- public static Task<Dictionary<Type, ModuleInfo>> BuildAsync(CommandService service, params TypeInfo[] validTypes);
+ public static Task<Dictionary<Type, ModuleInfo>> BuildAsync(CommandService service, IServiceProvider services, params TypeInfo[] validTypes);

- public static async Task<Dictionary<Type, ModuleInfo>> BuildAsync(IEnumerable<TypeInfo> validTypes, CommandService service);
+ public static async Task<Dictionary<Type, ModuleInfo>> BuildAsync(IEnumerable<TypeInfo> validTypes, CommandService service, IServiceProvider services)

# CommandService.cs
- public async Task<ModuleInfo> CreateModuleAsync(string primaryAlias, Action<ModuleBuilder> buildFunc);
+ public async Task<ModuleInfo> CreateModuleAsync(string primaryAlias, Action<ModuleBuilder> buildFunc, IServiceProvider services);

# ModuleInfo.cs
- internal ModuleInfo(ModuleBuilder builder, CommandService service, ModuleInfo parent = null);
+ internal ModuleInfo(ModuleBuilder builder, CommandService service, IServiceProvider services, ModuleInfo parent = null);

- private List<ModuleInfo> BuildSubmodules(ModuleBuilder parent, CommandService service);
+ private List<ModuleInfo> BuildSubmodules(ModuleBuilder parent, CommandService service, IServiceProvider services);
```

Joe4evr: Honestly, why aren't we asking users to set the `IServiceProvider` with the `CommandServiceConfig`? Then wherever the `CommandService` flows, we could access it, and we wouldn't have to change *any* signatures like this. 🤔

FiniteReality: Passing the services in CommandServiceConfig would not allow scoped services to work, unless we forcibly ran commands in a new scope internally.

Joe4evr: Is that the only place where scoped dependencies could be an issue? Because that doesn't sound *that* bad of a change.

Joe4evr:
I decided to push this particular commit for feedback to see if it looks good. As stated in my comment, I found it too bothersome to change signatures everywhere just to pass in the `IServiceProvider` instance - this changes that by storing the instance on the `CommandService`. According to Finite, this changes semantics around Scoped dependencies, so commands now also run in a scope.

**Breaking Change:** Because of this, `ExecuteAsync()` no longer takes `IServiceProvider` as a parameter.

Joe4evr:
Remind me to remove some of the comments if this gets approved to merge.

Joe4evr:
~~So the experimental change needs more work, this currently blocks registering services that are dependent on the `CommandService` instance.~~

~~EDIT: This can be solved by [doing this](https://sharplab.io/#v2:EYLgtghgzgLgpgJwD4AEAMACFBGA3AWACgiUAmDAYQHsxIA7AEwGVEA3ASwGM5q6AzdgHMiAbyIYJWAMxYArAB5qtCIxYIO3ADQYAkmo1wACgiocGiAHwZ9XOFAwiMguDFwYoLtwF8iP4oTJKGnpmNltRcUl2OngEOggAG10bbmNTdnMEazDue0dnVww/SIkUGSUQlLgACgqVUPVbXgFBDE4qfiEAShKHXslsxtyMAF42jpaAOiqoapgAC3YoLoJCAb9iwgB6LYwAVQ8s9vMSchwAdgi1yQAHBHZWCHgMBDgIBg6EgE8g5VUcuAYAD67T+DCgq16ZSw2HO1R61wkYkRA2BoJC9jGdDgAHdfpUAdVsXi6v8hjwJkJ+pJkai6YMDJiMNVOJirM0hABXV4zFnLBH0oorXqbAZ3B5PQE4ABsyQBaTMiCCnUE3LgvNJDQM4zBy16tPpKHOGDonISCVW618RCAA===).~~

Joe4evr:
~~[*I  got it!*](https://sharplab.io/#v2:EYLgtghgzgLgpgJwD4AEAMACFBGA3AWACgiB6EjAGQEtgMBjAewBM4iUBmLAJgwGEGwkAHZMAyogBuVOnH5CAZlQDmRAN5EMmrJwCS4hFJkAFBAyksEGfYbgmzVCxlUYlcGLgxQ3HgL4atHFgArAA8/IIQItbScAA0GHqSMXbmiAB8VknGpqkIAGIQdDAMCACeTi7enlV+hLVsnCg84cJiWayE6oRaGFRC8AhCEAA2CdHZ9o7jtjkOiBWu7hj13QGcLZFtBjEAFBtR7XKKSvQMCsoAlP6aXT090ylzlgC8p+dKAHQPsxbXdz0AfgBb2OX3ajwsBSKJVKAI+OiEEgYAGs4DsYAALKhQK6rf49MhAjAAUTAAAcYKVvpNEPChLBIjI/j0AJAEPHLIgrQJNDAAEWxjAQYgYdFRMF4wyocH6FRWZAwAFUvJZGCw2DwcAB2NR/MkIKgSCDwDAIOAQJhnYblfZbGwYAD6jAiIig7J6+sNxrgpvNlqE1v5gpKIrFbkl0tlTqlMvcRD+POwWp2uJ6t3x0cjMAwryEcAA7kGoELQ+KI7GU+6M87WlAcxg84XbdMdo2+AJWtMjspmVp0/itNTclDimV6zs6HXnhlu0oAK5m6ZQHaZ2PxScXVP4nwXKuaFYeg1Gk04ABsY3BP3ms4XcCXO17mgFxZDolFZZjsron5gsUf7ZdO0YlOQCcT+fs7hQLUGznYZhj3Tk6iIIA) And I can't believe I didn't think of this the first time.~~

FiniteReality:
Hmm... I'm still not sure about this. I like the concept and ideas, but I'm really not sure about providing the services this way.

Joe4evr:
Closing, as this ~~has been merged via @foxbot~~ is now a feature branch.

PR #935, opened 2018-01-16
 Allow parameters to store the collection of TypeReaders for their type 

Joe4evr:
After [saying this](https://github.com/RogueException/Discord.Net/pull/934#issuecomment-357909588) I figured I might as well put my proverbial money where my mouth is and do it myself.

### DIFF
--- a/src/Discord.Net.Commands/Info/ParameterInfo.cs
+++ b/src/Discord.Net.Commands/Info/ParameterInfo.cs
@@ -1,15 +1,15 @@
-using Discord.Commands.Builders;
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
+using Discord.Commands.Builders;
 
 namespace Discord.Commands
 {
     public class ParameterInfo
     {
-        private readonly TypeReader _reader;
+        private readonly IReadOnlyCollection<TypeReader> _readers;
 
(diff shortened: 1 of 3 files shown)


src/Discord.Net.Commands/Info/ParameterInfo.cs
```diff
@@ -59,10 +59,23 @@ internal ParameterInfo(ParameterBuilder builder, CommandInfo command, CommandSer
         public async Task<TypeReaderResult> ParseAsync(ICommandContext context, string input, IServiceProvider services = null)
         {
             services = services ?? EmptyServiceProvider.Instance;
-            return await _reader.ReadAsync(context, input, services).ConfigureAwait(false);
+            var failedResults = new List<TypeReaderResult>();
```
FiniteReality, line 62: I'm not sure that this is the best approach. Due to the asynchronous nature of TypeReaders, we could be performing many expensive operations per typereader per parameter.

Joe4evr: Maybe you're right. I guess the alternative (and maybe more robust, but likely more complex way) would be some additional metadata/key. The question is: "how will the `CommandService` know which extra key to use?"

Joe4evr:
This is basically obsolete as I now think this is a very rare case that can be worked around in user code.

However, there should at least be some value in logging a `Debug` or something for when more than one type reader is registered for a given type, just to inform the user.

PR #941, opened 2018-01-28
Resolve #936

### LINKED ISSUE #936: OverrideTypeReaderAttribute incorrectly applies to other parameters of the same type
```
[glen@UltimateBuilder ~]$ dotnet --version
2.1.3
[glen@UltimateBuilder ~]$ dotnet --info
.NET Command Line Tools (2.1.3)

Product Information:
 Version:            2.1.3
 Commit SHA-1 hash:  a0ca411ca5

Runtime Environment:
 OS Name:     arch
 OS Version:  
 OS Platform: Linux
 RID:         linux-x64
 Base Path:   /opt/dotnet/sdk/2.1.3/

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.4
  Build    : 7f262f453d8c8479b9af91d34c013b3aa05bc1ff

[glen@UltimateBuilder ~]$ 
```

Tested with `Discord.Net` AppVeyor build version `2.0.0-beta-00880` (for commit 05cd1ff85bd4e13ea2aad276eebc09295c9621df).

----

Consider the following two classes, a type reader and a module (the type reader is not registered with the command service for any type - see [the main method](https://gist.github.com/glen3b/81802e0b9b3d2eca2b6ebfc04a466382)):
```cs
// For demonstration purposes; a real "validating" type reader would be more useful
public class NumericStringTypeReader : TypeReader
{
    public override Task<TypeReaderResult> ReadAsync(ICommandContext context, string input, IServiceProvider services)
    {
        if (int.TryParse(input, out int val))
        {
            return Task.FromResult(TypeReaderResult.FromSuccess(input));
        }
        return Task.FromResult(TypeReaderResult.FromError(CommandError.ParseFailed, "Could not parse stringy-integer argument."));
    }
}

public class TestCommandModule : ModuleBase
{
    [Command("stringintecho")]
    public Task StringIntEcho([OverrideTypeReader(typeof(NumericStringTypeReader))] string argument) => ReplyAsync("Message (stringIntEcho): " + argument);
    [Command("stringecho")]
    public Task StringEcho(string argument) => ReplyAsync("Message (stringEcho): " + argument);
}
```

Expected behavior:
- The `stringecho` command echoes any argument given without error
- Due to the presence of the overriding type reader, the `stringintecho` method only accepts and echoes parseable integers

Reality:
- *Both commands* are restricted via the type reader (type reader's invocation evidenced by the distinct parse error message) to accepting only parseable integers
(Note especially the first command invocation in the screenshot)
message screenshot

Notes:
This issue ***does not occur*** if the declaration order of the commands is reversed. However, the order of invocation once the bot has started appears to be irrelevant.

The dependence on the order of declaration leads me to believe that the issue is in [ModuleClassBuilder](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs#L280) - `BuildParameter`'s logic for the default type reader of a parameter, called while the module is built:
```cs
            if (builder.TypeReader == null)
            {
                var readers = service.GetTypeReaders(paramType);
                TypeReader reader = null;

                if (readers != null)
                    reader = readers.FirstOrDefault().Value;
                else
                    reader = service.GetDefaultTypeReader(paramType);

                builder.TypeReader = reader;
            }
```
`GetTypeReaders` returns an `IDictionary<Type, TypeReader>`, which appears to be mapping from `Type`s of TypeReaders to their cached instances. This dictionary is populated in the [GetTypeReader method](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs#L288), called earlier in our example by the `BuildParameters` invocation for our "stringintecho" command. "stringecho"'s parameter is of type `string` with no explicit type reader. Since "stringintecho" has already been built, however, `readers` is non-null and non-empty, containing a `NumericStringTypeReader`. However, the default type reader is desired. Since [GetDefaultTypeReader](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Commands/CommandService.cs#L234) already uses an internal dictionary for caching instances of the default (correct) type reader type, unless I'm missing something, this whole default case can be simplified to:
```cs
            if (builder.TypeReader == null)
            {
                builder.TypeReader = service.GetDefaultTypeReader(paramType);
            }
```
because if a type reader has not been specified, we don't want a random initialized type reader from the cache; we want the default for that type.

Joe4evr:
Since #936 has been open for over a week now, I decided to take a stab at this. I'm *pretty sure* the logic is sound?

### DIFF
--- a/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
@@ -273,15 +273,7 @@ namespace Discord.Commands
 
             if (builder.TypeReader == null)
             {
-                var readers = service.GetTypeReaders(paramType);
-                TypeReader reader = null;
-
-                if (readers != null)
-                    reader = readers.FirstOrDefault().Value;
-                else
-                    reader = service.GetDefaultTypeReader(paramType);
-
-                builder.TypeReader = reader;
+                builder.TypeReader = service.GetDefaultTypeReader(paramType);
             }
         }
 
--- a/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
@@ -273,7 +273,8 @@ namespace Discord.Commands
 
             if (builder.TypeReader == null)
             {
-                builder.TypeReader = service.GetDefaultTypeReader(paramType);
+                builder.TypeReader = service.GetDefaultTypeReader(paramType)
+                    ?? service.GetTypeReaders(paramType)?.FirstOrDefault().Value;
             }
         }
 


FiniteReality (approved):

foxbot (changes requested):
Maybe I'm reading this incorrectly, but this PR seems to favor the default TypeReader over any custom TypeReaders a user would have added?

FiniteReality:
I don't think it would, but it would be worthwhile to test just to be certain.

Joe4evr:
@foxbot *Only* if `TypeReader` was not set before-hand by use of `[OverrideTypeReader]`. In that case, this consults the lib's built-in set of type readers first, and then look for a type reader added to the `CommandService` by the user.

Nensec:
This seems to break overriding of the default type readers at the commandservice registration level and will always take the build in default rather than a newly registered one. I have a custom typereader for IUser (in order to do levensthein distance stuff) which is now no longer called

Joe4evr:
@Nensec #975 and #976.

Nensec:
#975 will not solve the issue I am facing with this change. I am using my own UserTypeReader which will call the original TypeReader for User (which I get through reflection by calling GetDefaultTypeReader() on CommandService) and only runs my code when the original type reader fails to produce a result.

```
var knownTypeReader = _commandService.GetDefaultTypeReader(typeof(IUser));
if (knownTypeReader == null)
    return TypeReaderResult.FromError(CommandError.ParseFailed, $"No typereader found for type {typeof(IUser)}");
else
{
    var knownParseResult = await knownTypeReader.ReadAsync(context, input, services);

    if (knownParseResult.IsSuccess)
    {
        return TypeReaderResult.FromSuccess(knownParseResult.Values);
    }
    else { ... }
```

The reason I am doing this is so I don't have to copy-paste the code from the original type reader. I would prefer to just inherit it so I can just call base.ReadAsync() rather than do this reflection though, it would also make #975 work as proposed.

Joe4evr:
> I am using my own UserTypeReader which will call the original TypeReader for User (which I get through reflection by calling GetDefaultTypeReader() on CommandService)

Well, you shouldn't be using reflection to get at the internal methods and classes in the first place.

I think the best thing to do is to just copy the [`UserTypeReader` code](https://github.com/Joe4evr/Discord.Net/blob/c66dbba11efac63792ce348803f68c35a5c36169/src/Discord.Net.Commands/Readers/UserTypeReader.cs), modify it to do your thing, and using the replace method once #975 is merged in. That way it's both faster *and* more correct.

Also, in the future please use [codeblocks for code samples](https://guides.github.com/features/mastering-markdown/#GitHub-flavored-markdown). Reading unformatted code is disgusting. :nauseated_face:

Nensec:
Why are the typereaders internal anyway? What is the added benefit of it over making them inheritable?

Joe4evr:
That is a good question. I also have a scenario where I want to piggyback off of an internal type reader, so maybe we should just open them up.

PR #943, opened 2018-01-28
Add support for parsing multiple types of quotation marks in commands, Fix #942

### LINKED ISSUE #942: Quotation marks characters
People that are using Discord on iPhone don't have classic quotation marks (Found out that it changed default's quotes after upgrade to iOS 11, its called smart quotes). Is it possible to extend CommandParser to handle these quotes?

For comparison:
Classic quotes - "text" [1](https://unicode-table.com/en/0022/)
Apple's smart quotes - “text” [2](https://unicode-table.com/en/201C/) [3](https://unicode-table.com/en/201D/)

Chris-Johnston:
#942 brings up the issue that non-remainder string parameters of commands cannot be parsed with non-standard double quotation marks (`"`). I added a list of quotation mark aliases to the `CommandServiceConfig`. This means the library can now parse unicode double quotation marks and also culture-specific ones (like `「` and `«`).

I used aliases that were mentioned in the d.net channel, but I'm not certain what the default values should be set to.

I would like to get feedback if anything needs improvement. My discord handle in the channel is `@ChrisJ#8703`.

**Testing Method:**
I wanted to test that a single string could be parsed correctly when wrapped in multiple types of quotation marks. Here's a command that I used:
```cs
[Command("testcmd")]
public async Task TestCmd(string txt, int aNumber)
{
    await ReplyAsync($"{txt} {aNumber}");
}
```

And sample execution yields:
`-testcmd "test" 123` => `test 123` (normal behavior)
`-testcmd “test\“cmd\"“ 123` => `test“cmd" 123` (escaping still works)
`-testcmd «this is a multiple word string» 123` => `this is a multiple word string 123`
`-testcmd «you can mix and match" 456` => `you can mix and match 456`


Emzi0767:
I don't think mix and match is correct behaviour. What happens if: `!test «test" argument»`? Your code would make it throw or parse otherwise incorrectly, whereas it should pass `test" arguments`.

For reference, here's a map of various quote kinds:
- `"` -> `"`
- `'` -> `'`
- `«` -> `»`
- `‘` -> `’`
- `“` -> `”`
- `„` -> `‟`
- `‹` -> `›`
- `‚` -> `‛`
- `《` -> `》`
- `〈` -> `〉`
- `「` -> `」`
- `『` -> `』`
- `〝` -> `〞`
- `﹁` -> `﹂`
- `﹃` -> `﹄`
- `＂` -> `＂`
- `＇` -> `＇`
- `｢` -> `｣`

Chris-Johnston:
For the purposes of outlining how I would approach fixing mismatched quotation marks, here's a gist that shows how I would do it it: https://gist.github.com/Chris-Johnston/37b37e70246562c27f0a42d6bd8eda74

Using this, I've been able to produce the correct output.

image

Do you think that this is the correct approach?

Emzi0767:
I think you should not allow mix and match behaviour at all. Just match the outermost quote.

khionu:
Mmm. Mix and match is an anti-pattern. Additionally, one might use one type of quotes in order to allow regular quotes inside the used quotes

Chris-Johnston:
Changing this to a map was a good idea.
Now using the same command I'm able to do: `-testcmd «all ‘the “quotes 「」”’  ! ! " " » 123` and get back `all ‘the “quotes 「」”’  ! ! " "  123`. 

So it's allowing for different types of quotation marks to be nested without needing to be escaped.
Mix and match is no longer allowed.

src/Discord.Net.Commands/CommandService.cs
```diff
@@ -45,6 +46,7 @@ public CommandService(CommandServiceConfig config)
             _ignoreExtraArgs = config.IgnoreExtraArgs;
             _separatorChar = config.SeparatorChar;
             _defaultRunMode = config.DefaultRunMode;
+            _quotationMarkAliasMap = config.QuotationMarkAliasMap;
```
Joe4evr: Make an `ImmutableDictionary` out of it here?

Well, I guess the consideration is: do we want to guarantee immutable semantics internally, or fake it and take a perf win but also allow the map to be mutated while the client is running (which isn't too hard to do while implemented this way)? cc: @foxbot @FiniteReality 

Either way, I'd make the storage field an `IReadOnlyDictionary`.

src/Discord.Net.Commands/CommandParser.cs
```diff
@@ -13,7 +14,33 @@ private enum ParserPart
             Parameter,
             QuotedParameter
         }
-        
+
+        private static bool isOpenQuote(Dictionary<char,char> map, char c)
```
Joe4evr: * Use PascalCase
* Can take the map in as an `IReadOnlyDictionary`
* Can be made a local function inside `ParseArgsAsync`

(Same goes for `getMatch`)

Joe4evr (changes requested):
Can do with a few minor tweaks.

src/Discord.Net.Commands/CommandService.cs
Joe4evr: For completion's sake: There *is* a third option where we can fake the immutable semantics, but make it much harder for users to modify the map after initialization. (Still wouldn't *completely* prevent it, though.)

Chris-Johnston: For now, I've left this as an `IReadOnlyDictionary`. I found that when the dictionary could be modified by the user externally it didn't break anything, but I could understand that being a poor design decision.

Chris-Johnston: Oh, I didn't see your second comment. I think I ended up doing your third option. I've been able to test it like this:

```cs
CommandServiceConfig config = new CommandServiceConfig();
var d = new Dictionary<char, char>()
{
    { '-', '-' },
    {'\"', '\"' }
};
config.QuotationMarkAliasMap = d;
commands = new CommandService(config);
d.Add('~', '~');
```

Again, I've found that it still works but might not be the intended idea.

Joe4evr: Nah, that's the first thing that I talked about. The third option would be to do
```cs
_quotationMarkAliasMap = new Dictionary(config.QuotationMarkAliasMap);
```

Also, the type on the `CommandServiceConfig` *didn't* need to be changed to `IReadOnlyDictionary`. This now prohibits doing the following:
```cs
new CommandServiceConfig()
{
    QuotationMarkAliasMap =
    {
        { '[', ']' } //for example
    }
};
```

Joe4evr: But anyway, I'd like to hear fox and finite's inputs. Do we want the internals to be semantically correct, or not take a perf hit?

src/Discord.Net.Commands/CommandParser.cs
```diff
@@ -13,7 +14,7 @@ private enum ParserPart
             Parameter,
             QuotedParameter
         }
-        
+
```
FiniteReality: Please remove unnecessary whitespace

src/Discord.Net.Commands/CommandParser.cs
```diff
@@ -24,7 +25,32 @@ private enum ParserPart
...
+                return c == '\"';
+            }
+
+            char GetMatch(IReadOnlyDictionary<char, char> dict, char ch)
+            {
+                // get the corresponding value for the key, if it exists
+                if (dict != null)
+                {
+                    char value;
+                    if (dict.TryGetValue(c, out value))
```
FiniteReality: `out var value` would probably preferable here

src/Discord.Net.Commands/CommandParser.cs
```diff
@@ -97,7 +125,7 @@ private enum ParserPart
                 }
                 else if (curPart == ParserPart.QuotedParameter)
                 {
-                    if (c == '\"')
+                    if(c == matchQuote)
```
FiniteReality: Missing whitespace after if

src/Discord.Net.Commands/CommandParser.cs
```diff
@@ -74,9 +100,11 @@ private enum ParserPart
                             argBuilder.Append(c);
                             continue;
                         }
-                        if (c == '\"')
+                        
+                        if(IsOpenQuote(command._quotationAliases, c))
```
FiniteReality: Missing whitespace after if

src/Discord.Net.Commands/CommandService.cs
FiniteReality: I'd say make an ImmutableDictionary out of it, because that's what we do in other places .

src/Discord.Net.Commands/CommandServiceConfig.cs
```diff
@@ -16,6 +17,30 @@ public class CommandServiceConfig
         /// <summary> Determines whether RunMode.Sync commands should push exceptions up to the caller. </summary>
         public bool ThrowOnError { get; set; } = true;
 
+        /// <summary> Collection of aliases that can wrap strings for command parsing. 
+        /// represents the opening quotation mark and the value is the corresponding closing mark.</summary>
+        public Dictionary<char, char> QuotationMarkAliasMap { get; set; }
+        = new Dictionary<char, char>()
```
FiniteReality: Redundant parentheses here. Do we need the type name too?

src/Discord.Net.Commands/Info/CommandInfo.cs
```diff
@@ -20,6 +20,7 @@ public class CommandInfo
 
         private readonly CommandService _commandService;
         private readonly Func<ICommandContext, object[], IServiceProvider, CommandInfo, Task> _action;
+        internal readonly IReadOnlyDictionary<char,char> _quotationAliases;
```
FiniteReality: IMO this should not be stored on the command, but I'm not sure if we have the ability to pass the command service to the parser where ParseArgsAsync is called. Might be worth taking a look?

FiniteReality (changes requested):

src/Discord.Net.Commands/CommandServiceConfig.cs
Joe4evr: Type name is required until dotnet/csharplang#100 lands. Do we omit the parentheses in other places too?

FiniteReality: I'm pretty sure we omit the parentheses where possible.

src/Discord.Net.Commands/Info/CommandInfo.cs
Chris-Johnston: I looked at this again, and found that I could use the similar behavior of `IgnoreExtraArgs`.

https://github.com/Chris-Johnston/Discord.Net/blob/46e9cf64f3d3635de872720c308e922b13ab8114/src/Discord.Net.Commands/Info/CommandInfo.cs#L124

src/Discord.Net.Commands/CommandServiceConfig.cs
Chris-Johnston: GitHub doesn't seem to show this as outdated correctly, but please see commit 8187f35975031971b17818aae516fdfaf5d2eb36

Joe4evr:
LGTM :shipit:

src/Discord.Net.Commands/CommandServiceConfig.cs
```diff
@@ -1,4 +1,5 @@
-﻿namespace Discord.Commands
+﻿using System.Collections.Generic;
+namespace Discord.Commands
```
FiniteReality: Newline between these for consistency

FiniteReality (changes requested):
Just a few more niggling things, and it'll be ready to merge

src/Discord.Net.Commands/CommandParser.cs
```diff
@@ -12,9 +13,8 @@ private enum ParserPart
             None,
             Parameter,
             QuotedParameter
-        }
-        
-        public static async Task<ParseResult> ParseArgsAsync(CommandInfo command, ICommandContext context, bool ignoreExtraArgs, IServiceProvider services, string input, int startPos)
+        }
```
SubZero0: Missing new line between enum and method

Chris-Johnston: I must have mistook "remove unnecessary whitespace" as the line instead of the space after. Fixed now.

FiniteReality (approved):

src/Discord.Net.Commands/CommandParser.cs
```diff
@@ -24,7 +25,31 @@ private enum ParserPart
...
+                if (dict != null)
+                    return dict.ContainsKey(ch);
+                // or otherwise if it is the default double quote
+                return c == '\"';
+            }
+
+            char GetMatch(IReadOnlyDictionary<char, char> dict, char ch)
+            {
+                // get the corresponding value for the key, if it exists
+                if (dict != null)
```
AntiTcb: This could just be reduced down to `(dict != null && dict.TryGetValue(c, out var value))`.

ufcpp:
Here is open/close and initial/final quote punctuation pairs generated by https://gist.github.com/ufcpp/5b2cf9a9bf7d0b8743714a0b88f7edc5

```
()༺༻༼༽᚛᚜‘’“”‹›⁅⁆⌈⌉⌊⌋❨❩❪❫❬❭❮❯❰❱❲❳❴❵⟅⟆⟦⟧⟨⟩⟪⟫⟬⟭⟮⟯⦃⦄⦅⦆⦇⦈⦉⦊⦋⦌⦍⦎⦏⦐⦑⦒⦓⦔⦕⦖⦗⦘⧘⧙⧚⧛⧼⧽⸂⸃⸄⸅⸉⸊⸌⸍⸜⸝⸠⸡⸢⸣⸤⸥⸦⸧⸨⸩〈〉《》「」『』【】〔〕〖〗〘〙〚〛〝〞
```

src/Discord.Net.Commands/CommandServiceConfig.cs
```diff
@@ -38,7 +38,56 @@ public class CommandServiceConfig
...
+            {'⸠', '⸡' },
+            {'⸢', '⸣' },
+            {'⸤', '⸥' },
+            {'⸦', '⸧' },
+            {'⸨', '⸩' },
+            {'【', '】'},
+            {'〔', '〕' },
+            {'〖', '〗' },
+            {'〘', '〙' },
+            {'〚', '〛' }
```
FiniteReality: IMO if we're adding a large dict of things like this, it might be better to use a separate file for it instead. I'm not sure though.

Chris-Johnston: Do you mean deserializing the dictionary from another file, or another file in the source that defines the default value of this dictionary?

I think that including all of these by default is a good idea. Alternatively we could shorten this list and include a more complete one in the documentation.

FiniteReality: Another file in source, yeah, like the emoji map text template. It may be possible for us to re-use some of the code suggested by @ufcpp so that we can generate these automatically and keep them up-to-date when Unicode inevitably changes.

Chris-Johnston: It looks like that gist uses features that aren't supported in .netstandard1.1 (or 1.3) https://docs.microsoft.com/en-us/dotnet/api/system.char.getunicodecategory?view=netcore-2.0#Applies_to 
https://docs.microsoft.com/en-us/dotnet/api/system.string.normalize?view=netcore-2.0#Applies_to

So for now it may be best to define this dictionary in another file and leave it, unless/until we target netstandard2

Chris-Johnston:
#1043 ~~~replaces ns1.3 with ns2.0~~~ targets netstandard2.0. if it is merged, we will be able to support generating quotation mark pairs programmatically, see the comment thread immediately above

Edit: I misread. Ns2.0 is added here, but not replacing ns1.3. So I'm not sure if the generation would be viable.

src/Discord.Net.Commands/CommandParser.cs
acid-chicken: @AntiTcb How about `(dict?.TryGetValue(c ,out var value) ?? false)`?

Joe4evr: Or just make sure that the collection that gets passed in is always initialized so that it's at least empty?

Chris-Johnston: I like the idea of assuming that the collection is always initialized. Added in 820f9e36483d63df64ae651f862d0226c4c8e52a

PR #944, opened 2018-01-30
Remove IGuild.DownloadUsersAsync() from SocketGuild

SubZero0:
SocketGuild implements this method and shouldn't throw. If a SocketGuild is passed as a IGuild, it'll throw.


FiniteReality (approved):
Is this tested? Should be good to merge if so.

SubZero0:
Yes. How it currently is:
error

And without that line:
fixed

PR #945, opened 2018-01-30
Fix SocketCategoryChannel properties

SubZero0:
Probably forgotten when copied from another file.

### DIFF
--- a/src/Discord.Net.WebSocket/Entities/Channels/SocketCategoryChannel.cs
+++ b/src/Discord.Net.WebSocket/Entities/Channels/SocketCategoryChannel.cs
@@ -15,7 +15,7 @@ namespace Discord.WebSocket
     public class SocketCategoryChannel : SocketGuildChannel, ICategoryChannel
     {
         public override IReadOnlyCollection<SocketGuildUser> Users
-            => Guild.Users.Where(x => x.VoiceChannel?.Id == Id).ToImmutableArray();
+            => throw new NotSupportedException();
 
         public IReadOnlyCollection<SocketGuildChannel> Channels
             => Guild.Channels.Where(x => x.CategoryId == CategoryId).ToImmutableArray();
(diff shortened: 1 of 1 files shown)


src/Discord.Net.WebSocket/Entities/Channels/SocketCategoryChannel.cs
```diff
@@ -15,7 +15,7 @@ namespace Discord.WebSocket
     public class SocketCategoryChannel : SocketGuildChannel, ICategoryChannel
     {
         public override IReadOnlyCollection<SocketGuildUser> Users
-            => Guild.Users.Where(x => x.VoiceChannel?.Id == Id).ToImmutableArray();
+            => throw new NotSupportedException();
```
FiniteReality: Wouldn't it be better to use something like SocketTextChannels.Users, and resolve the permissions instead of throwing an exception?

FiniteReality (changes requested):

src/Discord.Net.WebSocket/Entities/Channels/SocketCategoryChannel.cs
SubZero0: Hm, makes sense. I'll update it.

SubZero0: As talked in the dev guild, resolving the permissions could be misleading, since if the users don't have read access to any channels inside, they won't be able to see the category too.

A suggestion would be `Guild.Channels.Where(x => x.CategoryId == Id).SelectMany(x => x.Users).Distinct().ToImmutableArray();` that would list all users that can see at least one channel (and the category consequently) but doesn't mean all of them can see all the channels inside it.

foxbot: That seems fine to me.

SubZero0:
Just a reminder that now this PR depends on #952

PR #953, opened 2018-02-10
Addition of the PostActionAttribute

ghost:
### PostActionAttribute
This addition will allow users to use attributes to for code that would have previously been put into the `AfterExecute` module method. Overriding this method will apply it to every command in the module. Providing command specific behavior is much more complicated, and requires dealing with the `CommandInfo` argument with runtime checks. 

Furthermore, if the same `AfterExecute` method is to be reused in several different modules, code repetition is inevitable. By using attributes, it provides the user with much more flexibility and control. This feature essentially boils down to the `Precondition` counterpart.

### Usage
Attribute:
```cs
public class Cooldown : PostActionAttribute
{
    public override Task ExecuteAsync(ICommandContext context, CommandInfo command,
                                      IResult result, IServiceProvider services)
    {
        if (result.IsSuccess)
            return ApplyCooldownAsync(context.User, command);

        return Task.CompletedTask;
    }
}
```
Command:
```cs
[Cooldown]
[Command("Claim")]
public Task ClaimAsync() => Task.ComletedTask;
```

### Additions
All additions made to the library have mimicked the `Precondition` class in every way, with the exception being when the `ExecutePostActionsAsync` method was called in the `CommandService`. No breaking changes have been made in this PR.

### DIFF
--- a/src/Discord.Net.Commands/Attributes/PostActionAttribute.cs
+++ b/src/Discord.Net.Commands/Attributes/PostActionAttribute.cs
@@ -1,0 +1,11 @@
+using System;
+using System.Threading.Tasks;
+
+namespace Discord.Commands
+{
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class, AllowMultiple = true, Inherited = true)]
+    public abstract class PostActionAttribute : Attribute
+    {
+        public abstract Task ExecuteAsync(ICommandContext context, CommandInfo command, IResult result, IServiceProvider services);
+    }
+}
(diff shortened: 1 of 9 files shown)


Joe4evr:
After discussion in the Discord the other day, this is more likely to get done by implementing a duplex pipeline instead (just like what ASP.NET Core does) which is (in most our opinion) a much better design to solve the underlying problem.

I think you said you're okay with that (feel free to rebut if that isn't the case). Still, thanks for getting the ball rolling on the topic or we might not have started thinking about solving it this way.

PR #957, opened 2018-02-15
Add request info to HttpException & RateLimitedException

AntiTcb:
## Overview
Adds a Request property to the exception, so when caught, information of the request can be handled by the user. Mostly useful for logging when you don't have the command service up to a level where the request being made isn't logged prior to the exception being thrown.

### DIFF
--- a/src/Discord.Net.Core/Net/IRequest.cs
+++ b/src/Discord.Net.Core/Net/IRequest.cs
@@ -1,14 +1,10 @@
 using System;
-using System.IO;
-using System.Threading.Tasks;
 
 namespace Discord.Net
 {
     public interface IRequest
     {
         DateTimeOffset? TimeoutAt { get; }
-        TaskCompletionSource<Stream> Promise { get; }
         RequestOptions Options { get; }
-        
     }
 }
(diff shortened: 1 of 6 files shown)


src/Discord.Net.Core/Net/IRequest.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Discord.Net
+{
+    public interface IRequest
+    {
+        DateTimeOffset? TimeoutAt { get; }
+        TaskCompletionSource<Stream> Promise { get; }
```
FiniteReality: I don't think we should expose the stream. Sounds like a gross implementation detail.

FiniteReality (changes requested):
I like the idea, but I'm not sure about exposing some of the things we currently expose.

src/Discord.Net.Core/Net/IRequest.cs
AntiTcb: So, also change the implementations' `Promise` property to internal?

foxbot: I wouldn't change it, just because RestRequest is already a public class, and the promise is a public member. While it's not like the consumer could ever really do anything with a RestRequest class, I don't want to introduce unnecessary breaking changes.

AntiTcb:
Probably no reason not to. Can't provide any useful information in the default error message that's provided by it, but I don't see that as an argument against letting a user get it themselves when they catch the exception.

PR #958, opened 2018-02-21
Set usage of TokenType.User as an error rather than a warning.

Joe4evr:
I think waiting for 3.0 to do this is taking too long. Discord disallowed external API access with user tokens since November, so we should be stopping users *completely* from continuing that practice ASAP.


AntiTcb:
They're going to be shut out of it in the future anyway, what's the point in delaying it when using the endpoints is a risk to their account's status?

Still34:
Agreed. It is going to be shut off anyways.

PR #962, opened 2018-02-23
Attempts to resolve #961

### LINKED ISSUE #961: Proposal: Expose `Shards` property from `GetBotGatewayResponse` object in some way.
Because of the way I have been developing my bot, I wanted it from the very beginning to be able to retrieve the correct shard count recommended by discord. During the login process the data is retrieved but it is not exposed in the public API.

I would like to propose exposing this property or it's data in the public API in someone wanted to use it when starting their shards manually without using the `DiscordShardedClient` object.

If this does not fit the "meta" is there an alternative method I can use to acquire this information myself?

Seeker1437:
fixes #961

### DIFF
--- a/src/Discord.Net.Rest/ClientHelper.cs
+++ b/src/Discord.Net.Rest/ClientHelper.cs
@@ -1,4 +1,4 @@
-﻿using Discord.API.Rest;
+using Discord.API.Rest;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.IO;
@@ -162,6 +162,12 @@ namespace Discord.Rest
         {
             var models = await client.ApiClient.GetVoiceRegionsAsync(options).ConfigureAwait(false);
             return models.Select(x => RestVoiceRegion.Create(client, x)).FirstOrDefault(x => x.Id == id);
+        }
+
+        public static async Task<int> GetRecommendShardCount(BaseDiscordClient client, RequestOptions options)
+        {
+            var response = await client.ApiClient.GetBotGatewayAsync(options);
+            return response.Shards;
         }
     }
 }
--- a/src/Discord.Net.Rest/ClientHelper.cs
+++ b/src/Discord.Net.Rest/ClientHelper.cs
@@ -164,7 +164,7 @@ namespace Discord.Rest
             return models.Select(x => RestVoiceRegion.Create(client, x)).FirstOrDefault(x => x.Id == id);
         }
 
-        public static async Task<int> GetRecommendShardCount(BaseDiscordClient client, RequestOptions options)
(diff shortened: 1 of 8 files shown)


Seeker1437:
I think everything is ready to be reviewed now, I double checked after work to make sure.

src/Discord.Net.Rest/ClientHelper.cs
```diff
@@ -163,5 +163,11 @@ internal static class ClientHelper
             var models = await client.ApiClient.GetVoiceRegionsAsync(options).ConfigureAwait(false);
             return models.Select(x => RestVoiceRegion.Create(client, x)).FirstOrDefault(x => x.Id == id);
         }
+
+        public static async Task<int> GetRecommendShardCountAsync(BaseDiscordClient client, RequestOptions options)
+        {
+            var response = await client.ApiClient.GetBotGatewayAsync(options);
```
foxbot: Needs ConfigureAwait

src/Discord.Net.Rest/DiscordRestApiClient.cs
```diff
@@ -459,6 +471,8 @@ internal virtual void Dispose(bool disposing)
                 endpoint = () => $"channels/{channelId}/messages?limit={limit}";
             return await SendAsync<IReadOnlyCollection<Message>>("GET", endpoint, ids, options: options).ConfigureAwait(false);
         }
+
```
foxbot: minor nitpick but this whitespace seems unnecessary

src/Discord.Net.Rest/DiscordRestApiClient.cs
```diff
@@ -277,6 +277,18 @@ internal virtual void Dispose(bool disposing)
             await SendAsync("GET", () => "auth/login", new BucketIds(), options: options).ConfigureAwait(false);
         }
 
+        //Gateway
```
foxbot, line 280: Why were these moved to REST instead of being left with gateway?

foxbot (changes requested):

src/Discord.Net.Rest/DiscordRestApiClient.cs
Seeker1437: I was told it was a better place for them because they themselves are REST requests and should be accessible from that client as well.

I this it should be this way too. As before mentioned, since I have my bot set up in a more advanced configuration, it would be nice to be able to just access the data from the rest client so I can pass it to the individual clients who will then be establishing docket client connections from there.

I with it being there it is more open and therefor more flexible, even though this appears to be an edge case.

foxbot: Seems fair to me

foxbot (approved):

Issue #965, opened 2018-02-28
Permissions utils implementation is incorrect


FiniteReality:
https://github.com/RogueException/Discord.Net/blob/edfbd055bb3ee62c565370af3abb84ab7214d2d6/src/Discord.Net.Core/Utils/Permissions.cs#L83 should be:
```cs
private static bool HasFlag(ulong value, ulong flag) => (value & flag) == flag;
```

https://github.com/RogueException/Discord.Net/blob/edfbd055bb3ee62c565370af3abb84ab7214d2d6/src/Discord.Net.Core/Utils/Permissions.cs#L163-L166 should be:
```cs
resolvedPermissions &= ~(ulong)ChannelPermission.SendTTSMessages;
resolvedPermissions &= ~(ulong)ChannelPermission.MentionEveryone;
resolvedPermissions &= ~(ulong)ChannelPermission.EmbedLinks;
resolvedPermissions &= ~(ulong)ChannelPermission.AttachFiles;
```

PR #967, opened 2018-02-28
Add more tests for Permissions class

Chris-Johnston:
Added more tests to the `Permissions` class that tries to cover more of the methods than were already being tested. These are similar to the existing tests for `GuildPermissions` and `ChannelPermissions` but work more closely with `Permissions` by converting them to `ulong` raw values first.

Added tests for `OverwritePermissions` that checks that values are set correctly and that flags can be toggled correctly, including when both allow and deny are set for the same bit.

Tried to make tests fail with issue described in #965 but I was unable to find a case that broke the tests. Would appreciate input and any suggestions on making a test case for that. (CC: @FiniteReality )

My discord handle is `@ChrisJ#8703` should you need to ping/DM me.

### DIFF
--- a/test/Discord.Net.Tests/Discord.Net.Tests.csproj
+++ b/test/Discord.Net.Tests/Discord.Net.Tests.csproj
@@ -27,7 +27,4 @@
     <PackageReference Include="xunit.runner.visualstudio" Version="2.2.0" />
     <PackageReference Include="xunit.runner.reporters" Version="2.2.0" />
   </ItemGroup>
-  <ItemGroup>
-    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
-  </ItemGroup>
 </Project>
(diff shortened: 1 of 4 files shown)


test/Discord.Net.Tests/Discord.Net.Tests.csproj
```diff
@@ -24,4 +27,7 @@
     <PackageReference Include="xunit.runner.visualstudio" Version="2.2.0" />
     <PackageReference Include="xunit.runner.reporters" Version="2.2.0" />
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
```
foxbot: what's this?

Chris-Johnston: I'm not sure, I don't recall adding anything explicitly. I'll remove it from the file.

Chris-Johnston:
I may be making a bad habit out of doing more than one thing in a PR.
The `GuildPermissions` and `ChannelPermissions` classes didn't need to be part of the `Test` class, and would fail if the test fixture wasn't supplied. I moved them into their own classes.

PR #969, opened 2018-03-02
Allow nested ModuleBase classes to be built when declared from non-module classes.

AntiTcb:
## Overview

This will allow a module such as:
```cs
public class Foo 
{
    public class Module : ModuleBase { }
}
```
to be built properly by the command service. This would previously prevent being able to use the command service in places where you're executing code contained within a wrapper class, such as in LINQPad, or if, for whatever organizational reason, you want to have a module be a nested class of a non-module.


Joe4evr:
For no real reason other than "why not?", does this also allow the outer class being `static`?
```cs
public static class Foo 
{
    public class Module : ModuleBase { }
}
```

AntiTcb:
Yes it does.

foxbot (approved):

PR #970, opened 2018-03-02
Add Spotify track support

Still34:
This PR: 
1. Implements rudimentary `SpotifyGame` support
2. Sets a default `appId` value for `GameAssets` because Spotify does not return an application ID (not during my testing anyways)

There is one thing I'm unsure about, though, that is the inclusion of Spotify CDN URL. Discord does not appear to host the album image, which makes sense, but that requires us to use Spotify CDN to get the album URL. Should we include the full album URL, or just the track ID?

### DIFF
--- a/src/Discord.Net.Core/Entities/Activities/Game.cs
+++ b/src/Discord.Net.Core/Entities/Activities/Game.cs
@@ -1,14 +1,15 @@
-﻿using System.Diagnostics;
+using System.Diagnostics;
 
 namespace Discord
 {
     [DebuggerDisplay(@"{DebuggerDisplay,nq}")]
     public class Game : IActivity
     {
+        internal Game() { }
+
         public string Name { get; internal set; }
         public ActivityType Type { get; internal set; }
 
-        internal Game() { }
         public Game(string name, ActivityType type = ActivityType.Playing)
         {
             Name = name;
--- a/src/Discord.Net.Core/Entities/Activities/Game.cs
+++ b/src/Discord.Net.Core/Entities/Activities/Game.cs
@@ -5,11 +5,10 @@ namespace Discord
     [DebuggerDisplay(@"{DebuggerDisplay,nq}")]
     public class Game : IActivity
     {
-        internal Game() { }
-
         public string Name { get; internal set; }
         public ActivityType Type { get; internal set; }
 
+        internal Game() { }
         public Game(string name, ActivityType type = ActivityType.Playing)
         {
             Name = name;
--- a/src/Discord.Net.Core/Entities/Activities/GameAsset.cs
+++ b/src/Discord.Net.Core/Entities/Activities/GameAsset.cs
@@ -1,15 +1,15 @@
-﻿namespace Discord
+namespace Discord
 {
     public class GameAsset
     {
         internal GameAsset() { }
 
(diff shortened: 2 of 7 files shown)


src/Discord.Net.Core/Entities/Activities/Game.cs
```diff
@@ -1,14 +1,15 @@
-﻿using System.Diagnostics;
+using System.Diagnostics;
 
 namespace Discord
 {
     [DebuggerDisplay(@"{DebuggerDisplay,nq}")]
     public class Game : IActivity
     {
+        internal Game() { }
```
FiniteReality: Why move the ctor?

src/Discord.Net.Core/Entities/Activities/SpotifyGame.cs
```diff
@@ -0,0 +1,23 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Discord
+{
+    [DebuggerDisplay(@"{DebuggerDisplay,nq}")]
+    public class SpotifyGame : Game
+    {
+        internal SpotifyGame() { }
```
FiniteReality: We generally put properties and fields before constructors.

src/Discord.Net.WebSocket/Extensions/EntityExtensions.cs
```diff
@@ -34,7 +55,7 @@ public static IActivity ToEntity(this API.Game model)
         }
 
         // (Small, Large)
-        public static GameAsset[] ToEntity(this API.GameAssets model, ulong appId)
+        public static GameAsset[] ToEntity(this API.GameAssets model, ulong appId = 0)
```
FiniteReality: IMO instead of specifying 0 we should make the app id nullable.

FiniteReality (changes requested):
Looking alright, but please configure your editor to not format unnecessary files if no code changes have been made. (RichGame for example)

src/Discord.Net.Core/Entities/Activities/Game.cs
Still34: Mainly because I saw other files like `GameAsset`'s ctor was at front. I'll change that back though.

src/Discord.Net.Core/Entities/Activities/GameAsset.cs
```diff
@@ -1,15 +1,15 @@
...
 
-        internal ulong ApplicationId { get; set; }
+        internal ulong? ApplicationId { get; set; }
         
         public string Text { get; internal set; }
         public string ImageId { get; internal set; }
         
         public string GetImageUrl(ImageFormat format = ImageFormat.Auto, ushort size = 128)
-            => CDN.GetRichAssetUrl(ApplicationId, ImageId, size, format);
+            => ApplicationId.HasValue ? CDN.GetRichAssetUrl(ApplicationId.Value, ImageId, size, format) : null;
```
FiniteReality, line 13: @foxbot Should we return null or throw here?

foxbot: I think returning nil is fine, we do that elsewhere in the library when an asset doesn't exist (current behavior of GetAvatarUrl)

PR #972, opened 2018-03-04
Add sample bots to the solution

foxbot:
# Summary

This change adds two sample bots to the repository.

## Additions

- Added a minimal, barebones sample to show how to start a Discord.Net bot
- Added a minimal bot using the Commands extension

# Motivations

I think one of the biggest hurdles to get over when starting a Discord.Net bot is just figuring out how to create a client. Unlike other .NET libraries, I wouldn't consider this library to be very intuitive or easy to just feel your way around when starting out.

I've noticed that other .NET projects (such as [Blazor](https://github.com/aspnet/blazor)), as well as [other](https://github.com/zeyla/serenity/tree/master/examples) [Discord](https://github.com/Rapptz/discord.py/tree/async/examples) [libraries](https://github.com/DV8FromTheWorld/JDA/tree/master/src/examples/java) include example bots directly within their main codebase.

This has the added advantages of ensuring that there's always an up-to-date example bot, rather than relying on community members to maintain their own.

### DIFF
--- a/samples/02_commands_framework/Modules/PublicModule.cs
+++ b/samples/02_commands_framework/Modules/PublicModule.cs
@@ -36,6 +36,19 @@ namespace _02_commands_framework.Modules
             await ReplyAsync(user.ToString());
         }
 
+        // Ban a user
+        [Command("ban")]
+        [RequireContext(ContextType.Guild)]
+        // make sure the user invoking the command can ban
+        [RequireUserPermission(GuildPermission.BanMembers)]
+        // make sure the bot itself can ban
+        [RequireBotPermission(GuildPermission.BanMembers)]
+        public async Task BanUserAsync(IGuildUser user, [Remainder] string reason = null)
+        {
+            await user.Guild.AddBanAsync(user, reason: reason);
+            await ReplyAsync("ok!");
+        }
+
         // [Remainder] takes the rest of the command's arguments as one argument, rather than splitting every space
         [Command("echo")]
         public Task EchoAsync([Remainder] string text)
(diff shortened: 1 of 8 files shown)


samples/02_commands_framework/Modules/PublicModule.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        // [Remainder] takes the rest of the command's arguments as one argument, rather than splitting every space
+        [Command("echo")]
+        public Task EchoAsync([Remainder] string text)
+            // Insert a ZWSP before the text to prevent triggering other bots!
+            => ReplyAsync('\u200B' + text);
+
+        // 'params' will parse space-separated elements into a list
+        public Task ListAsync(params string[] objects)
```
AntiTcb: Missed a command tag.

samples/02_commands_framework/Modules/PublicModule.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            // Get a stream containing an image of a cat
+            var stream = await PictureService.GetCatPictureAsync();
+            // Streams must be seeked to their beginning before being uploaded!
+            stream.Seek(0, SeekOrigin.Begin);
+            await Context.Channel.SendFileAsync(stream, "cat.png");
+        }
+
+        // Get info on a user, or the user who invoked the command if one is not specified
+        [Command("userinfo")]
```
AntiTcb, line 31: Could be a good place to show off some preconditions. `RequireContext` and `RequireUserPermission` would be thematically correct.

PR #976, opened 2018-03-09
Fix override attribute and rollback to old behavior

SubZero0:
This will fix the override issue ( #936 ) and rollback the behavior change introduced ( #941 )

The actual issue was that the OverrideTypeReaderAttribute was actually adding the reader to the command service (so it would be registered as a global type reader and replace the current one if it existed [if I add a reader for a non-primitive type and override it somewhere, it'll still present the same issue]).

This PR will turn the previously breaking change into a simple bug fix (no behavior change).

PS: Sorry for the commits, I couldn't properly update my fork. I need to git gud with this.

### DIFF
--- a/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
@@ -42,8 +42,8 @@ namespace Discord.Commands
         }
 
 
-        public static Task<Dictionary<Type, ModuleInfo>> BuildAsync(CommandService service, params TypeInfo[] validTypes) => BuildAsync(validTypes, service);
-        public static async Task<Dictionary<Type, ModuleInfo>> BuildAsync(IEnumerable<TypeInfo> validTypes, CommandService service)
+        public static Task<Dictionary<Type, ModuleInfo>> BuildAsync(CommandService service, IServiceProvider services, params TypeInfo[] validTypes) => BuildAsync(validTypes, service, services);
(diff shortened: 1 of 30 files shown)


SubZero0:
Just forgot to mention that if this PR is merged, #975 will become obsolete.
And right now the issue with override still exists, it's just for added types (that don't have a default type reader). And this PR will really fix it now.

src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
```diff
@@ -246,7 +246,7 @@ private static void BuildParameter(ParameterBuilder builder, System.Reflection.P
                         builder.Summary = summary.Text;
                         break;
                     case OverrideTypeReaderAttribute typeReader:
-                        builder.TypeReader = GetTypeReader(service, paramType, typeReader.TypeReader, services);
+                        builder.TypeReader = ReflectionUtils.CreateObject<TypeReader>(typeReader.TypeReader.GetTypeInfo(), service, EmptyServiceProvider.Instance);
```
foxbot: I figure this should probably be using the `IServiceProvider` passed to `BuildParameter`, no?

foxbot (changes requested):

SubZero0:
After looking this PR again, my premise wasn't true, since the CommandService will store more than one reader. But since the built-in was registered in another list (_defaultTypeReaders) and the override attribute add the type reader to the "user added list" (_typeReaders), it caused the unwanted override for all.

Issue #796, opened 2017-08-24
Allow Embeds to be sent in the same message as an attachment.


danny-may:
Currently there is no way (that I can see) of sending a message that contains both an attachment and an embed through d.net. 

It is definitely possible to do through the discord API, as shown by [another bot](http://imgur.com/MCZxpLe) however the IMessageChannel interface seems to only allow sending of text along with an attachment.

foxbot:
This was merged in e9f9b484b660bc682239c8de2afb3efd84f60ff1.

PR #978, opened 2018-03-10
Allow attaching embeds alongside a file upload.

WamWooWam:
# Abstract
This PR adds support for attaching embeds alongside file uploads, hopefully fixing #796.

## Additions
Added an `Embed` parameter to `IMessageChannel#SendFileAsync` and modified the rest of the code handling this to allow it, as such it should work now in most cases.

## Drawbacks
This is, naturally, a breaking change, which may cause issues for people passing `RequestOptions` to these methods, but shouldn't effect anyone else.

### DIFF
--- a/src/Discord.Net.Rest/API/Rest/UploadFileParams.cs
+++ b/src/Discord.Net.Rest/API/Rest/UploadFileParams.cs
@@ -1,18 +1,26 @@
-﻿#pragma warning disable CS1591
+#pragma warning disable CS1591
+using Discord.Net.Converters;
 using Discord.Net.Rest;
+using Newtonsoft.Json;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
+using System.Text;
 
 namespace Discord.API.Rest
 {
     internal class UploadFileParams
     {
+        // Fuckin Discord man.
+        private static JsonSerializer _serializer = new JsonSerializer { ContractResolver = new DiscordContractResolver() };
+
         public Stream File { get; }
 
         public Optional<string> Filename { get; set; }
         public Optional<string> Content { get; set; }
         public Optional<string> Nonce { get; set; }
         public Optional<bool> IsTTS { get; set; }
+        public Optional<Embed> Embed { get; set; }
 
         public UploadFileParams(Stream file)
         {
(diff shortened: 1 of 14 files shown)


src/Discord.Net.Rest/API/Rest/UploadFileParams.cs
```diff
@@ -29,6 +37,17 @@ public UploadFileParams(Stream file)
                 d["tts"] = IsTTS.Value.ToString();
             if (Nonce.IsSpecified)
                 d["nonce"] = Nonce.Value;
+            if (Embed.IsSpecified)
+            {
+                var sb = new StringBuilder(256);
```
foxbot, line 41: Is there any cleaner approach here? The DefaultRestClient seems to support serializing Streams here, could use a MemoryStream+StreamWriter instead?

src/Discord.Net.Rest/API/Rest/UploadFileParams.cs
```diff
@@ -1,18 +1,26 @@
...
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
+using System.Text;
 
 namespace Discord.API.Rest
 {
     internal class UploadFileParams
     {
+        // Fuckin Discord man.
```
foxbot: i'd replace the sentimental comments with something more informational, like "needed to serialize embeds" or something

src/Discord.Net.Rest/API/Rest/UploadFileParams.cs
```diff
@@ -29,6 +37,17 @@ public UploadFileParams(Stream file)
                 d["tts"] = IsTTS.Value.ToString();
             if (Nonce.IsSpecified)
                 d["nonce"] = Nonce.Value;
+            if (Embed.IsSpecified)
+            {
+                var sb = new StringBuilder(256);
+                using (TextWriter text = new StringWriter(sb, CultureInfo.InvariantCulture))
+                using (JsonWriter writer = new JsonTextWriter(text))
+                {
+                    // I apologise for the mess I've created here, god damn is this annoying.
```
foxbot: don't apologize, that's what `git blame` is for

src/Discord.Net.Rest/API/Rest/UploadFileParams.cs
```diff
@@ -29,6 +37,17 @@ public UploadFileParams(Stream file)
...
+                    _serializer.Serialize(writer, new Dictionary<string, object>() { { "embed", Embed.Value } });
```
foxbot: I believe we generally prefer Dictionary-initializer style (`["key"] = "value"`)

foxbot (changes requested):

src/Discord.Net.Rest/API/Rest/UploadFileParams.cs
WamWooWam: Not really, the embed data needs to be URL encoded, something I'm not sure I can do from a MemoryStream without even more mess

PR #973, opened 2018-03-04
added UserDefaultAvatar to IUser

Fyers:
Sorry for the automatic formatting

PR contains request from Issue #971

PR #979, opened 2018-03-11
Added the ability to get the default avatar

### LINKED ISSUE #971: Return default avatar for user
At the moment, Discord.NET does not care about user's default avatar at all. If no avatar ID is set, the `IUser#GetAvatarUrl` method returns `null`. It is possible to retrieve the user's default URL based on their discriminator with `discrimValue % 5`.

There should be an option for people to retrieve this value, and I propose these solutions.

a. Implement a `DefaultAvatarUrl` property in `IUser`.
b. Add a bool `defaultOnNone` or something under the `GetAvatarUrl` method.
c. Enforce the return of a default URL when the avatar ID is `null`.

I personally prefer a) as it allows people who has implemented a null fallback for their `GetAvatarUrl` to not be affected, and for whatever reason, allows people to retrieve the default URL for those who already has an avatar.

WorkingRobot:
This PR fixes #971 and adds the ability to get the default avatar of a user. This may be a breaking change to some, if people are checking against the return value of `GetAvatarUrl`. It will ignore any parameters passed to the method if returning the default avatar.

### DIFF
--- a/src/Discord.Net.Core/DiscordConfig.cs
+++ b/src/Discord.Net.Core/DiscordConfig.cs
@@ -4,16 +4,17 @@ namespace Discord
 {
     public class DiscordConfig
     {
-        public const int APIVersion = 6;        
+        public const int APIVersion = 6;
         public static string Version { get; } =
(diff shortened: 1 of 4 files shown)


AntiTcb:
Duplicate of #973

aStonePenguin:
@AntiTcb Its the different implementation I suggested in #971

PR #984, opened 2018-03-16
Expose VoiceServerUpdate events

### LINKED ISSUE #983: Expose VoiceServerUpdate event
The VoiceServerUpdate event should be exposed for those who would like to make use of other voice-sending clients (e.g. Lavalink)

devoxin:
Resolves #983 by exposing VoiceServerUpdates to the user.

This should not break any existing functionality within the lib, and adds an event that will be helpful in supporting external audio clients such as Lavalink.

### DIFF
--- a/src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
+++ b/src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
@@ -1,4 +1,3 @@
-using Discord.WebSocket.Entities.Guilds;
 using System;
 using System.Threading.Tasks;
 
--- a/src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
+++ b/src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
@@ -165,7 +165,7 @@ namespace Discord.WebSocket
             remove { _userVoiceStateUpdatedEvent.Remove(value); }
         }
         internal readonly AsyncEvent<Func<SocketUser, SocketVoiceState, SocketVoiceState, Task>> _userVoiceStateUpdatedEvent = new AsyncEvent<Func<SocketUser, SocketVoiceState, SocketVoiceState, Task>>();
-        /// <summary> Fired when the bot connects/disconnects to a Discord voice server. </summary>
+        /// <summary> Fired when the bot connects to a Discord voice server. </summary>
         public event Func<SocketVoiceServer, Task> VoiceServerUpdated
         {
             add { _voiceServerUpdatedEvent.Add(value);  }
--- a/src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
+++ b/src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
@@ -1,4 +1,5 @@
-﻿using System;
+using Discord.WebSocket.Entities.Guilds;
+using System;
 using System.Threading.Tasks;
 
 namespace Discord.WebSocket
@@ -165,6 +166,13 @@ namespace Discord.WebSocket
             remove { _userVoiceStateUpdatedEvent.Remove(value); }
         }
         internal readonly AsyncEvent<Func<SocketUser, SocketVoiceState, SocketVoiceState, Task>> _userVoiceStateUpdatedEvent = new AsyncEvent<Func<SocketUser, SocketVoiceState, SocketVoiceState, Task>>();
+        /// <summary> Fired when the bot connects/disconnects to a Discord voice server. </summary>
+        public event Func<SocketVoiceServer, Task> VoiceServerUpdated
+        {
+            add { _voiceServerUpdatedEvent.Add(value);  }
+            remove { _voiceServerUpdatedEvent.Remove(value); }
+        }
+        internal readonly AsyncEvent<Func<SocketVoiceServer, Task>> _voiceServerUpdatedEvent = new AsyncEvent<Func<SocketVoiceServer, Task>>();
         /// <summary> Fired when the connected account is updated. </summary>
         public event Func<SocketSelfUser, SocketSelfUser, Task> CurrentUserUpdated {
             add { _selfUpdatedEvent.Add(value); }
--- a/src/Discord.Net.WebSocket/DiscordSocketClient.cs
+++ b/src/Discord.Net.WebSocket/DiscordSocketClient.cs
@@ -5,6 +5,7 @@ using Discord.Net.Converters;
 using Discord.Net.Udp;
 using Discord.Net.WebSockets;
 using Discord.Rest;
+using Discord.WebSocket.Entities.Guilds;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using System;
@@ -1474,6 +1475,9 @@ namespace Discord.WebSocket
                                         await UnknownGuildAsync(type, data.GuildId).ConfigureAwait(false);
                                         return;
                                     }
+
+                                    SocketVoiceServer VoiceServer = new SocketVoiceServer(data.GuildId, data.Endpoint, data.Token);
+                                    await TimedInvokeAsync(_voiceServerUpdatedEvent, nameof(UserVoiceStateUpdated), VoiceServer).ConfigureAwait(false);
                                 }
                                 break;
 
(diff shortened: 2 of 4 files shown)


src/Discord.Net.WebSocket/Entities/Guilds/SocketVoiceServer.cs
```diff
@@ -0,0 +1,19 @@
+using System.Diagnostics;
+
+namespace Discord.WebSocket.Entities.Guilds
```
foxbot: This should be namespaced just to `Discord.WebSocket`

src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -1474,6 +1475,9 @@ public override RestVoiceRegion GetVoiceRegion(string id)
                                         await UnknownGuildAsync(type, data.GuildId).ConfigureAwait(false);
                                         return;
                                     }
+
+                                    SocketVoiceServer VoiceServer = new SocketVoiceServer(data.GuildId, data.Endpoint, data.Token);
```
foxbot: We prefer to use `var`, and local variables should be camel-cased

src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -1474,6 +1475,9 @@ public override RestVoiceRegion GetVoiceRegion(string id)
...
+                                    await TimedInvokeAsync(_voiceServerUpdatedEvent, nameof(UserVoiceStateUpdated), VoiceServer).ConfigureAwait(false);
```
foxbot: I reckon this should probably be moved up a few lines, so that Lavalink implementations can properly intercept the event and throw an exception to stop the event handler before the library opens its own voice connection.

src/Discord.Net.WebSocket/Entities/Guilds/SocketVoiceServer.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    [DebuggerDisplay(@"{DebuggerDisplay,nq}")]
+    public class SocketVoiceServer
+    {
+        public ulong GuildId { get; private set; }
```
foxbot: This could probably be a `Cacheable<ulong, IGuild>` instead.

foxbot (changes requested):

src/Discord.Net.WebSocket/Entities/Guilds/SocketVoiceServer.cs
devoxin: Hmm, not really sure how I'd go about using the Cacheable class...

src/Discord.Net.WebSocket/SocketVoiceServer.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Discord.WebSocket
+{
+    [DebuggerDisplay(@"{DebuggerDisplay,nq}")]
+    public class SocketVoiceServer
+    {
+        public ulong GuildId { get; private set; }
+        public string Endpoint { get; private set; }
+        public string Token { get; private set; }
+
+        internal SocketVoiceServer(ulong GuildId, string Endpoint, string Token)
```
AntiTcb: Parameters should be camelCased, and `this` should be avoided per style consistency with the lib.

src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
```diff
@@ -165,6 +165,13 @@ public partial class BaseSocketClient
             remove { _userVoiceStateUpdatedEvent.Remove(value); }
         }
         internal readonly AsyncEvent<Func<SocketUser, SocketVoiceState, SocketVoiceState, Task>> _userVoiceStateUpdatedEvent = new AsyncEvent<Func<SocketUser, SocketVoiceState, SocketVoiceState, Task>>();
+        /// <summary> Fired when the bot connects/disconnects to a Discord voice server. </summary>
```
AntiTcb: According to [Discord's docs](https://discordapp.com/developers/docs/topics/gateway#voice-server-update), this event fires only on connecting, not disconnecting, as well as when the current voice instances changes servers.

src/Discord.Net.WebSocket/SocketVoiceServer.cs
```diff
@@ -0,0 +1,19 @@
+using System.Diagnostics;
+
+namespace Discord.WebSocket
+{
+    [DebuggerDisplay(@"{DebuggerDisplay,nq}")]
```
AntiTcb, line 5: You need to implement a function named `DebuggerDisplay` for attribute to do its job. See [SocketRole](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.WebSocket/Entities/Roles/SocketRole.cs#L58) for an example.

src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -5,6 +5,7 @@
 using Discord.Net.Udp;
 using Discord.Net.WebSockets;
 using Discord.Rest;
+using Discord.WebSocket;
```
AntiTcb: This namespace shouldn't need to be imported given that the `DiscordSocketClient` class is already defined within it.

AntiTcb (changes requested):

src/Discord.Net.WebSocket/Entities/Guilds/SocketVoiceServer.cs
AntiTcb: There are quite a few events in the `DiscordSocketClass` that use `Cacheable<TId,TEntity>`, take a look at them for a guideline. Basically gist is you're going to have a `bool` for if the entity is cached, check the cache for the item, and construct a new `Cacheable<,>` with the guild id, the cached guild, the isCached boolean, and a `Func<>` to retrieve the object via REST if it is not cached.

devoxin:
Anything else need changing?

src/Discord.Net.WebSocket/Entities/Guilds/SocketVoiceServer.cs
devoxin: Alright, done

src/Discord.Net.WebSocket/SocketVoiceServer.cs
```diff
@@ -5,17 +5,17 @@ namespace Discord.WebSocket
...
+        internal SocketVoiceServer(Cacheable<IGuild, ulong> guild, ulong guildId, string endpoint, string token)
         {
-            GuildId = guildId;
+            Guild = guild;
             Endpoint = endpoint;
             Token = token;
         }
 
-        private string DebuggerDisplay => $"SocketVoiceServer ({GuildId})";
+        private string DebuggerDisplay => $"SocketVoiceServer ({Guild.Id})";
```
AntiTcb: This will null ref if the guild isn't cached (literally should never happen, but worth pointing out. Opt for using the ID directly from the cacheable.

AntiTcb (changes requested):

src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -1464,6 +1464,12 @@ public override RestVoiceRegion GetVoiceRegion(string id)
 
                                     var data = (payload as JToken).ToObject<VoiceServerUpdateEvent>(_serializer);
                                     var guild = State.GetGuild(data.GuildId);
+                                    var cacheable = new Cacheable<IGuild, ulong>(guild, data.GuildId, guild != null,
+                                        async () => (IGuild) await ApiClient.GetGuildAsync(data.GuildId));
```
AntiTcb: Was the cast necessary? Not sure if the compiler would actually complain about that or not, but my first instinct just from reading is that it shouldn't.

src/Discord.Net.WebSocket/SocketVoiceServer.cs
devoxin: I thought .Id did return the ID (ulong)?

src/Discord.Net.WebSocket/DiscordSocketClient.cs
devoxin: It did https://i.serux.pro/6c425faf71.png

src/Discord.Net.WebSocket/SocketVoiceServer.cs
AntiTcb: Ah, my bad, misread.

AntiTcb (approved):

devoxin:
@foxbot?

Giggitybyte:
fox pls

PR #986, opened 2018-03-19
Expose the internal entity type readers

Joe4evr:
After some discussion in #941, I thought that this might be a favorable change.

I decided not to expose the Primitive, Enum, and Nullable readers since there seems to be a bit more complexity involved with those and it seems just too unlikely someone wants to piggyback off of the existing behavior outside of the entities.

### DIFF
--- a/src/Discord.Net.Commands/Readers/ChannelTypeReader.cs
+++ b/src/Discord.Net.Commands/Readers/ChannelTypeReader.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
@@ -6,7 +6,7 @@ using System.Threading.Tasks;
 
 namespace Discord.Commands
 {
-    internal class ChannelTypeReader<T> : TypeReader
+    public class ChannelTypeReader<T> : TypeReader
         where T : class, IChannel
     {
         public override async Task<TypeReaderResult> ReadAsync(ICommandContext context, string input, IServiceProvider services)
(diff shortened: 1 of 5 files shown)


Joe4evr:
Open question: Since after this change, it'll be likely that people inherit one of these readers and call the base implementation to see if that succeeds:
```cs
var baseResult = await base.ReadAsync(context, input, services);
if (baseResult.IsSuccess)
{
    //.....
}
return baseResult;
```
would it be prudent to add a new property to `TypeReaderResult` that returns the best match provided the result was successful?
```cs
public object BestMatch => IsSuccess
    ? (Values.Count == 1 ? Values.Single().Value : Values.OrderByDescending(v => v.Score).First().Value)
    : throw new InvalidOperationException("TypeReaderResult was not successful.");
```

Nensec:
Would't it be better that BestMatch simply returned `null` rather than throw an exception? Doing the `IsSuccess` check seems more like something the caller should do, plus it prevents doing a simple `??` fallback. Now it requires an entire `try{} catch{}`

Joe4evr:
Not really.
1. The caller *should* check `result.IsSuccess` before accessing `BestMatch` (this is the same principle as when handling arbitrary `Stream`s, where you should check `CanRead`/`CanWrite`/`CanSeek` etc. before performing the respective operation).
2. There could be the case where [`null` was actually the value given for that parameter](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Commands/Readers/NullableTypeReader.cs#L29) and the caller should respect that.

PR #992, opened 2018-03-23
Throw when attempting to modify a message not made by the current user

Joe4evr:
Too many times people forget that they cannot change the contents of another user's message, and we are not communicating that fact properly.

### DIFF
--- a/src/Discord.Net.Rest/Entities/Messages/MessageHelper.cs
+++ b/src/Discord.Net.Rest/Entities/Messages/MessageHelper.cs
@@ -1,4 +1,4 @@
-﻿using Discord.API.Rest;
+using Discord.API.Rest;
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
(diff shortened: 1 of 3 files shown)


src/Discord.Net.Rest/Entities/Messages/MessageHelper.cs
```diff
@@ -13,6 +13,9 @@ internal static class MessageHelper
         public static async Task<Model> ModifyAsync(IMessage msg, BaseDiscordClient client, Action<MessageProperties> func,
             RequestOptions options)
         {
+            if (msg.Author.Id != client.CurrentUser.Id)
+                throw new InvalidOperationException("Discord allows only the author of a message to change it.");
```
foxbot: I think this would look better if it dropped the 'Discord' part and instead read `Only the author of a message may change it`. Passive-aggressive jabs at Discord look better in commit messages than public-facing errors :^)

foxbot (approved):

PR #994, opened 2018-03-24
Allows command arguments to consist of only 'escaped' characters

Quahu:
Say, we have a command that looks like this:
```cs
[Command("test")]
public Task TestAsync(string parameter)
    => ReplyAsync(parameter);
```
Our prefix will be an exclamation mark.
Invoking it as `!test :x:` will work perfectly fine, but invoking it as `!test \:x:` will fail, returning a ParseResult with the error message: `The input text has too few parameters.`.
Also, invoking commands this way with `params T[] parameters` will invoke the actual command, but with an empty array.

I tested it after my changes, and it seemed to work as expected.

~~I hope I don't break the lib with this.~~

### DIFF
--- a/src/Discord.Net.Commands/CommandParser.cs
+++ b/src/Discord.Net.Commands/CommandParser.cs
@@ -33,6 +33,13 @@ namespace Discord.Commands
                 else
                     c = '\0';
 
+                //If we're processing a remainder parameter, ignore all other logic
+                if (curParam != null && curParam.IsRemainder && curPos != endPos)
+                {
+                    argBuilder.Append(c);
+                    continue;
+                }
+
                 //If we're not currently processing one, are we starting the next argument yet?
                 if (curPart == ParserPart.None)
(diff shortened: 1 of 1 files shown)


Joe4evr:
How is this different from using `[Remainder]`?

AntiTcb:
Well, you could have parameters after it for one thing.

Quahu:
@Joe4evr RemainderAttribute wouldn't prevent it from breaking. It would still return the ParseResult in the case I mentioned.

FiniteReality (commented):
Is there any reason why we couldn't move the remainder param logic above the escape logic rather than move the escape logic to after handling the default state for the parser?

Quahu:
Erm, if I understood your question correctly - it wouldn't fix the characters being escaped. It's not an issue with remainder parameters, but normal parameters.

foxbot (approved):
lgtm

PR #996, opened 2018-03-24
Implement IMessageChannel#DeleteMessageAsync

Still34:
Not sure if this was left out during development for a reason, but I don't see why `IMessageChannel#DeleteMessageAsync` wasn't included. There are users out there who might have a message ID in a channel that they wish to delete, but do not want to retrieve the message again or simply do not have the access (`READ_MESSAGE_HISTORY` in a guild) to.


src/Discord.Net.Rest/Entities/Channels/ChannelHelper.cs
```diff
@@ -184,6 +184,12 @@ internal static class ChannelHelper
             var model = await client.ApiClient.UploadFileAsync(channel.Id, args, options).ConfigureAwait(false);
             return RestUserMessage.Create(client, channel, client.CurrentUser, model);
         }
+        
+        public static async Task DeleteMessageAsync(IMessageChannel channel, BaseDiscordClient client,
```
foxbot: I figure this can either reference MessageHelper, or have MessageHelper reference this - would like to avoid code duplication where possible.

PR #998, opened 2018-03-25
Allow setting IgnoreExtraArgs on an individual basis

Joe4evr:
Someone in DAPI requested the ability to override the config setting on an individual command basis.

Also removes some stray comments and unused usings.


Joe4evr:
I don't like that approach because it'd make it impossible to set the value on a Module-wide basis.

I'd rather push for the `RunMode` setting to be promoted to its own attribute like this instead.

pengowray:
Was this really never merged?

Still34:
Pretty sure it is merged?

pengowray:
Apologies. You are correct.

For easy/future reference:

This works:
```C#
[Command("hello", ignoreExtraArgs: true)]
```

This **does not** work:
```C#
[Command("hello", IgnoreExtraArgs = true)]
```

PR #1003, opened 2018-03-27
Fix ChannelPermissions Modify parameter to be correct default value

Chris-Johnston:
The types in the `Modify` methods follow a pattern of using nullable bool values with a default value of null. It looks like one of the parameters were copied from the constructor and the type and default value was not changed.

 I added tests that address when this becomes a problem, including this specific case but also general cases for most permissions. The tests should fail using the old version of ChannelPermissions.cs

**Actual Behavior**
```cs
var a = new ChannelPermissions(useExternalEmojis: true);
var b = a.Modify(); // no params
// b.UseExternalEmojis is false
```
**Expected Behavior**
```cs
var a = new ChannelPermissions(useExternalEmojis: true);
var b = a.Modify(); // no params
// b.UseExternalEmojis is true
```

~~~I don't see this inconsistency in `GuildPermissions` or `OverwritePermissions`.~~~

Still found one in the `GuildPermissions` cstr, some parameters are nullable, some aren't. This was added in 27b1a5be6f7b37d844e55ed13d65c9a8d950c9b7 .


Chris-Johnston:
9207782b4f8ea15a2f2ce76d12647e0bbd06a487 makes some minor syntax and code formatting changes just for readability. I can remove it if it doesn't match the style guidelines.

(also, not sure what that temp file was but it was removed in the next commit)

PR #1004, opened 2018-03-27
Fix #995, Move Category Implementation from IGuildChannel to INestedChannel

### LINKED ISSUE #995: Category ID should not exist under ICategoryChannel
This causes `ICategoryChannel` to have a `CategoryId` which is invalid for a category, since a category cannot have a parent category on-top. Thus, `CategoryId` should either not be an implementation detail under `IGuildChannel`, or cause `ICategoryChannel#CategoryId` to throw an `NotSupportedException` instead.

Chris-Johnston:
This change adds the `INestedChannel` interface and extends `ITextChannel` and `IVoiceChannel` with this interface. All category-related stuff like CategoryID and GetCategoryAsync are moved from `IGuildChannel` to `INestedChannel`. `INestedChannel`s are only channels that can be nested within a category in a guild.

This is a breaking change, not all `IGuildChannel`s will have the category property or methods anymore. The types have also changed slightly.

~~~This change makes `ICategoryChannel#CategoryID` throw a `NotSupportedException`. This also includes `RestGuildChannel#GetCategoryAsync` and `SocketGuildChannel#GetCategoryAsync`, since they call `CategoryId#get`.~~~

~~~This could be done in a more OOP way, but may end up with a more significant change to the code base, I would make a `INestedChannel` interface that provides `CategoryID`, and remove that from the `IGuildChannel` interface. (Actually now that I think about it, it doesn't sound too bad. If this is preferred I could try this instead.)~~~

~~~Edit: @Still34 made a great point for keeping the exceptions, `parent_id` is defined for all channel types at the api level, and this would prevent future issues should discord allow nested channel categories. Also there's other stuff that throws during runtime already.~~~

~~~Expected output is outlined in the tests, when `CategoryId` is accessed or when `GetCategoryAsync` is called, both should throw `NotSupportedException`.~~

This PR also adds minimal tests for channel categories. This should be more complete, but could be addressed in a different PR.

### DIFF
--- a/src/Discord.Net.Core/Entities/Channels/ITextChannel.cs
+++ b/src/Discord.Net.Core/Entities/Channels/ITextChannel.cs
@@ -5,7 +5,7 @@ using System.Threading.Tasks;
 
 namespace Discord
 {
-    public interface ITextChannel : IMessageChannel, IMentionable, IGuildChannel, INestedChannel
+    public interface ITextChannel : IMessageChannel, IMentionable, INestedChannel
     {
         /// <summary> Checks if the channel is NSFW. </summary>
         bool IsNsfw { get; }
--- a/src/Discord.Net.Core/Entities/Channels/ITextChannel.cs
+++ b/src/Discord.Net.Core/Entities/Channels/ITextChannel.cs
@@ -1,11 +1,11 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Threading.Tasks;
 
 namespace Discord
 {
-    public interface ITextChannel : IMessageChannel, IMentionable, IGuildChannel
+    public interface ITextChannel : IMessageChannel, IMentionable, IGuildChannel, INestedChannel
     {
         /// <summary> Checks if the channel is NSFW. </summary>
         bool IsNsfw { get; }
--- a/src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs
+++ b/src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs
@@ -9,9 +9,10 @@ using Model = Discord.API.Channel;
 namespace Discord.Rest
 {
     [DebuggerDisplay(@"{DebuggerDisplay,nq}")]
-    public class RestTextChannel : RestGuildChannel, IRestMessageChannel, ITextChannel
+    public class RestTextChannel : RestGuildChannel, IRestMessageChannel, ITextChannel, INestedChannel
     {
         public string Topic { get; private set; }
+        public ulong? CategoryId { get; private set; }
 
         public string Mention => MentionUtils.MentionChannel(Id);
 
@@ -167,6 +168,14 @@ namespace Discord.Rest
                 return GetUsersAsync(options);
             else
                 return AsyncEnumerable.Empty<IReadOnlyCollection<IGuildUser>>();
+        }
+
+        // INestedChannel
+        async Task<ICategoryChannel> INestedChannel.GetCategoryAsync()
+        {
+            if (CategoryId.HasValue)
+                return (await Guild.GetChannelAsync(CategoryId.Value).ConfigureAwait(false)) as ICategoryChannel;
+            return null;
         }
     }
 }
--- a/src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs
+++ b/src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs
@@ -175,10 +175,10 @@ namespace Discord.Rest
         }
 
         // INestedChannel
-        async Task<ICategoryChannel> INestedChannel.GetCategoryAsync()
+        async Task<ICategoryChannel> INestedChannel.GetCategoryAsync(CacheMode mode, RequestOptions options)
         {
             if (CategoryId.HasValue)
-                return (await Guild.GetChannelAsync(CategoryId.Value).ConfigureAwait(false)) as ICategoryChannel;
+                return (await Guild.GetChannelAsync(CategoryId.Value, mode, options).ConfigureAwait(false)) as ICategoryChannel;
             return null;
         }
     }
--- a/src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs
+++ b/src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs
@@ -32,7 +32,7 @@ namespace Discord.Rest
         internal override void Update(Model model)
         {
             base.Update(model);
-
+            CategoryId = model.CategoryId;
             Topic = model.Topic.Value;
             _nsfw = model.Nsfw.GetValueOrDefault();
         }
@@ -47,7 +47,7 @@ namespace Discord.Rest
             => ChannelHelper.GetUserAsync(this, Guild, Discord, id, options);
         public IAsyncEnumerable<IReadOnlyCollection<RestGuildUser>> GetUsersAsync(RequestOptions options = null)
             => ChannelHelper.GetUsersAsync(this, Guild, Discord, null, null, options);
-
+        
         public Task<RestMessage> GetMessageAsync(ulong id, RequestOptions options = null)
             => ChannelHelper.GetMessageAsync(this, Discord, id, options);
         public IAsyncEnumerable<IReadOnlyCollection<RestMessage>> GetMessagesAsync(int limit = DiscordConfig.MaxMessagesPerBatch, RequestOptions options = null)
@@ -84,6 +84,9 @@ namespace Discord.Rest
             => ChannelHelper.GetWebhookAsync(this, Discord, id, options);
         public Task<IReadOnlyCollection<RestWebhook>> GetWebhooksAsync(RequestOptions options = null)
             => ChannelHelper.GetWebhooksAsync(this, Discord, options);
+    
+        public Task<ICategoryChannel> GetCategoryAsync(RequestOptions options = null)
+            => ChannelHelper.GetCategoryAsync(this, Discord, options);
 
         private string DebuggerDisplay => $"{Name} ({Id}, Text)";
 
@@ -110,6 +113,7 @@ namespace Discord.Rest
             else
                 return AsyncEnumerable.Empty<IReadOnlyCollection<IMessage>>();
         }
+        
         IAsyncEnumerable<IReadOnlyCollection<IMessage>> IMessageChannel.GetMessagesAsync(ulong fromMessageId, Direction dir, int limit, CacheMode mode, RequestOptions options)
         {
             if (mode == CacheMode.AllowDownload)
--- a/src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs
+++ b/src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs
@@ -92,11 +92,11 @@ namespace Discord.Rest
 
         //ITextChannel
         async Task<IWebhook> ITextChannel.CreateWebhookAsync(string name, Stream avatar, RequestOptions options)
-            => await CreateWebhookAsync(name, avatar, options);
+            => await CreateWebhookAsync(name, avatar, options).ConfigureAwait(false);
         async Task<IWebhook> ITextChannel.GetWebhookAsync(ulong id, RequestOptions options)
-            => await GetWebhookAsync(id, options);
+            => await GetWebhookAsync(id, options).ConfigureAwait(false);
         async Task<IReadOnlyCollection<IWebhook>> ITextChannel.GetWebhooksAsync(RequestOptions options)
-            => await GetWebhooksAsync(options);
+            => await GetWebhooksAsync(options).ConfigureAwait(false);
 
         //IMessageChannel
         async Task<IMessage> IMessageChannel.GetMessageAsync(ulong id, CacheMode mode, RequestOptions options)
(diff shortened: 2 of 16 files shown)


SubZero0:
I don't really like the throw approach. Imo, the `INestedChannel` is a lot prettier.
Right now, it returns null and users forget they need the Id property in the category channel to compare.
This solution would just make them notice it only at runtime with an exception. I would prefer them noticing it before and the interface would perfectly fit.

> and this would prevent future issues should discord allow nested channel categories

Why would it? The solution for the `INestedChannel` would be basically make category inherit it and now we would need to replace the throw.

> `parent_id` is defined for all channel types at the api level

The lib is at a higher level than the api, so imo we don't need to follow exactly how it is (and there's a lot more properties defined at the api level).

Chris-Johnston:
I'm torn on which approach to use and could go either way. Maybe the `INestedChannel` approach would be more appropriate for a larger-scale fix, and this could be a temporary solution?

Would appreciate the input of others on this as well.

Still34:
I lean towards the nested channel interface approach as well. Throwing upon runtime is just not a good idea overall, even if you try to warn the user of the exception.

Chris-Johnston:
The interface approach is now implemented. I didn't realize at first that the CI doesn't run tests on PRs, and I'm having some issues running the tests on my machine that I will investigate later.

As a temporary solution, I've made [some sample code to validate correct behavior for a `DiscordSocketClient`.](https://gist.github.com/Chris-Johnston/0dc231b1a523d04865a838764e0b5e53) It appears to be working correctly, but actual tests would be more ideal.

I was somewhat unsure about implementing `INestedChannel.GetCategoryAsync` as an explicit interface definition, and then doing it again with `ChannelHelper.GetCategoryAsync`. Was this the correct pattern to follow? I tried to base this off of the other code as a guide.

Chris-Johnston:
I was able to fix some of the issues I encountered when testing by deleting the `CachedRestClient` cache, since I had to create new channels (the categories themselves), and the cache wasn't updating. This doesn't affect CI, but might be a relevant detail to add to the `Contributing` file explaining the testing procedure.

I also rewrote the category tests to match the pattern established by the existing channel tests.

I believe the implementation is working now, feedback would be appreciated.

src/Discord.Net.Core/Entities/Channels/ITextChannel.cs
```diff
@@ -1,11 +1,11 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Threading.Tasks;
 
 namespace Discord
 {
-    public interface ITextChannel : IMessageChannel, IMentionable, IGuildChannel
+    public interface ITextChannel : IMessageChannel, IMentionable, IGuildChannel, INestedChannel
```
SubZero0: How should we deal with redundant interfaces? `INestedChannel` inherits `IGuildChannel` already.
1) Keep both
2) Remove only `IGuildChannel` from here (and other redundant places) and keep `INestedChannel` (I prefer this one)
3) Make `INestedChannel` not inherit `IGuildChannel` (makes hard accessing specific properties when casting, dont recommend)

src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs
```diff
@@ -9,9 +9,10 @@
 namespace Discord.Rest
 {
     [DebuggerDisplay(@"{DebuggerDisplay,nq}")]
-    public class RestTextChannel : RestGuildChannel, IRestMessageChannel, ITextChannel
+    public class RestTextChannel : RestGuildChannel, IRestMessageChannel, ITextChannel, INestedChannel
```
SubZero0: Redundant inheritance since `ITextChannel` already has it.

src/Discord.Net.Rest/Entities/Channels/RestVoiceChannel.cs
```diff
@@ -9,10 +9,11 @@
 namespace Discord.Rest
 {
     [DebuggerDisplay(@"{DebuggerDisplay,nq}")]
-    public class RestVoiceChannel : RestGuildChannel, IVoiceChannel, IRestAudioChannel
+    public class RestVoiceChannel : RestGuildChannel, IVoiceChannel, IRestAudioChannel, INestedChannel
```
SubZero0: Redundant inheritance since `IVoiceChannel` should has it.
(Forgot to make `IVoiceChannel` inherit `INestedChannel`)

src/Discord.Net.WebSocket/Entities/Channels/SocketCategoryChannel.cs
```diff
@@ -20,7 +20,7 @@ public override IReadOnlyCollection<SocketGuildUser> Users
                ChannelPermission.ViewChannel)).ToImmutableArray();
 
         public IReadOnlyCollection<SocketGuildChannel> Channels
-            => Guild.Channels.Where(x => x.CategoryId == Id).ToImmutableArray();
+            => Guild.Channels.Where(x => x is INestedChannel && (x as INestedChannel).CategoryId == Id).ToImmutableArray();
```
SubZero0: C#7: `x => x is INestedChannel nestedChannel && nestedChannel.CategoryId == Id`

src/Discord.Net.WebSocket/Entities/Channels/SocketTextChannel.cs
```diff
@@ -11,11 +11,14 @@
 namespace Discord.WebSocket
 {
     [DebuggerDisplay(@"{DebuggerDisplay,nq}")]
-    public class SocketTextChannel : SocketGuildChannel, ITextChannel, ISocketMessageChannel
+    public class SocketTextChannel : SocketGuildChannel, ITextChannel, ISocketMessageChannel, INestedChannel
```
SubZero0: Redundant inheritance since ITextChannel already has it.

src/Discord.Net.WebSocket/Entities/Channels/SocketVoiceChannel.cs
```diff
@@ -11,10 +11,13 @@
 namespace Discord.WebSocket
 {
     [DebuggerDisplay(@"{DebuggerDisplay,nq}")]
-    public class SocketVoiceChannel : SocketGuildChannel, IVoiceChannel, ISocketAudioChannel
+    public class SocketVoiceChannel : SocketGuildChannel, IVoiceChannel, ISocketAudioChannel, INestedChannel
```
SubZero0: Redundant inheritance since `IVoiceChannel` should has it.
(Forgot to make `IVoiceChannel` inherit `INestedChannel`)

Chris-Johnston:
Cleaned up some of the use of interfaces w/ feedback from SubZero0.

We were stuck on a design decision and were unsure about the best option was. Should both `ITextChannel` and `INestedChannel` implement `IGuildChannel`? Or should `ITextChannel` and `IVoiceChannel` be changed to implement only `INestedChannel` instead of `IGuildChannel` and `INestedChannel` ?

We also considered `INestedChannel` not implement anything, but I don't think that makes sense, since nested channels must be guild channels.

Thoughts?

Chris-Johnston:
Also, depending on if/when this is merged relative to #1016 , `SyncWithCategory` would go under `INestedChannel`

src/Discord.Net.Core/Entities/Channels/ITextChannel.cs
Chris-Johnston:
~~~we could make `ITextChannel` inherit `INestedChannel` (which inherits `IGuildChannel`)~~~

this is option # 2

src/Discord.Net.WebSocket/Entities/Channels/SocketTextChannel.cs
```diff
@@ -121,6 +124,10 @@ public Task<RestWebhook> GetWebhookAsync(ulong id, RequestOptions options = null
         public Task<IReadOnlyCollection<RestWebhook>> GetWebhooksAsync(RequestOptions options = null)
             => ChannelHelper.GetWebhooksAsync(this, Discord, options);
 
+        // Categories
+        public Task<ICategoryChannel> GetCategoryAsync(RequestOptions options = null)
```
SubZero0: Remove this and leave only `INestedChannel.GetCategoryAsync()` since theres already the `Category` property to retrieve it.

Same pattern as how we deal to retrieve users.

src/Discord.Net.WebSocket/Entities/Channels/SocketVoiceChannel.cs
```diff
@@ -52,7 +55,10 @@ public override SocketGuildUser GetUser(ulong id)
                 return user;
             return null;
         }
-        
+
+        public Task<ICategoryChannel> GetCategoryAsync(RequestOptions options = null)
```
SubZero0: Remove this and leave only INestedChannel.GetCategoryAsync() since theres already the Category property to retrieve it.

Same pattern as how we deal to retrieve users.

src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs
```diff
@@ -168,5 +173,13 @@ IAsyncEnumerable<IReadOnlyCollection<IUser>> IChannel.GetUsersAsync(CacheMode mo
             else
                 return AsyncEnumerable.Empty<IReadOnlyCollection<IGuildUser>>();
         }
+
+        // INestedChannel
+        async Task<ICategoryChannel> INestedChannel.GetCategoryAsync()
```
SubZero0: I think we need to change this signature (add CacheMode and RequestOptions) to follow the same pattern as other rest entities and call the inner `GetCategoryAsync`.

See:
https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Rest/Entities/Guilds/RestGuild.cs#L295
https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs#L141

src/Discord.Net.Rest/Entities/Channels/RestVoiceChannel.cs
```diff
@@ -48,5 +52,13 @@ Task<IGuildUser> IGuildChannel.GetUserAsync(ulong id, CacheMode mode, RequestOpt
             => Task.FromResult<IGuildUser>(null);
         IAsyncEnumerable<IReadOnlyCollection<IGuildUser>> IGuildChannel.GetUsersAsync(CacheMode mode, RequestOptions options)
             => AsyncEnumerable.Empty<IReadOnlyCollection<IGuildUser>>();
+
+        // INestedChannel
+        async Task<ICategoryChannel> INestedChannel.GetCategoryAsync()
```
SubZero0: Same thing as above.

SubZero0 (commented):
Could you add `.ConfigureAwait(false)` to lines 95, 97, and 99 (webhook related ones) while at it too?

src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs
```diff
@@ -168,5 +173,13 @@ IAsyncEnumerable<IReadOnlyCollection<IUser>> IChannel.GetUsersAsync(CacheMode mo
             else
                 return AsyncEnumerable.Empty<IReadOnlyCollection<IGuildUser>>();
         }
+
+        // INestedChannel
+        async Task<ICategoryChannel> INestedChannel.GetCategoryAsync(CacheMode mode, RequestOptions options)
+        {
```
SubZero0, line 179: I believe this would be a better way (to not rely on the Guild on Rest):
```cs
if (mode == CacheMode.AllowDownload)
    return await GetCategoryAsync(options).ConfigureAwait(false);
else
    return null;
```

This is what I meant with "follow the same pattern as other rest entities and call the inner `GetCategoryAsync`" in my last review.
See:https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs#L141

src/Discord.Net.Rest/Entities/Channels/RestVoiceChannel.cs
```diff
@@ -48,5 +52,13 @@ Task<IGuildUser> IGuildChannel.GetUserAsync(ulong id, CacheMode mode, RequestOpt
             => Task.FromResult<IGuildUser>(null);
         IAsyncEnumerable<IReadOnlyCollection<IGuildUser>> IGuildChannel.GetUsersAsync(CacheMode mode, RequestOptions options)
             => AsyncEnumerable.Empty<IReadOnlyCollection<IGuildUser>>();
+
+        // INestedChannel
+        async Task<ICategoryChannel> INestedChannel.GetCategoryAsync(CacheMode mode, RequestOptions options)
+        {
```
SubZero0, line 58: See above.

src/Discord.Net.Rest/Entities/Channels/RestTextChannel.cs
Chris-Johnston: This is implemented, but isn't showing as outdated in Github.

src/Discord.Net.Core/Entities/Channels/ITextChannel.cs
Chris-Johnston: I went with option # 2 for now.

Chris-Johnston:
@Herdo Please see comments on most branch for #1004 , changes have been made since original commit.

PR #1005, opened 2018-03-27
Add a dedicated TimeSpan reader so it doesn't suck

Joe4evr:
I considered also adding a dedicated reader for `DateTime{Offset}`, but because timezones and cultural differences in notation, I decided against it.

### DIFF
--- a/src/Discord.Net.Commands/Readers/TimeSpanTypeReader.cs
+++ b/src/Discord.Net.Commands/Readers/TimeSpanTypeReader.cs
@@ -27,7 +27,7 @@ namespace Discord.Commands
 
         public override Task<TypeReaderResult> ReadAsync(ICommandContext context, string input, IServiceProvider services)
         {
-            return (TimeSpan.TryParseExact(input, _formats, CultureInfo.InvariantCulture, out var timeSpan))
+            return (TimeSpan.TryParseExact(input.ToLowerInvariant(), _formats, CultureInfo.InvariantCulture, out var timeSpan))
                 ? Task.FromResult(TypeReaderResult.FromSuccess(timeSpan))
                 : Task.FromResult(TypeReaderResult.FromError(CommandError.ParseFailed, "Failed to parse TimeSpan"));
         }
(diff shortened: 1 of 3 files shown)


src/Discord.Net.Commands/Readers/TimeSpanTypeReader.cs
```diff
@@ -0,0 +1,35 @@
...
+            "%h'h'%s's'",           //  3h  1s
+            "%h'h'",                //  3h
+            "%m'm'%s's'",           //    2m1s
+            "%m'm'",                //    2m
+            "%s's'",                //      1s
+        };
+
+        public override Task<TypeReaderResult> ReadAsync(ICommandContext context, string input, IServiceProvider services)
+        {
+            return (TimeSpan.TryParseExact(input, _formats, CultureInfo.InvariantCulture, out var timeSpan))
```
Chris-Johnston: What about `input.ToLower()` so input is case insensitive?

khionu: While I would admit, as a developer, weeks seems a bit excessive, I see it used in various bots and actually by the consumers of the bots. I would consider it worth adding.

Joe4evr: @Chris-Johnston Will do. I *was* hoping to cache the `InvariantCulture` so the system doesn't have to excessively load it (especially because the backing field in NetFX is `volitile`), but NS1.1 doesn't support `.ToLower(CultureInfo)`. Oh well....

@khionu Sadly, [there is no format specifier for weeks](https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-timespan-format-strings), so this is as good as it's gonna get.

khionu: Ahh... oh well. Could always manually parse, but that I think that might be considered a bit much for this lib

Joe4evr:
Interesting: It appears the JIT will inline `.ToLowerInvariant()` to call `.ToLower(CultureInfo)` directly - I *guess* that also means it *could* hang on to the `InvariantInfo` instance as if it was a local rather than calling the getter method twice. I can't read ASM to say for sure.

The more you know.

PR #1012, opened 2018-03-30
Fix Embed.Length behavior

### LINKED ISSUE #1011: Embed.Length Evaluates to 0 If Any Operand Is Null
If any of the `Embed`'s properties which are operands in the calculation are `null`, the whole [expression](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Core/Entities/Messages/Embed.cs#L60) that is the left operand of the null-coalescing operator will evaluate to `null`. This causes the operator to return its right operand `0`. Note that the null-coalescing operator has lower precedence than the addition operator. 

Related [documentation](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/nullable-types/using-nullable-types#operators):

> The predefined unary and binary operators and any user-defined operators that exist for value types may also be used by nullable types. **These operators produce a null value if the operands are null;** otherwise, the operator uses the contained value to calculate the result.

Probably among other things, I've noticed this has the consequence of `EmbedBuilder.Build()` not throwing `InvalidOperationException`.

Is this working as intended? Seems like an oversight to me.

Still34:
This should resolve #1011. 

## Summary
- The `get` method is now split up into various lines to better debug any value that may be `null`. 
- Each property now falls back to `0` in case any of them is null.

### DIFF
--- a/src/Discord.Net.Core/Entities/Messages/Embed.cs
+++ b/src/Discord.Net.Core/Entities/Messages/Embed.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


src/Discord.Net.Core/Entities/Messages/Embed.cs
```diff
@@ -57,7 +57,18 @@ internal Embed(EmbedType type)
...
-        public int Length => Title?.Length + Author?.Name?.Length + Description?.Length + Footer?.Text?.Length + Fields.Sum(f => f.Name.Length + f.Value.ToString().Length) ?? 0;
+        public int Length
+        {
+            get
+            {
+                int titleLength = Title?.Length ?? 0;
+                int authorLength = Author?.Name?.Length ?? 0;
+                int descriptionLength = Description?.Length ?? 0;
+                int footerLength = Footer?.Text?.Length ?? 0;
+                int fieldSum = Fields.Sum(f => f.Name?.Length + f.Value?.ToString().Length) ?? 0;
```
AntiTcb, line 68: Should be safe to drop the null check for field names and values.

AntiTcb (changes requested):

Still34:
Resolved above.
EDIT: Nope, read wrong.

PR #1016, opened 2018-04-04
Add the ability to sync guild channel permission to category

Still34:
## Summary

This PR will add a new bool under `GuildChannelProperties` called `SyncWithCategory`, allowing the permission of a guild channel to be synced up with its parent/category.

### DIFF
--- a/src/Discord.Net.Core/Entities/Channels/GuildChannelProperties.cs
+++ b/src/Discord.Net.Core/Entities/Channels/GuildChannelProperties.cs
@@ -1,4 +1,4 @@
-﻿namespace Discord
+namespace Discord
 {
     /// <summary>
     /// Modify an IGuildChannel with the specified changes.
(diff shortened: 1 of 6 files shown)


src/Discord.Net.Core/Entities/Channels/GuildChannelProperties.cs
```diff
@@ -30,5 +30,9 @@ public class GuildChannelProperties
         /// Sets the category for this channel
         /// </summary>
         public Optional<ulong?> CategoryId { get; set; }
+        /// <summary>
+        /// Syncs the permission with the channel's parent (category).
+        /// </summary>
+        public Optional<bool> SyncWithCategory { get; set; }
```
Chris-Johnston: Why an `Optional<bool>` and not a regular `bool` that defaults to `false`?

Issue #1007, opened 2018-03-27
Paginating GetReactionUsersAsync


HelpfulStranger999:
Should a method be added/reworked that paginates the IUserMessage#GetReactionUsersAsync call and returns the complete list of users? As it stands, the bot developer must write his own logic to get all the reaction users or use another library that does so.

Vanilla-Chan:
![pr0text 4](https://user-images.githubusercontent.com/36478190/38333999-754e30da-385a-11e8-8fe9-610ea6a889a7.jpg)
The limit is not working, and i could not get afteruserid to work

foxbot:
This was merged via 79811d0e618c011e6cca542811031d824650dae5

PR #1017, opened 2018-04-04
Fix GetReactionUsersAsync in DiscordRestApiClient

Kiritsu:
This should fix the issue with the `limit` parameter and `afterUserId` parameter that were completely ignored.

Just added the usage of limit and after (https://discordapp.com/developers/docs/resources/channel#get-reactions) in the request


Still34:
Might as well attempt to address #1007 as well.

Issue #1019, opened 2018-04-06
CreateTextChannel/CreateVoiceChannel should have the ability to specify parameters


JaydenMaalouf:
In the Discord Documentation (https://discordapp.com/developers/docs/resources/guild#create-guild-channel) you will see there are parameters in which you can specify the: topic, bitrate, user_limit, permission_overwrites, parent_id, nsfw of a channel during its creation. I'd like to be able to do this within Discord.NET as currently I am having to ModifyAsync after the fact. 

Having the ModifyAsync the parent_id after creation results in a channel being created then moved and it causes all other channels to "bump" slightly.

Thank you for your time!

foxbot:
Merged via bf5275e071b78a25fd4501a930e69c5dc8cd3b19

PR #1020, opened 2018-04-06
Add ability to specify parameters on channel creation

Joe4evr:
Proposing this solution for #1019.

I'm not 100% sure if this'll Just Work™, so please tell me if more needs to be done.

### DIFF
--- a/src/Discord.Net.Core/Entities/Guilds/IGuild.cs
+++ b/src/Discord.Net.Core/Entities/Guilds/IGuild.cs
@@ -1,4 +1,4 @@
-﻿using Discord.Audio;
+using Discord.Audio;
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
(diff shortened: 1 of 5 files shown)


JaydenMaalouf:
I can't wait to try this!

Still34:
I'm not sure if I agree with the `Func` approach. Doesn't seem to fit the style.

Joe4evr:
It's literally the same thing we do for `ModifyAsync()`.

src/Discord.Net.Core/Entities/Guilds/IGuild.cs
```diff
@@ -91,9 +91,9 @@ public interface IGuild : IDeletable, ISnowflakeEntity
         Task<ITextChannel> GetDefaultChannelAsync(CacheMode mode = CacheMode.AllowDownload, RequestOptions options = null);
         Task<IGuildChannel> GetEmbedChannelAsync(CacheMode mode = CacheMode.AllowDownload, RequestOptions options = null);
         /// <summary> Creates a new text channel. </summary>
-        Task<ITextChannel> CreateTextChannelAsync(string name, RequestOptions options = null);
+        Task<ITextChannel> CreateTextChannelAsync(string name, RequestOptions options = null, Action<TextChannelProperties> func = null);
```
AntiTcb, line 94: I really don't think having the action being the last parameter is worth avoiding making a breaking change. Just way too inconsistent.

src/Discord.Net.Rest/API/Rest/CreateGuildChannelParams.cs
```diff
@@ -10,9 +10,20 @@ internal class CreateGuildChannelParams
         public string Name { get; }
         [JsonProperty("type")]
         public ChannelType Type { get; }
```
AntiTcb, line 12: This field's actually optional. 
<https://discordapp.com/developers/docs/resources/guild#create-guild-channel>
>All parameters for this endpoint are optional excluding `name`

AntiTcb (changes requested):

src/Discord.Net.Rest/API/Rest/CreateGuildChannelParams.cs
Joe4evr: I left that how I found it.

JaydenMaalouf (approved):
Can confirm this works! Channels (both text and voice alike) are now created within their desired categories, with their desired parameters.

src/Discord.Net.Core/Entities/Guilds/IGuild.cs
foxbot: yeah, this should go before RequestOptions for sure.

FiniteReality:
@Joe4evr Is this PR still alive?

PR #1022, opened 2018-04-07
Paginate reactions - solved #1007

HelpfulStranger999:
# Changes
This pull requests changes the nature of IUserMessage#GetReactionUsersAsync to paginate the discord endpoint, thereby solving #1007. However, the signature of this method has been changed according, meaning all calls will now be broken.

Ignore the other commits where my inexperience with Git was at work.


src/Discord.Net.Core/Entities/Messages/EmbedBuilder.cs
```diff
@@ -231,7 +231,6 @@ public class EmbedFieldBuilder
     {
         private string _name;
         private string _value;
-        private EmbedField _field;
```
SubZero0: Unrelated.

src/Discord.Net.Rest/Entities/Messages/MessageHelper.cs
```diff
@@ -46,13 +46,39 @@ internal static class MessageHelper
...
             Action<GetReactionUsersParams> func, BaseDiscordClient client, RequestOptions options)
         {
             var args = new GetReactionUsersParams();
             func(args);
-            string emoji = (emote is Emote e ? $"{e.Name}:{e.Id}" : emote.Name);
-            return (await client.ApiClient.GetReactionUsersAsync(msg.Channel.Id, msg.Id, emoji, args, options).ConfigureAwait(false)).Select(u => RestUser.Create(client, u)).ToImmutableArray();
+
+            var emoji = (emote is Emote e ? $"{e.Name}:{e.Id}" : emote.Name);
+            return new PagedAsyncEnumerable<IUser>(
+                DiscordConfig.MaxUsersPerBatch,
```
SubZero0: This value is wrong, it's 1000.
Create a new const (example: MaxUserReactionsPerBatch) with a value of 100 (don't reuse others or they might change and break this).

src/Discord.Net.Rest/Entities/Messages/MessageHelper.cs
```diff
@@ -46,13 +46,39 @@ internal static class MessageHelper
...
+                async (info, ct) =>
+                {
+                    if (info.Position != null)
+                        args.AfterUserId = info.Position.Value;
```
SubZero0: The correct way imo would be creating a local "args" with the page size and the position.

src/Discord.Net.Rest/Entities/Messages/MessageHelper.cs
```diff
@@ -46,13 +46,39 @@ internal static class MessageHelper
...
+                        builder.Add(RestUser.Create(client, model));
+
+                    return builder.ToImmutable();
+                },
+                nextPage: (info, lastPage) =>
+                {
+                    if (lastPage.Count != DiscordConfig.MaxUsersPerBatch)
+                        return false;
+
+                    info.Position = lastPage.OrderBy(u => u.Id).First().Id;
```
SubZero0: Use `Max(x => x.Id)`.

src/Discord.Net.Rest/Entities/Messages/MessageHelper.cs
```diff
@@ -46,13 +46,39 @@ internal static class MessageHelper
...
+                    return true;
+                },
+                count: args.Limit.Value
```
SubZero0: Shouldn't there be a `start:` here with the AfterUserId value?

src/Discord.Net.Rest/Entities/Messages/MessageHelper.cs
```diff
@@ -46,13 +46,39 @@ internal static class MessageHelper
...
-        public static async Task<IReadOnlyCollection<IUser>> GetReactionUsersAsync(IMessage msg, IEmote emote,
+        public static IAsyncEnumerable<IReadOnlyCollection<IUser>> GetReactionUsersAsync(IMessage msg, IEmote emote,
             Action<GetReactionUsersParams> func, BaseDiscordClient client, RequestOptions options)
         {
             var args = new GetReactionUsersParams();
             func(args);
-            string emoji = (emote is Emote e ? $"{e.Name}:{e.Id}" : emote.Name);
-            return (await client.ApiClient.GetReactionUsersAsync(msg.Channel.Id, msg.Id, emoji, args, options).ConfigureAwait(false)).Select(u => RestUser.Create(client, u)).ToImmutableArray();
+
+            var emoji = (emote is Emote e ? $"{e.Name}:{e.Id}" : emote.Name);
```
SubZero0: Better check for nulls before (msg and emote).
This will throw if emote is null.

SubZero0 (commented):
You should also change `DiscordRestApiClient.GetReactionUsersAsync`.

- Change the `MaxUsersPerBatch` there to your new const (that I said in the review).
- Change the `int.MaxValue` to your new const as well.
- Add `?limit={limit}&after={afterUserId}` to the endpoint string (or reference the other PR that only does this, this PR alone won't work).

src/Discord.Net.Rest/Entities/Messages/MessageHelper.cs
HelpfulStranger999: The message shouldn't be null, since the whole class is internal, and the places where it is called (Message objects) can't be null else Nullrefs will be thrown before it enters that method. So is it strictly necessary?

HelpfulStranger999: Or is it standard throughout the rest of the library?

SubZero0: You do have a point about the message, but `emote` could still receive null and throw.

src/Discord.Net.Rest/DiscordRestApiClient.cs
```diff
@@ -618,15 +618,15 @@ internal virtual void Dispose(bool disposing)
...
             Preconditions.GreaterThan(args.Limit, 0, nameof(args.Limit));
-            Preconditions.AtMost(args.Limit, DiscordConfig.MaxUsersPerBatch, nameof(args.Limit));
+            Preconditions.AtMost(args.Limit, DiscordConfig.MaxUserReactionsPerBatch, nameof(args.Limit));
             Preconditions.GreaterThan(args.AfterUserId, 0, nameof(args.AfterUserId));
             options = RequestOptions.CreateOrClone(options);
 
-            int limit = args.Limit.GetValueOrDefault(int.MaxValue);
-            ulong afterUserId = args.AfterUserId.GetValueOrDefault(0);
+            var limit = args.Limit.GetValueOrDefault(DiscordConfig.MaxUserReactionsPerBatch);
+            var afterUserId = args.AfterUserId.GetValueOrDefault(0);
```
SubZero0: I'm pretty sure the standard for this lib is to only use `var` when the type is explicit.

src/Discord.Net.Rest/Entities/Messages/MessageHelper.cs
```diff
@@ -46,13 +46,45 @@ internal static class MessageHelper
...
+            var arguments = new GetReactionUsersParams();
+            func(arguments);
+
+            return new PagedAsyncEnumerable<IUser>(
+                DiscordConfig.MaxUserReactionsPerBatch,
+                async (info, ct) =>
+                {
+                    var args = new GetReactionUsersParams();
+                    func(args);
+
```
SubZero0: `args.Limit = info.PageSize;`
To request the correct amount.

src/Discord.Net.Rest/Entities/Messages/MessageHelper.cs
```diff
@@ -46,13 +46,45 @@ internal static class MessageHelper
...
+                },
+                nextPage: (info, lastPage) =>
+                {
+                    if (lastPage.Count != DiscordConfig.MaxUsersPerBatch)
+                        return false;
+
+                    info.Position = lastPage.Max(x => x.Id);
+                    return true;
+                },
+                start: arguments.AfterUserId.Value,
```
SubZero0: This can throw.
`start: arguments.AfterUserId.IsSpecified ? args.AfterUserId.Value : (ulong?)null,`

src/Discord.Net.Rest/Entities/Messages/MessageHelper.cs
```diff
@@ -46,13 +46,45 @@ internal static class MessageHelper
...
+                count: arguments.Limit.Value
```
SubZero0: This could also throw but it's harder since it's internal. The best way would be:
`count: args.Limit.GetValueOrDefault(DiscordConfig.MaxUserReactionsPerBatch)`

src/Discord.Net.Rest/DiscordRestApiClient.cs
HelpfulStranger999: Ohhh, I see, ok

src/Discord.Net.Core/Entities/Messages/IUserMessage.cs
```diff
@@ -23,7 +23,7 @@ public interface IUserMessage : IMessage
         /// <summary> Removes all reactions from this message. </summary>
         Task RemoveAllReactionsAsync(RequestOptions options = null);
         /// <summary>  Gets all users that reacted to a message with a given emote </summary>
-        Task<IReadOnlyCollection<IUser>> GetReactionUsersAsync(IEmote emoji, int limit = 100, ulong? afterUserId = null, RequestOptions options = null);
+        IAsyncEnumerable<IReadOnlyCollection<IUser>> GetReactionUsersAsync(IEmote emoji, int limit = 100, ulong? afterUserId = null, RequestOptions options = null);
```
FiniteReality: Make limit required and remove afterUserId

src/Discord.Net.WebSocket/Entities/Messages/SocketUserMessage.cs
```diff
@@ -130,7 +130,7 @@ public Task RemoveReactionAsync(IEmote emote, IUser user, RequestOptions options
             => MessageHelper.RemoveReactionAsync(this, user, emote, Discord, options);
         public Task RemoveAllReactionsAsync(RequestOptions options = null)
             => MessageHelper.RemoveAllReactionsAsync(this, Discord, options);
-        public Task<IReadOnlyCollection<IUser>> GetReactionUsersAsync(IEmote emote, int limit = 100, ulong? afterUserId = null, RequestOptions options = null)
+        public IAsyncEnumerable<IReadOnlyCollection<IUser>> GetReactionUsersAsync(IEmote emote, int limit = 100, ulong? afterUserId = null, RequestOptions options = null)
```
FiniteReality: Make limit required and remove afterUserId

src/Discord.Net.Rest/Entities/Messages/RestUserMessage.cs
```diff
@@ -136,7 +136,7 @@ public Task RemoveReactionAsync(IEmote emote, IUser user, RequestOptions options
             => MessageHelper.RemoveReactionAsync(this, user, emote, Discord, options);
         public Task RemoveAllReactionsAsync(RequestOptions options = null)
             => MessageHelper.RemoveAllReactionsAsync(this, Discord, options);
-        public Task<IReadOnlyCollection<IUser>> GetReactionUsersAsync(IEmote emote, int limit = 100, ulong? afterUserId = null, RequestOptions options = null)
+        public IAsyncEnumerable<IReadOnlyCollection<IUser>> GetReactionUsersAsync(IEmote emote, int limit = 100, ulong? afterUserId = null, RequestOptions options = null)
```
FiniteReality: Make limit required and remove afterUserId

src/Discord.Net.Rest/Entities/Messages/MessageHelper.cs
```diff
@@ -46,13 +46,46 @@ internal static class MessageHelper
             await client.ApiClient.RemoveAllReactionsAsync(msg.Channel.Id, msg.Id, options);
         }
 
-        public static async Task<IReadOnlyCollection<IUser>> GetReactionUsersAsync(IMessage msg, IEmote emote,
+        public static IAsyncEnumerable<IReadOnlyCollection<IUser>> GetReactionUsersAsync(IMessage msg, IEmote emote,
```
FiniteReality: Remove GetReactionUserParams func here and add `ulong? from`, `int? limit` params instead

src/Discord.Net.Rest/Entities/Messages/MessageHelper.cs
```diff
@@ -46,13 +46,46 @@ internal static class MessageHelper
...
+            var emoji = (emote is Emote e ? $"{e.Name}:{e.Id}" : emote.Name);
+
+            var arguments = new GetReactionUsersParams();
+            func(arguments);
+
+            return new PagedAsyncEnumerable<IUser>(
+                DiscordConfig.MaxUserReactionsPerBatch,
+                async (info, ct) =>
+                {
+                    var args = new GetReactionUsersParams();
```
FiniteReality: Replace with ctor which sets Limits here, and remove func invocation

src/Discord.Net.Rest/Entities/Messages/MessageHelper.cs
```diff
@@ -46,13 +46,46 @@ internal static class MessageHelper
...
+                    func(args);
+                    args.Limit = info.PageSize;
+
+                    if (info.Position != null)
+                        args.AfterUserId = info.Position.Value;
+
+                    var models = await client.ApiClient.GetReactionUsersAsync(msg.Channel.Id, msg.Id, emoji, args, options).ConfigureAwait(false);
+                    var builder = ImmutableArray.CreateBuilder<IUser>();
```
FiniteReality: IMO this should be replaced with models.Select and you should call ToImmutableArray on it

src/Discord.Net.Rest/Entities/Messages/MessageHelper.cs
```diff
@@ -46,13 +46,46 @@ internal static class MessageHelper
...
             Action<GetReactionUsersParams> func, BaseDiscordClient client, RequestOptions options)
         {
-            var args = new GetReactionUsersParams();
-            func(args);
-            string emoji = (emote is Emote e ? $"{e.Name}:{e.Id}" : emote.Name);
-            return (await client.ApiClient.GetReactionUsersAsync(msg.Channel.Id, msg.Id, emoji, args, options).ConfigureAwait(false)).Select(u => RestUser.Create(client, u)).ToImmutableArray();
+            Preconditions.NotNull(emote, nameof(emote));
+            var emoji = (emote is Emote e ? $"{e.Name}:{e.Id}" : emote.Name);
+
+            var arguments = new GetReactionUsersParams();
```
FiniteReality: Remove this and the func invocation

src/Discord.Net.Rest/Entities/Messages/MessageHelper.cs
```diff
@@ -46,13 +46,46 @@ internal static class MessageHelper
...
+                nextPage: (info, lastPage) =>
+                {
+                    if (lastPage.Count != DiscordConfig.MaxUsersPerBatch)
+                        return false;
+
+                    info.Position = lastPage.Max(x => x.Id);
+                    return true;
+                },
+                start: arguments.AfterUserId.IsSpecified ? arguments.AfterUserId.Value : (ulong?)null,
+                count: arguments.Limit.GetValueOrDefault(DiscordConfig.MaxUserReactionsPerBatch)
```
FiniteReality: Replace these with from and limit suggested earlier

FiniteReality (changes requested):

src/Discord.Net.Core/Entities/Messages/IUserMessage.cs
SubZero0: Why make it required when we dont do that for `GetMessagesAsync` that follow a similar pattern? (Could change the `= 100` to `= DiscordConfig.MaxUserReactionsPerBatch` tho)
And why remove an option that the discord api gives to us?

FiniteReality: It looks like GetMessagesAsync is the only one where we have a default: everywhere else limit is required (and even in IMessageChannel, limit is required!) or is not present at all.

src/Discord.Net.Rest/Entities/Messages/RestUserMessage.cs
HelpfulStranger999: Which should afterUserId go then?

FiniteReality: It's handled as part of the async enumerable returned, so it's unnecessary.

HelpfulStranger999: How does the parameter get passed to MessageHelper.GetReactionUsersAsync then?

FiniteReality: It doesn't need to be; our async enumerable code handles it automatically.

HelpfulStranger999: Ohhhh, I see what you mean, where do I get the starting point for the Paged Enumerable then or should I leave that out?

FiniteReality: The paged enumerable defaults to 0, which would retrieve all of the results anyway.

HelpfulStranger999:
@FiniteReality Should be ready for you to re-review

FiniteReality (approved):

PR #1023, opened 2018-04-08
Fix/Implement various invite-related behaviors

Still34:
# Summary

This PR does 3 things.

1. Fix `IDiscordClient.GetInviteAsync` behaviour
    - Previously, this method would have returned much less information than what the API provided. This is fixed by changing the return type from `RestInvite` to `RestInviteMetadata`. 

2. Add the ability to get the approximated member/presence count provided by the API
    - The current design is debatable; I feel like we should just always return this count instead of having to add a new parameter asking whether the user want the count included.

3.  Deprecate the `IInvite.AcceptAsync` method
    - The accept endpoint has been deprecated by Discord around the end of March.

### DIFF
--- a/src/Discord.Net.Core/Entities/Invites/IInvite.cs
+++ b/src/Discord.Net.Core/Entities/Invites/IInvite.cs
@@ -1,4 +1,4 @@
-﻿using System.Threading.Tasks;
+using System.Threading.Tasks;
 
 namespace Discord
 {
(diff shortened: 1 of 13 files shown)


Still34:
Probably shouldn't create an extra arg and use `SendJsonAsync`, but use in-url style param like `AddBanAsync`. Will fix later.

PR #1026, opened 2018-04-10
Fix build fails when built on macOS and Linux

acid-chicken:
Adding some nodes to the project files.


src/Discord.Net.Core/Discord.Net.Core.csproj
```diff
@@ -1,14 +1,26 @@
...
+        <Description>The core components for the Discord.Net library.</Description>
+        <TargetFrameworks>net45;netstandard1.1;netstandard1.3</TargetFrameworks>
+      </PropertyGroup>
+    </When>
+    <Otherwise>
+      <PropertyGroup>
+        <AssemblyName>Discord.Net.Core</AssemblyName>
+        <RootNamespace>Discord</RootNamespace>
+        <Description>The core components for the Discord.Net library.</Description>
+        <TargetFrameworks>netstandard1.1;netstandard1.3</TargetFrameworks>
```
Still34: Couldn't we just change the `TargetFrameworks` and leave the rest out of the condition? You're duplicating the `AssemblyName` and such and isn't very clean.

acid-chicken: @Still34 Duplicating elements isn't clean indeed, but, unfortunately, the `When` element can have only one of `Choose`, `ItemGroup` and `PropertyGroup` elements as a child. Would you like to specify these properties with the `Condition` attributes?

Still34: Yes, we should be able to just use something like
```xml
  <PropertyGroup>
    <AssemblyName>Discord.Net.Core</AssemblyName>
    <Description>The core components for the Discord.Net library.</Description>
    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">net45;netstandard1.1;netstandard1.3</TargetFrameworks>
    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netstandard1.1;netstandard1.3</TargetFrameworks>
  </PropertyGroup>
```

src/Discord.Net.Core/Discord.Net.Core.csproj
```diff
@@ -1,14 +1,26 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="../../Discord.Net.targets" />
-  <PropertyGroup>
-    <AssemblyName>Discord.Net.Core</AssemblyName>
-    <RootNamespace>Discord</RootNamespace>
-    <Description>The core components for the Discord.Net library.</Description>
-    <TargetFrameworks>net45;netstandard1.1;netstandard1.3</TargetFrameworks>
-  </PropertyGroup>
+  <Choose>
+    <When Condition=" '$(OS)' == 'Windows_NT' ">
```
Joe4evr: Why even do this in the first place? #RemoveNet45Support

Still34: One can dream.

src/Discord.Net.Core/Discord.Net.Core.csproj
acid-chicken: @Still34 Fixed in 319bbcffad16e856fd23ffa19ea1c93b9324af7b

PR #1037, opened 2018-04-21
Add SocketUser.MutualGuilds + various command ext. methods.

AntiTcb:


src/Discord.Net.Commands/Extensions/CommandServiceExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+                if (result.IsSuccess)
+                    executableCommands.Add(command);
+            }
+
+            return executableCommands;
+        }
+        public static Task<IReadOnlyCollection<CommandInfo>> GetExecutableCommandsAsync(this CommandService commandService, ICommandContext context, IServiceProvider provider)
+            => GetExecutableCommandsAsync(commandService.Commands, context, provider);
+        public static Task<IReadOnlyCollection<CommandInfo>> GetExecutableCommandAsync(this ModuleInfo module, ICommandContext context, IServiceProvider provider)
+            => GetExecutableCommandsAsync(module.Commands, context, provider);
```
FiniteReality: I don't think ModuleInfo.Commands stores all of the submodule's commands; this may be confusing for some users who have submodules.

FiniteReality (changes requested):

src/Discord.Net.Commands/Extensions/CommandServiceExtensions.cs
AntiTcb: Hm. Optional flag to recurse through submodules then?

FiniteReality: I'd say make it recurse by default, no need for a flag.

src/Discord.Net.Commands/Extensions/CommandServiceExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+                if (result.IsSuccess)
+                    executableCommands.Add(command);
+            }
+
+            return executableCommands;
+        }
+        public static Task<IReadOnlyCollection<CommandInfo>> GetExecutableCommandsAsync(this CommandService commandService, ICommandContext context, IServiceProvider provider)
+            => GetExecutableCommandsAsync(commandService.Commands, context, provider);
+        public static Task<IReadOnlyCollection<CommandInfo>> GetExecutableCommandAsync(this ModuleInfo module, ICommandContext context, IServiceProvider provider)
+            => GetExecutableCommandsAsync(module.Commands.Concat(module.Submodules.SelectMany(sm => sm.Commands)), context, provider);
```
FiniteReality: This only recurses through the first level of commands; it won't work for something like this:
```cs
class ManageCommands
{
    class ManageAdminCommands
    {
        class ManageInviteLimitCommands
        {
            // commands here
        }
    }
}
```

FiniteReality (changes requested):

src/Discord.Net.Commands/Extensions/CommandServiceExtensions.cs
AntiTcb: Dang it, I had that thought in my head and forgot about it hours later when I actually opened VS again.

src/Discord.Net.Commands/Extensions/CommandServiceExtensions.cs
```diff
@@ -10,7 +9,7 @@ public static class CommandServiceExtensions
         public static async Task<IReadOnlyCollection<CommandInfo>> GetExecutableCommandsAsync(this IEnumerable<CommandInfo> commands, ICommandContext context, IServiceProvider provider)
         {
             var executableCommands = new List<CommandInfo>();
-
+
```
FiniteReality: Extraneous whitespace here

src/Discord.Net.Commands/Discord.Net.Commands.csproj
```diff
@@ -6,6 +6,9 @@
     <Description>A Discord.Net extension adding support for bot commands.</Description>
     <TargetFrameworks>netstandard1.1</TargetFrameworks>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|netstandard1.1|AnyCPU'">
```
FiniteReality: Why do we need C# 7.3?

FiniteReality (changes requested):

src/Discord.Net.Commands/Discord.Net.Commands.csproj
AntiTcb: Leftover from me fiddling with ValueTuples before deciding against them, forgot about this.

src/Discord.Net.Commands/Extensions/CommandServiceExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Discord.Commands
+{
+    public static class CommandServiceExtensions
+    {
+        public static async Task<IReadOnlyCollection<CommandInfo>> GetExecutableCommandsAsync(this IEnumerable<CommandInfo> commands, ICommandContext context, IServiceProvider provider)
+        {
+            var executableCommands = new List<CommandInfo>();
```
FiniteReality, line 12: Use `yield return` here instead of constructing a list

src/Discord.Net.Commands/Extensions/CommandServiceExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+                    executableCommands.Add(command);
+            }
+
+            return executableCommands;
+        }
+        public static Task<IReadOnlyCollection<CommandInfo>> GetExecutableCommandsAsync(this CommandService commandService, ICommandContext context, IServiceProvider provider)
+            => GetExecutableCommandsAsync(commandService.Commands, context, provider);
+        public static async Task<IReadOnlyCollection<CommandInfo>> GetExecutableCommandsAsync(this ModuleInfo module, ICommandContext context, IServiceProvider provider)
+        {
+            var executableCommands = new List<CommandInfo>();
```
FiniteReality, line 30: Again, `yield return` instead of constructing a list.

FiniteReality (changes requested):

src/Discord.Net.Commands/Extensions/CommandServiceExtensions.cs
AntiTcb: Can't yield return in async methods.

FiniteReality: Quick solution: have an internal function which does the main body, returning an IEnumerable<Task<CommandInfo>>, and then use Task.WhenAll to return an array for them all.

FiniteReality: Actually, I just looked at the docs, and it doesn't mention not being able to `yield return` in async methods: only anonymous methods and unsafe methods.

Joe4evr: [But you still can't do it.](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxAN0QSwGYE8A+ArlJgHYDmABAMq6xxgB0AwgPYA27cAxjJq6SiMA4nFJZuAbgCwAKGJkqtekwAqACwRwIAE0WNV0ANZQZsgA6Fg7TN0rd20KJWZyA3nMpfKl67crQuKR2hlBGADwAkgCipIRgiBDWcOFkMAB86ZQAsgAUAJSe3h6y3mUBAO4QmDCUoUaMAJqYcOw6BWbl3tisCJS5aZSYlAC8lAAMkkOU4ZQAjJNDANRLhaVd5bgtbZRaMIQIpEOd3gC+cqdAA)

Plus, you can't `yield return` when the return type is `IReadOnlyCollection`, so what now?

FiniteReality: Hmmpfh. I suppose so. In that case, I suppose there's no point jumping through hoops just to use yield return syntax.

AntiTcb: Task.WhenAll's probably a good idea though.

FiniteReality (approved):
~~LGTM~~ Wait one moment, I just realised something.

src/Discord.Net.Commands/Extensions/CommandServiceExtensions.cs
```diff
@@ -0,0 +1,42 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Discord.Commands
+{
+    public static class CommandServiceExtensions
+    {
+        public static async Task<IReadOnlyCollection<CommandInfo>> GetExecutableCommandsAsync(this IEnumerable<CommandInfo> commands, ICommandContext context, IServiceProvider provider)
```
FiniteReality: Now that I think about it, it's probably not a good idea to just allow `IEnumerable` for this. Maybe we should accept `ICollection<CommandInfo>` instead, since that's what we return.

FiniteReality (changes requested):
I missed some things when I was reviewing before >.<

src/Discord.Net.Commands/Extensions/CommandServiceExtensions.cs
AntiTcb: The `Commands` property on the service is exposed as an `IEnumerable`, so this change would make `GetExecutableCommandsAsync(commandService.Commands)` impossible without a type change/casting.

PR #1042, opened 2018-04-28
Makes text parameter of sending messages optional

HelpfulStranger999:
Makes the `text` parameter of `SendMessageAsync` optional, since it is not required if an embed or file is attached to the message. Non-breaking, useful change.

### DIFF
--- a/src/Discord.Net.Core/Entities/Channels/IMessageChannel.cs
+++ b/src/Discord.Net.Core/Entities/Channels/IMessageChannel.cs
@@ -8,7 +8,7 @@ namespace Discord
     public interface IMessageChannel : IChannel
     {
         /// <summary> Sends a message to this message channel. </summary>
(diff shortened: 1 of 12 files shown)


Yucked:
Is this necessary? Users can use `string.Empty` if they don't wanna use the text parameter.

Giggitybyte:
I feel it'd be a nice addition. I'd love to pass in an embed without the need to `String.Empty` all the time.

FiniteReality:
This won't work. `text` is a required parameter API side, so passing `null` will cause an error.

HelpfulStranger999:
Oh it is? I thought it was either an embed, file, or content

FiniteReality:
Try passing null explicitly. It'll error.

AntiTcb:
I think the docs are out of date and `content` actually is nullable (at least, when creating a message with an embed).
image

PR #1043, opened 2018-04-29
 Add support casting System.Drawing.Color to Discord.Color 

acid-chicken:
.NET Standard 2.0 has [`Color` Struct](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.color?view=netstandard-2.0), so I want to support it.


ObsidianMinor:
You might want to use the built-in `NETSTANDARD2_0` conditional instead of rolling your own. [(see here)](https://docs.microsoft.com/en-us/dotnet/standard/frameworks#how-to-specify-target-frameworks)

acid-chicken:
@ObsidianMinor Thank you for your advice. Updated.

PR #1046, opened 2018-04-30
Update Dependencies and support NS2.0 builds

Hawxy:
* Since #1043 is adding NS2.0 to Core, I thought I'd go through and add it to all packages.
* Remove NS 1.1 support since net45+ns1.3 builds essentially cover the [exact same platforms](https://docs.microsoft.com/en-us/dotnet/standard/net-standard) without the need for constants
* Removed obsolete JSON.net call in DiscordContractResolver.
* Also fixed various warnings within Tests due to the updates


FiniteReality:
Have you tested building on multiple platforms using different targets? I don't want to accidentally remove support for some configuration of .NET/OS/whatever which we may have supported previously.

Hawxy:
ns1.1 -> ns1.3 only major change is the increase in .NETFX minimum from 4.5 to 4.6, which is backfilled with the net45 builds I added. 

ns1.3 is supported by .NET Core 1.0 & UWP < W10 1709
ns2.0 is supported by .NET Core 2.0 & UWP ≥ W10 1709
net45 maintains support for .NET 4.5+

Nothing is changed in terms of platform support.

PR #1051, opened 2018-05-04
Fix issues with #984, remove extraneous whitespace

FiniteReality:
- Removed unnecessary parameter in SocketVoiceServer
- Moved SocketVoiceServer into `Entities/Voice`
- Fixed a bug where trying to download the cached guild would return null due to invalid cast (not throw as the commit description wrongly states)
- Fixed a potential bug where Discord might not give us a port when connecting to voice

Since we don't have access to GetGuildAsync from the rest client, I decided to make the Cacheable pull from the state. I'm not sure if this is the best approach, but it's infinitely better than doing an invalid cast


FiniteReality:
Another thing: I'm not sure what the recommended way to stop the built-in voice client from connecting is. Throwing an exception seems ugly, but adding a flag of some sort into SocketVoiceServer seems even uglier.

PR #1055, opened 2018-05-05
Audit Logs implementation

FiniteReality:
I suck at git, so after fucking up one branch locally I'm just gonna brute-force copy files in and save myself the hassle.

Apparently audit logs have changed a little, but I'm not sure what has changed so things may be missing. However, I did manage to successfully pull 1000 audit log entries from Discord API and they all seemed to be complete.

I have changed some things since the last PR, namely adding read-only structs to store values instead of using the built-in read-write types I had before. I also removed a bunch of TODOs as I realise it wouldn't be worth grabbing channels for each entry.

### DIFF
--- a/src/Discord.Net.Rest/Entities/AuditLogs/AuditLogHelper.cs
+++ b/src/Discord.Net.Rest/Entities/AuditLogs/AuditLogHelper.cs
@@ -1,0 +1,79 @@
+﻿using Model = Discord.API.AuditLog;
+using EntryModel = Discord.API.AuditLogEntry;
+
+namespace Discord.Rest
+{
+    internal static class AuditLogHelper
+    {
+        public static IAuditLogData CreateData(BaseDiscordClient discord, Model log, EntryModel entry)
+        {
+            switch (entry.Action)
+            {
+                case ActionType.GuildUpdated: //1
+                    return GuildUpdateAuditLogData.Create(discord, log, entry);
+
+                case ActionType.ChannelCreated: //10
+                    return ChannelCreateAuditLogData.Create(discord, log, entry);
+                case ActionType.ChannelUpdated:
+                    return ChannelUpdateAuditLogData.Create(discord, log, entry);
+                case ActionType.ChannelDeleted:
+                    return ChannelDeleteAuditLogData.Create(discord, log, entry);
+                case ActionType.OverwriteCreated:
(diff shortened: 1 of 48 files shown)


src/Discord.Net.Rest/Entities/AuditLogs/AuditLogHelper.cs
```diff
@@ -0,0 +1,79 @@
+﻿using Model = Discord.API.AuditLog;
+using EntryModel = Discord.API.AuditLogEntry;
+
+namespace Discord.Rest
+{
+    internal static class AuditLogHelper
+    {
+        public static IAuditLogData CreateData(BaseDiscordClient discord, Model log, EntryModel entry)
```
foxbot: I'm keen to investigating if there's a better solution for this method - whether it be scoped methods, or a class-instance. Passing and  managing three public parameters for this method, with some 40 invocations (or however many there are below) seems like it will become a hassle in the future.

foxbot (commented):
looks good to me, just cleanup TODOs

src/Discord.Net.Rest/Entities/AuditLogs/AuditLogHelper.cs
FiniteReality: Yeah, I couldn't really think of a better way when I implemented it initially, and I wanted to keep clutter in RestAuditLogEntry low. The only other approach I could see working right now is creating a dictionary and using that, e.g.:
```cs
private static Dictionary<ActionType, Func<BaseDiscordClient, Model, EntryModel, IAuditLogData>> CreateMapping =
{
    {ActionType.GuildUpdated, GuildUpdateAuditLogData.Create}
    // etc
};

public static IAuditLogData CreateData(BaseDiscordClient discord, Model log, EntryModel entry)
    => CreateMapping[entry.Action](discord, log, entry);
```

foxbot: I kind of like that one more truthfully

FiniteReality: Mkay, I'll use that approach then.

FiniteReality:
Should be ready for re-review @foxbot :+1:

PR #1057, opened 2018-05-09
Allow external VoiceChannel client API implementation

NovusTheory:
Following what #984 offered and building upon the knowledge of being able to implement Lavalink there was a few inconsistencies with it that seemed to cause a lot of problems and didn't let Lavalink work properly. These changes are meant to allow developers to mark connections to a Discord voice channel as externally handled like with a Lavalink implementation.

These API changes just bypass the Discord.Net AudioClient creation process and just go straight to sending the voice state update payload.

**Reasoning for change regarding #984**
This pull request was to add the VoiceServerUpdate event to DiscordSocketClient but was also modified to allow an exception to be thrown back to prevent the default AudioClient, this had *major* problems and did not work as expected. The session id was still getting invalidated, discord.net was still creating a client and throwing session invalid errors, and if the VoiceServerUpdate event was Task.Delay'd the AudioClient would not throw a session invalid error on the websocket but the session id when used by Lavalink was invalid (This basically left a sort of dead AudioClient created that logged unnecessarily)

**API Changes**
IAudioChannel.ConnectAsync(Action<AudioClient> configAction, bool external) 
    => If external is true AudioClient is null and configAction is not called
IAudioChannel.DisconnectAsync()
    => DisconnectAsync calls the SocketGuild.DisconnectAudioAsync() method. In the case of an external audio client this API exists to send a voice state update payload where an AudioClient doesn't exist.

~~*These API changes **DO NOT** affect anything majorly but just add onto the current API*~~
Edit: After review and other changes not related to to the "opening pr commits" it now has breaking API changes but was stated that it was apis which were likely never used and won't break any current bots

### DIFF
--- a/src/Discord.Net.Core/Entities/Channels/IAudioChannel.cs
+++ b/src/Discord.Net.Core/Entities/Channels/IAudioChannel.cs
@@ -7,7 +7,7 @@ namespace Discord
     public interface IAudioChannel : IChannel
     {
         /// <summary> Connects to this audio channel. </summary>
-        Task<IAudioClient> ConnectAsync(Action<IAudioClient> configAction = null, bool external = false);
+        Task<IAudioClient> ConnectAsync(bool selfDeaf = false, bool selfMute = false, bool external = false);
 
         /// <summary> Disconnects from this audio channel. </summary>
         Task DisconnectAsync();
--- a/src/Discord.Net.Core/Entities/Channels/IAudioChannel.cs
+++ b/src/Discord.Net.Core/Entities/Channels/IAudioChannel.cs
@@ -1,4 +1,4 @@
-﻿using Discord.Audio;
+using Discord.Audio;
 using System;
 using System.Threading.Tasks;
 
@@ -8,5 +8,8 @@ namespace Discord
     {
         /// <summary> Connects to this audio channel. </summary>
         Task<IAudioClient> ConnectAsync(Action<IAudioClient> configAction = null);
+
+        /// <summary> Connects to this audio channel but can specify if client is handled externally. </summary>
+        Task<IAudioClient> ConnectAsync(bool external, Action<IAudioClient> configAction = null);
     }
 }
--- a/src/Discord.Net.Core/Entities/Channels/IAudioChannel.cs
+++ b/src/Discord.Net.Core/Entities/Channels/IAudioChannel.cs
@@ -10,6 +10,9 @@ namespace Discord
         Task<IAudioClient> ConnectAsync(Action<IAudioClient> configAction = null);
 
         /// <summary> Connects to this audio channel but can specify if client is handled externally. </summary>
-        Task<IAudioClient> ConnectAsync(bool external, Action<IAudioClient> configAction = null);
+        Task<IAudioClient> ConnectAsync(Action<IAudioClient> configAction = null, bool external = false);
+
+        /// <summary> Disconnects from this audio channel if applicable. </summary>
+        Task DisconnectAsync();
     }
 }
--- a/src/Discord.Net.Core/Entities/Channels/IAudioChannel.cs
+++ b/src/Discord.Net.Core/Entities/Channels/IAudioChannel.cs
@@ -12,7 +12,7 @@ namespace Discord
         /// <summary> Connects to this audio channel but can specify if client is handled externally. </summary>
         Task<IAudioClient> ConnectAsync(Action<IAudioClient> configAction = null, bool external = false);
 
-        /// <summary> Disconnects from this audio channel if applicable. </summary>
+        /// <summary> Disconnects from this audio channel. </summary>
         Task DisconnectAsync();
     }
 }
--- a/src/Discord.Net.Core/Entities/Channels/IAudioChannel.cs
+++ b/src/Discord.Net.Core/Entities/Channels/IAudioChannel.cs
@@ -7,9 +7,6 @@ namespace Discord
     public interface IAudioChannel : IChannel
     {
         /// <summary> Connects to this audio channel. </summary>
-        Task<IAudioClient> ConnectAsync(Action<IAudioClient> configAction = null);
-
-        /// <summary> Connects to this audio channel but can specify if client is handled externally. </summary>
         Task<IAudioClient> ConnectAsync(Action<IAudioClient> configAction = null, bool external = false);
 
         /// <summary> Disconnects from this audio channel. </summary>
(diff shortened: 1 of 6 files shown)


src/Discord.Net.Core/Entities/Channels/IAudioChannel.cs
```diff
@@ -8,5 +8,11 @@ public interface IAudioChannel : IChannel
     {
         /// <summary> Connects to this audio channel. </summary>
         Task<IAudioClient> ConnectAsync(Action<IAudioClient> configAction = null);
+
+        /// <summary> Connects to this audio channel but can specify if client is handled externally. </summary>
+        Task<IAudioClient> ConnectAsync(Action<IAudioClient> configAction = null, bool external = false);
```
FiniteReality: Wouldn't this cause an error when you call `ConnectAsync((client) => {});`? I'd suggest you merge both of these methods.

src/Discord.Net.WebSocket/Entities/Guilds/SocketGuild.cs
```diff
@@ -518,29 +518,37 @@ internal AudioInStream GetAudioStream(ulong userId)
                 promise = new TaskCompletionSource<AudioClient>();
                 _audioConnectPromise = promise;
 
-                if (_audioClient == null)
+                if (!external)
```
FiniteReality: I think it might be cleaner if we invert this condition and bail-out instead

src/Discord.Net.WebSocket/Entities/Guilds/SocketGuild.cs
```diff
@@ -504,7 +504,7 @@ internal AudioInStream GetAudioStream(ulong userId)
         {
             return _audioClient?.GetInputStream(userId);
         }
-        internal async Task<IAudioClient> ConnectAudioAsync(ulong channelId, bool selfDeaf, bool selfMute, Action<IAudioClient> configAction)
+        internal async Task<IAudioClient> ConnectAudioAsync(ulong channelId, bool selfDeaf, bool selfMute, Action<IAudioClient> configAction, bool external)
         {
             selfDeaf = false;
```
FiniteReality: A little unrelated, but @foxbot is there any reason we explicitly set these to false here?

src/Discord.Net.WebSocket/Entities/Channels/SocketVoiceChannel.cs
```diff
@@ -42,9 +42,17 @@ public Task ModifyAsync(Action<VoiceChannelProperties> func, RequestOptions opti
 
         public async Task<IAudioClient> ConnectAsync(Action<IAudioClient> configAction = null)
         {
-            return await Guild.ConnectAudioAsync(Id, false, false, configAction).ConfigureAwait(false);
+            return await Guild.ConnectAudioAsync(Id, false, false, configAction, false).ConfigureAwait(false);
         }
 
+        public async Task<IAudioClient> ConnectAsync(Action<IAudioClient> configAction = null, bool external = false)
+        {
+            return await Guild.ConnectAudioAsync(Id, false, false, configAction, external).ConfigureAwait(false);
```
FiniteReality: Potentially instead of passing null for deaf/mute here we could add extra params?

src/Discord.Net.WebSocket/Entities/Guilds/SocketGuild.cs
```diff
@@ -518,29 +518,37 @@ internal AudioInStream GetAudioStream(ulong userId)
...
+                                    await promise.TrySetCanceledAsync();
+                                return;
+                            }
+                        };
+                        audioClient.Connected += () =>
+                        {
+                            var _ = promise.TrySetResultAsync(_audioClient);
+                            return Task.Delay(0);
+                        };
+                        configAction?.Invoke(audioClient);
```
FiniteReality: Is this even necessary anymore? I think we might be able to remove it without breaking any existing code, even if it would be a breaking change.

FiniteReality (changes requested):
I like the idea, and it definitely seems to go hand-in-hand with the voice state update events, but there are a few things I think we could change. /cc @foxbot

src/Discord.Net.WebSocket/Entities/Guilds/SocketGuild.cs
FiniteReality: Okay, remove these assignments so that it can be passed up from the parent call.

src/Discord.Net.WebSocket/Entities/Channels/SocketVoiceChannel.cs
```diff
@@ -40,11 +40,14 @@ internal override void Update(ClientState state, Model model)
         public Task ModifyAsync(Action<VoiceChannelProperties> func, RequestOptions options = null)
             => ChannelHelper.ModifyAsync(this, Discord, func, options);
 
-        public async Task<IAudioClient> ConnectAsync(Action<IAudioClient> configAction = null)
+        public async Task<IAudioClient> ConnectAsync(Action<IAudioClient> configAction = null, bool external = false)
         {
-            return await Guild.ConnectAudioAsync(Id, false, false, configAction).ConfigureAwait(false);
+            return await Guild.ConnectAudioAsync(Id, false, false, configAction, external).ConfigureAwait(false);
```
FiniteReality: Add params for selfDeaf and selfMute and pass them here

src/Discord.Net.Core/Entities/Channels/IAudioChannel.cs
```diff
@@ -7,6 +7,9 @@ namespace Discord
     public interface IAudioChannel : IChannel
     {
         /// <summary> Connects to this audio channel. </summary>
-        Task<IAudioClient> ConnectAsync(Action<IAudioClient> configAction = null);
+        Task<IAudioClient> ConnectAsync(Action<IAudioClient> configAction = null, bool external = false);
```
FiniteReality: Add params for mute/deaf to the interfaces and impls too

FiniteReality: Furthermore, remove the Action here and in the implementations too

src/Discord.Net.WebSocket/Entities/Guilds/SocketGuild.cs
```diff
@@ -504,7 +504,7 @@ internal AudioInStream GetAudioStream(ulong userId)
         {
             return _audioClient?.GetInputStream(userId);
         }
-        internal async Task<IAudioClient> ConnectAudioAsync(ulong channelId, bool selfDeaf, bool selfMute, Action<IAudioClient> configAction)
+        internal async Task<IAudioClient> ConnectAudioAsync(ulong channelId, bool selfDeaf, bool selfMute, Action<IAudioClient> configAction, bool external)
```
FiniteReality: Remove the configAction parameter as we don't need it

FiniteReality (approved):

PR #1070, opened 2018-05-27
Improve XML docs for commands quotation mark aliases feature

Chris-Johnston:
I realized that the XML docs around `QuotationMarkAliasMap` were either incorrect or not up to a level of quality I'm happy with (compared to the recent documentation effort).

This improves the documentation by explaining the default behavior and providing an example.

I'm unsure if this is a big enough topic to warrant it's own guide in the documentation, but I would like to get thoughts on that. If it might be helpful I could write something up.


Chris-Johnston:
Moved to Still's Docs PR. (https://github.com/Still34/Discord.Net/pull/1)

PR #1072, opened 2018-05-27
Added AudioLock & Modified Throw

ComputerMaster1st:
AudioLock << Copied to catch block to release before throwing an exception. Previously, this was known to lockup & prevent new audio clients from spawning.

Modified Throw << Now throws the caught exception. This previously wasn't producing a stacktrace.


ComputerMaster1st:
This is currently under real-world testing to confirm it's working.

PR #1073, opened 2018-05-28
feature: Add OffloadAllHandlers to config

foxbot:
## Summary

This change allows users to request that all event handlers be invoked off of the gateway thread.

### Additions

- Added `OffloadAllHandlers` to `DiscordSocketConfig`.

## Possible Caveats

Setting this flag to true may lead to data races, users may ignorantly set this flag without being aware of the potential consequences.


FiniteReality:
I don't like this idea. I think a better solution would be to provide a helper method to convert a handler to an offloaded handler, e.g.:

```cs
client.MessageReceived += EventHelpers.OffloadToThread(OnMessageReceived);
```

This way, we can set the "flag" for specific event handlers rather than all of them, which would reduce the chance for race conditions to occur.

FiniteReality:
Another solution, as suggested by Voltana, would be to implement an event queue system for these instead. That could be potentially incorporated into my previous suggestion, so that way our events are still in order, yet they're detached from the main lib.

PR #1081, opened 2018-05-31
Update Discord.Net.Core.csproj

xXBuilderBXx:
Package fix for https://github.com/RogueException/Discord.Net/issues/1080


Still34:
I would probably do conditional package reference if possible.

PR #1093, opened 2018-06-23
Solved 4003 Authentication Error (Race Condition)

ComputerMaster1st:
Hi There,

The current build (00969) with the change to VoiceAPI is producing a 4003 Authentication Error. This was caused by a possible race condition created by sending the heartbeat before the identity payload.

This commit solves this by having the heartbeat start after receiving the ready payload. The discord developer documents also shows the identity payload being sent before performing heartbeat payloads. I've added a local variable to store the interval being received by the hello payload.

This change has been tested with NETCore 2.1. It is unknown if this works with 2.0, although there shouldn't be a problem.

**NOTE :** (Please refer to discord dev docs) There is a bug with the hello payload which is sending the incorrect heartbeat interval. This should be changed in the future or you can make the interval a static integer value of 41250.

Regards,

CM1

### DIFF
--- a/src/Discord.Net.WebSocket/Audio/AudioClient.cs
+++ b/src/Discord.Net.WebSocket/Audio/AudioClient.cs
@@ -45,6 +45,7 @@ namespace Discord.Audio
         private ulong _userId;
         private uint _ssrc;
         private bool _isSpeaking;
+        private int _heartbeatInterval = 0;
 
         public SocketGuild Guild { get; }
         public DiscordVoiceAPIClient ApiClient { get; private set; }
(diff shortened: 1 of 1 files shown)


src/Discord.Net.WebSocket/Audio/AudioClient.cs
```diff
@@ -221,7 +222,7 @@ internal AudioInStream GetInputStream(ulong id)
                             await _audioLogger.DebugAsync("Received Hello").ConfigureAwait(false);
                             var data = (payload as JToken).ToObject<HelloEvent>(_serializer);
 
-                            _heartbeatTask = RunHeartbeatAsync(data.HeartbeatInterval, _connection.CancelToken);
+                            _heartbeatInterval = (data.HeartbeatInterval / 4) * 3;
```
foxbot: I'd rather just hardcode this value than introduce a potential data-race where for whatever reason the HELLO event isn't parsed properly or never received.

ComputerMaster1st:
Made changes.

foxbot (approved):

PR #1094, opened 2018-06-23
Mark guild optional in invite & general invite improvement

Still34:
# Changes

* Per https://github.com/discordapp/discord-api-docs/pull/614, guild information is now considered optional for invite objects.
* Mark several `InviteMetadata` members as optional, as some of the data are not sent when using the generic GET invite endpoint.
* Move `ChannelType` property from REST to Core; add `ChannelType` to invite object
+ Add `GetVanityInviteAsync` for guilds that support vanity invite URL.
- Remove `GetInviteParams` to always get presence and member count since it's trivial to ask the user about whether to fill those 2 fields or not.


Joe4evr:
LGTM. :shipit:

PR #1095, opened 2018-06-23
Fixed AudioClient Disconnect Deadlock & UdpClient ObjectDisposedException

ComputerMaster1st:
Hi There,

I've found (after going back & forth a lot) that when calling DisconnectAsync within the ConnectionManager "_task", that invokes the Disconnected AsyncEvent. After going through all the code, I found the event handler that was attached to AudioClient.Disconnect in SocketGuild called the audioClient.Dispose. This goes onto calling stopasync which then goes onto the stopasync in the connection manager.

The deadlock was found when awaiting the "_task".

TL:DR : "_task" was awaiting on itself to finish before completing the disconnected event.

This commit solves the deadlock by simply removing the await "_task" from stopasync.

This commit has been tested numerous times by recreating the 4003 auth error exception to see if the deadlock was no more.

As a side note, I've moved the disconnectevent.invoke before the logger to also further prove the deadlock is gone.

Regards,

CM1


ComputerMaster1st:
UdpClient (located in DefaultUdpSocket) is already being disposed when stopped and was causing ObjectAlreadyDisposed Exceptions within RunAsync method. Try..Catch simply swallows it & breaks.

Unobserved event from Task scheduler kept pinging about this.

LDSpits:
the changes in the DefaultUdpClient were made because of hidden exceptions. The previous implementation of ``RunAsync``was using ``Task.WhenAny()`` along with 2 tasks, 1 for cancellation and one for recieving the udp data.

a weird aspect of the  udp socket is that it throws a ``ObjectDisposedException`` when the async method ``RecieveAsync`` is running (and being cancelled) if the client was disposed of while running. however the ``Task.WhenAny()``  ignores (or swallowed) the faulty task. so it would be propagated to the unobservedtaskexception event instead. 

The resulting behaviour was us seeing ``ObjectdisposedException``s if you hooked into the ``UnobservedTaskExceptions`` event.

Hope this clarifies the resulting changes 😄

FiniteReality:
UDP does not have Dispose in .NET 4.5 - please test your code in all supported platforms. Furthermore, dispose should not be used as a control flow, which is why the code was as it was before.

LDSpits:
> UDP does not have Dispose in .NET 4.5

the discord.net.core only contains a reference to .net 4.6 ([here](https://github.com/RogueException/Discord.Net/blob/322d46e47b47e44f8b62b1ddcdcf39280cac6771/src/Discord.Net.Core/Discord.Net.Core.csproj#L7)) the only thing targeting net4.5 is the nuget specification. Is the nuget correct or the projects? @FiniteReality

FiniteReality:
Slight correction, my bad: it's there since since .NET 4.5 - but the issue about relying on dispose behaviour is still bad

LDSpits:
alright, as an alternative we can do this
```cs
private async Task RunAsync(CancellationToken cancelToken)
        {
            var closeTask = Task.Delay(-1, cancelToken);
            while (!cancelToken.IsCancellationRequested)
            {
                var receiveTask = _udp.ReceiveAsync();

#pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
                receiveTask.ContinueWith((recieveResult) =>
                {
                    //observe the exception as to not recieve as unhandled exception
                    _ = recieveResult.Exception;

                }, TaskContinuationOptions.OnlyOnFaulted);
#pragma warning restore CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed

                var task = await Task.WhenAny(closeTask, receiveTask).ConfigureAwait(false);
                if (task == closeTask)
                    break;

                var result = receiveTask.Result;
                await ReceivedDatagram(result.Buffer, 0, result.Buffer.Length).ConfigureAwait(false);
            }
```
which would still observe the exception and (mostly) keep using the original implementation

would that be acceptable? @FiniteReality

ComputerMaster1st:
I've applied what @LDSpits has suggested & tested via netcore 2.1 (I don't have any other runtime sadly). The ObjectDisposedException is no longer popping up. Task Scheduler doesn't see it either.

ComputerMaster1st:
Just as a side note, we've seem to have indirectly solved a possible memory leak within the library for audio. I've been monitoring this for a while now. I already knew the library had a memory leak in it due to the blocked code that got fixed in my previous pull request. However, there was still another possible memory leak. This pull request indirectly seemed to have solved it.

src/Discord.Net.WebSocket/ConnectionManager.cs
```diff
@@ -106,9 +106,7 @@ internal class ConnectionManager
         public virtual async Task StopAsync()
         {
             Cancel();
-            var task = _task;
-            if (task != null)
-                await task.ConfigureAwait(false);
+            await Task.CompletedTask;
```
foxbot: style: unflag this task as async, return `Task.CompletedTask`

src/Discord.Net.WebSocket/ConnectionManager.cs
```diff
@@ -159,9 +157,9 @@ internal class ConnectionManager
 
             await _onDisconnecting(ex).ConfigureAwait(false);
 
-            await _logger.InfoAsync("Disconnected").ConfigureAwait(false);
-            State = ConnectionState.Disconnected;
             await _disconnectedEvent.InvokeAsync(ex, isReconnecting).ConfigureAwait(false);
+            State = ConnectionState.Disconnected;
```
foxbot, line 161: Why were these moved to below the event invocation?

src/Discord.Net.WebSocket/Net/DefaultUdpSocket.cs
```diff
@@ -117,6 +117,16 @@ public void SetCancelToken(CancellationToken cancelToken)
             while (!cancelToken.IsCancellationRequested)
             {
                 var receiveTask = _udp.ReceiveAsync();
+
+#pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
```
foxbot: style: remove these, use a discard on `receiveTask` instead

foxbot (commented):
here are some style changes at least, i'm still not sure whether or not we'll merge this though

would rather find a proper cause+fix than just a hacky patch, as this is probably related to the regular socket client's disconnect deadlocks as well

src/Discord.Net.WebSocket/ConnectionManager.cs
ComputerMaster1st: For this, you simply wish to have the `async` within the method removed?

src/Discord.Net.WebSocket/ConnectionManager.cs
ComputerMaster1st: It was moved below to check if the the disconnect deadlock for audio was fixed since async events were expecting a task to be returned which causes deadlocks (depending on the code within the event handler).

src/Discord.Net.WebSocket/Net/DefaultUdpSocket.cs
ComputerMaster1st: @LDSpits You've handled this part afaik.

ComputerMaster1st:
Hi There,

I'm going to close this pull request as too many changes have been made and may or may not be of use to the library.

Regards,

CM1

PR #1097, opened 2018-06-26
RemarksAttribute name to RemarkAttribute

Kiritsu:
Changed RemarksAttribute to RemarkAttribute because the attribute usage is set to `AllowMultiple = false` so the 's' in 'Remarks' was misleading.

### DIFF
--- a/src/Discord.Net.Commands/Attributes/RemarkAttribute.cs
+++ b/src/Discord.Net.Commands/Attributes/RemarkAttribute.cs
@@ -4,11 +4,11 @@ namespace Discord.Commands
 {
     // Extension of the Cosmetic Summary, for Groups, Commands, and Parameters
     [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class, AllowMultiple = false, Inherited = true)]
-    public class RemarksAttribute : Attribute
+    public class RemarkAttribute : Attribute
     {
         public string Text { get; }
 
-        public RemarksAttribute(string text)
+        public RemarkAttribute(string text)
         {
             Text = text;
         }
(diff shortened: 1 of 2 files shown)


FiniteReality:
As misleading as it is, I believe `Remarks` reads better than `Remark`.

k-boyle:
Or allow us to have multiple `Remarks`

AntiTcb:
The attribute effectively is a clone of https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/remarks. Nothing is wrong with the name of it necessarily, and it really doesn't warrant the breaking change of renaming it.

Still34:
The word `Remark` implies a general comment on something. When it comes to adding additional comments on an item in an elaborate fashion, you'd want to use the term `Remarks` since you are explaining the said item in details; in other words, when using commenting system like documentation or cosmetic attribute like `RemarksAttribute` to explain to the user what the method does, of course you'd want to add more than one single comment; the purpose of them is describe the behavior or purpose in details.

In laymen's term, here's an example
***
I have a command that scours through the internet to find an image. I want to explain to the user what this command does. 

A "remark" would be something like "This command searches on Bing for the desired item."; "remark**s**" would be something like "This command searches on Bing for the desired item. If an item is not found, a fallback image would be used; if a suitable image is found, relative information such as its resolution and its URL would be returned."; in a sense, each individual sentence is a "remark" of its own, an additional comment.
***
Normally when you want to explain something in further details, you'd want to be as verbose as possible depending on the target audience, so which means the term "remarks" is more commonly used than "remark," hence the naming of `remarks` across various documentation utility, or in this case, a cosmetic attribute that plays such role.

PR #1098, opened 2018-07-01
Added GetJumpUrl() to IMessage, RestMessage, SocketMessage, and MessageHelper

k-boyle:
Used a Get method instead of a property to be consistent with GetAvatarUrl and other GetUrl methods. The `"https://discordapp.com/"` can be exchanged with `DiscordConfig.CDNUrl` and the links work fine still. Except this removes the embed preview from the links. This has been tested in both dms and guilds by checking the value of `GetJumpUrl()` on received and sent messages. I think I got all occurrences of where this should be implemented.

### DIFF
--- a/src/Discord.Net.WebSocket/Entities/Messages/SocketMessage.cs
+++ b/src/Discord.Net.WebSocket/Entities/Messages/SocketMessage.cs
@@ -1,4 +1,4 @@
-﻿using Discord.Rest;
+using Discord.Rest;
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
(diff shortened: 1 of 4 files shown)


ObsidianMinor:
Why make it a method as part of the interface for IMessage? Why not make it an extension method?

FiniteReality:
This sort of thing looks more useful as an extension method, yes.

k-boyle:
Okay, I originally added it to `IMessage` to see where I'd need to be implementing it, will remove it now.

Joe4evr:
Default Interface Implementations *soon* :tm:.

src/Discord.Net.WebSocket/Entities/Messages/SocketMessage.cs
```diff
@@ -58,6 +58,10 @@ public Task DeleteAsync(RequestOptions options = null)
             => MessageHelper.DeleteAsync(this, Discord, options);
 
         public override string ToString() => Content;
+
```
foxbot, line 61: As per other comments, this would best exist as an extension method on IMessage.

foxbot (changes requested):

PR #1102, opened 2018-07-02
Added GetJumpUrl() as an extension method for IMessage

k-boyle:
Tested on messages sent and received. Hoping it being in this namespace is fine as it's not related to commands but this was where the MessageExtensions was. Sorry for the extra PR I'm quite new to GitHub

### DIFF
--- a/src/Discord.Net.Core/Extensions/MessageExtensions.cs
+++ b/src/Discord.Net.Core/Extensions/MessageExtensions.cs
@@ -1,0 +1,11 @@
+namespace Discord
+{
+    public static class MessageExtensions
+    {
+        public static string GetJumpUrl(this IMessage msg)
+        {
+            var channel = msg.Channel;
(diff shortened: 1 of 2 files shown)


src/Discord.Net.Core/Extensions/MessageExtensions.cs
```diff
@@ -0,0 +1,11 @@
+namespace Discord
+{
+    public static class MessageExtensions
+    {
+        public static string GetJumpUrl(this IMessage msg)
+        {
+            var channel = msg.Channel;
+            return $"https://discordapp.com/channels/{(channel is IDMChannel ? "@me" : $"{(channel as IGuildChannel).GuildId}")}/{channel.Id}/{msg.Id}";
```
Still34: I'd cast it the latter half as `ITextChannel` instead, since we know it can't be any other type of channel - or go as far as pattern matching in case it cannot be casted as such.

foxbot (approved):

PR #1107, opened 2018-07-14
Add missing fields to new MemberInfo type

### LINKED ISSUE #1105: MemberUpdateAuditLogData does not correctly handle all cases
Particularly, here: https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/MemberUpdateAuditLogData.cs#L20-L23 - we access the `changes` object when it may be null.

There are also extra members I need to add: `deaf`, `mute` and `avatar_hash`.

FiniteReality:
Fixes #1105, and adds a MemberInfo type to contain the data. I wasn't sure whether to name the type 'MemberUpdateInfo' or not, though.

### DIFF
--- a/src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/MemberInfo.cs
+++ b/src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/MemberInfo.cs
@@ -1,0 +1,18 @@
+namespace Discord.Rest
+{
+    public struct MemberInfo
+    {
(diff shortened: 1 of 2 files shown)


ghost:
I would keep it named `MemberInfo` to be consistent with the other information models (`InviteInfo`, `GuildInfo`, `ChannelInfo`).

foxbot (approved):
lgtm

PR #1108, opened 2018-07-15
feature: adjust the ratelimit for reactions

foxbot:
allows users to add reactions quickly

### DIFF
--- a/src/Discord.Net.Rest/Net/Queue/RequestQueueBucket.cs
+++ b/src/Discord.Net.Rest/Net/Queue/RequestQueueBucket.cs
@@ -252,7 +252,7 @@ namespace Discord.Net.Queue
                     resetTick = info.Reset.Value.AddSeconds(info.Lag?.TotalSeconds ?? 1.0);
 
                     if (request.Options.IsReactionBucket)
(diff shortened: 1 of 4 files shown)


src/Discord.Net.Rest/Net/Queue/RequestQueueBucket.cs
```diff
@@ -250,6 +250,10 @@ private void UpdateRateLimit(int id, RestRequest request, RateLimitInfo info, bo
                 else if (info.Reset.HasValue)
                 {
                     resetTick = info.Reset.Value.AddSeconds(info.Lag?.TotalSeconds ?? 1.0);
+
+                    if (request.Options.IsReactionBucket)
+                        resetTick = new DateTimeOffset(resetTick.Value.Ticks / 4, TimeSpan.Zero);
```
FiniteReality: Isn't this dividing the actual unix timestamp by 4? It looks like you actually mean to divide the `diff` value below by 4 instead

foxbot: `diff` isn't actually used anywhere, except for the log message right below it (you can see it goes out of scope immediately)

PR #1111, opened 2018-07-18
Example Project

do4k:
Added my personal discord bot project, includes an example of how to incorporate an extensible command framework using reflection.


Hawxy:
> includes an example of how to incorporate an extensible command framework using reflection.

You're hopefully aware that Discord.Net.Commands already exists, right? It's generally not recommended that users (especially new users that read the examples) go and write their own command framework outside of highly custom use cases. D.N.C already has a feature set appropriate for 95% of people.

brianebeling:
Hey, I'd like to give you some actual feedback. A little review, if you will.

As far as I saw you are using your own kind of dependency injector. That is a bit of unnecessary work and potentially dangerous or doesn't work. I'd recommend you using an existing dependency injection framework.

Every command currently has its own prefix. While that in itself might be a wanted feature, you should also make a dependency out of it. So you have the possibility to either use one prefix for everything or change it out individually depending on what you injected. Or if your goal was to have everything have the same prefix, then I'd localize that somewhere, maybe in your MessageReceived event?

I love having a lot of projects. But I think you overdid it a bit. The furthest I would go is having one project for your bot and one for your modules/plugins. But one for each module/plugin is a bit over the top. It's cool that you thought about the possibility to add external modules/plugins though.

`async void MethodName()` is a big no-no. Rather have it return `async Task`. I highly recommend you to read up on it. :)

Your Modules have a method "CanHandle". I'd perhaps look for a different design where you only invoke "Handle" and then move on. Maybe have it return a 'Result' or attach other MetaData to your command (hint: Discord.Net.Commands)?

There surely is a bit more. But overall I think you did a good job. While I only pointed out negatives here, I think these help the most and my own time is also spare. But don't worry, there are also positives! For example, you kept writing consistent code. Although maybe not the optimized or "correct" code, you kept with doing it rather than creating a mess. That is a very valuable skill and I hope you stick to it.

Still34:
I assume https://github.com/RogueException/Discord.Net/pull/1111/commits/059c74f5f6553096a9bca0c0468e915cd77b698f refers to your other similarly named repository? If so, I disagree with the overall structure of the files. It is simply far too complex to be a "sample" project, and not to mention how the layout is completely out of place compared to the other existing samples. That along with the points brought up above, I do not think this PR is suitable.

FiniteReality:
I agree with the other comments. It's good code, and has a consistent style, but it just doesn't follow our recommended guidelines and IMO is too cluttered. You should really be using our commands framework in an example bot, and I think most learners would prefer if the majority of code was in a single project.

PR #1116, opened 2018-08-01
Minor fixes around OnModuleBuilding

Joe4evr:
* Don't attempt instantiation of an abstract module
* Attempt associating a TypeReader in case one is registered late

### DIFF
--- a/src/Discord.Net.Commands/Builders/ModuleBuilder.cs
+++ b/src/Discord.Net.Commands/Builders/ModuleBuilder.cs
@@ -120,7 +120,7 @@ namespace Discord.Commands.Builders
             if (Name == null)
                 Name = _aliases[0];
 
-            if (TypeInfo != null)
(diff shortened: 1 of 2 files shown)


src/Discord.Net.Commands/Builders/ModuleBuilder.cs
```diff
@@ -120,7 +120,7 @@ private ModuleInfo BuildImpl(CommandService service, IServiceProvider services,
             if (Name == null)
                 Name = _aliases[0];
 
-            if (TypeInfo != null)
+            if (TypeInfo != null && !TypeInfo.IsAbstract)
```
AntiTcb, line 123: What's the reason for excluding abstract modules?

Joe4evr: CLR throws an exception.

foxbot (approved):

PR #1120, opened 2018-08-08
Added MessageBulkDeleted Event

AnotherZane:


src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
```diff
@@ -39,6 +40,13 @@ public partial class BaseSocketClient
             remove { _messageDeletedEvent.Remove(value); }
         }
         internal readonly AsyncEvent<Func<Cacheable<IMessage, ulong>, ISocketMessageChannel, Task>> _messageDeletedEvent = new AsyncEvent<Func<Cacheable<IMessage, ulong>, ISocketMessageChannel, Task>>();
+        /// <summary> Fired when multiple messages bulk deleted. </summary>
+        public event Func<List<Cacheable<IMessage, ulong>>, ISocketMessageChannel, Task> MessageBulkDeleted {
```
AntiTcb: Change the collection type to be an `IReadOnlyList<Cacheable<IMessage, ulong>>`

src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
```diff
@@ -39,6 +40,13 @@ public partial class BaseSocketClient
             remove { _messageDeletedEvent.Remove(value); }
         }
         internal readonly AsyncEvent<Func<Cacheable<IMessage, ulong>, ISocketMessageChannel, Task>> _messageDeletedEvent = new AsyncEvent<Func<Cacheable<IMessage, ulong>, ISocketMessageChannel, Task>>();
+        /// <summary> Fired when multiple messages bulk deleted. </summary>
```
AntiTcb: When multiple messages are* bulk deleted.

src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -1270,13 +1270,17 @@ private async Task ProcessMessageAsync(GatewayOpCode opCode, int? seq, string ty
                                             return;
                                         }
 
+                                        var cacheablelist = new List<Cacheable<IMessage, ulong>>();
                                         foreach (ulong id in data.Ids)
                                         {
                                             var msg = SocketChannelHelper.RemoveMessage(channel, this, id);
                                             bool isCached = msg != null;
                                             var cacheable = new Cacheable<IMessage, ulong>(msg, id, isCached, async () => await channel.GetMessageAsync(id));
-                                            await TimedInvokeAsync(_messageDeletedEvent, nameof(MessageDeleted), cacheable, channel).ConfigureAwait(false);
```
AntiTcb, line 1278: I'm not sure if we should remove this; I could see reasons for still wanting to capture single message deletes when a bulk delete occurs.

src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -1270,13 +1270,17 @@ private async Task ProcessMessageAsync(GatewayOpCode opCode, int? seq, string ty
...
+                                            cacheablelist.Add(cacheable);
                                         }
+                                        await TimedInvokeAsync(_messageBulkDeletedEvent, nameof(MessageBulkDeleted), cacheablelist, channel).ConfigureAwait(false);
```
AntiTcb: `cacheablelist.ToReadOnlyCollection()`

src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -1270,13 +1270,17 @@ private async Task ProcessMessageAsync(GatewayOpCode opCode, int? seq, string ty
                                             return;
                                         }
 
+                                        var cacheablelist = new List<Cacheable<IMessage, ulong>>();
```
FiniteReality, line 1273: This would probably be better as an immutable list IMO.

src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
FiniteReality: I would prefer a `IReadOnlyList<T>` here, personally

AntiTcb (changes requested):

FiniteReality (changes requested):
This is probably a good idea to add, but it needs a few tweaks

src/Discord.Net.WebSocket/DiscordSocketClient.cs
AnotherZane: You could still do that as it returns a list of IMessages

AnotherZane:
@FiniteReality sure, that could be done

src/Discord.Net.WebSocket/DiscordSocketClient.cs
Joe4evr: In addition, the lib's consensus is to have it be `ImmutableArray<T>`, specifically.

PR #1123, opened 2018-08-09
Add a way to invoke a command specifying optional values by name

Joe4evr:
Requires:
* 1 (one) POCO to hold argument values (Application of the attribute automatically sets `IsRemainder` for that parameter)
* All properties in the POCO are an optional parameter (Required parameters should be specified before the POCO parameter)
* When invoking, parameters need to be a single word, ending with a colon (`:`); arguments need to be either a single word or wrapped in quotes
* Supports reading into an `IEnumerable<T>` property; user's input needs to be comma-separated


src/Discord.Net.Commands/Builders/ParameterBuilder.cs
```diff
@@ -56,11 +56,27 @@ internal void SetType(Type type)
...
         {
-            var readers = Command.Module.Service.GetTypeReaders(type);
+            var commands = Command.Module.Service;
+            if (type.GetTypeInfo().GetCustomAttribute<NamedArgumentTypeAttribute>() != null)
+            {
+                IsRemainder = true;
+                var reader = commands.GetTypeReaders(type)?.FirstOrDefault().Value;
+                if (reader == null)
+                {
+                    var readerType = typeof(NamedArgumentTypeReader<>).MakeGenericType(new[] { type });
```
Joe4evr: This would throw if the provided type argument doesn't fit in the constraints. Do we want that exception to bubble up to user-code? I've leaned towards 'yes', but I dunno if we want to change it with a better message or something. #Done

Joe4evr:
Just had another thought, should we make it such that at least one property needs to be specified in order to make the typeread succeed by default? I'm thinking we could allow modifying that behavior by whether the parameter itself has a default value (`= null`).

Joe4evr:
~~Okay, so I just worked on what that would look like...~~

~~aaaaaaaaaaaaaaaaaaand it requires breaking the signature of `ReadAsync` for that to work.~~


~~Now, we're *technically* still in beta, so it's not strictly speaking a problem, but we do have a good number of people on the nightlies, so I wanted to get some input before I push it.~~

Turns out it doesn't need to break after all. Okay then!

src/Discord.Net.Commands/Attributes/NamedArgumentTypeAttribute.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+
+namespace Discord.Commands
+{
+    /// <summary>
+    /// Instructs the command system to treat command paramters of this type
```
jacksonrakena, line 6:
```suggestion
    /// Instructs the command system to treat command parameters of this type
```

test/Discord.Net.Tests/Discord.Net.Tests.csproj
```diff
@@ -3,6 +3,7 @@
     <OutputType>Exe</OutputType>
     <RootNamespace>Discord</RootNamespace>
     <TargetFramework>netcoreapp1.1</TargetFramework>
+    <DebugType>portable</DebugType>
```
jacksonrakena, line 6: Does this need to be here?

jacksonrakena (approved):

foxbot (approved):

src/Discord.Net.Commands/Builders/ParameterBuilder.cs
foxbot: seems fine to me

PR #1124, opened 2018-08-12
Allow setting custom error messages for preconditions

Joe4evr:


src/Discord.Net.Commands/Attributes/Preconditions/RequireBotPermissionAttribute.cs
```diff
@@ -52,7 +53,7 @@ public override async Task<PreconditionResult> CheckPermissionsAsync(ICommandCon
                 if (guildUser == null)
                     return PreconditionResult.FromError("Command must be used in a guild channel");
```
k-boyle: What about a second custom message for this one?

Joe4evr: Can add more properties to set, but I wasn't sure about doing that.

src/Discord.Net.Commands/Attributes/Preconditions/RequireOwnerAttribute.cs
```diff
@@ -10,17 +10,19 @@ namespace Discord.Commands
...
             {
                 case TokenType.Bot:
                     var application = await context.Client.GetApplicationInfoAsync();
                     if (context.User.Id != application.Owner.Id)
-                        return PreconditionResult.FromError("Command can only be run by the owner of the bot");
+                        return PreconditionResult.FromError(ErrorMessage ?? "Command can only be run by the owner of the bot");
                     return PreconditionResult.FromSuccess();
                 default:
-                    return PreconditionResult.FromError($"{nameof(RequireOwnerAttribute)} is not supported by this {nameof(TokenType)}.");                    
+                    return PreconditionResult.FromError($"{nameof(RequireOwnerAttribute)} is not supported by this {nameof(TokenType)}.");
```
k-boyle, line 50: Not sure if this one also needs its own custom message or not considering the other token types.

Joe4evr: I don't think we want to encourage creation of non-Bot applications.

FiniteReality (commented):
I think this would make more sense as a method which returns a string, and takes an object[] for formatting parameters. That way, we can pass in contextual data to the string too.

src/Discord.Net.Commands/Attributes/Preconditions/RequireUserPermissionAttribute.cs
```diff
@@ -11,6 +11,8 @@ public class RequireUserPermissionAttribute : PreconditionAttribute
     {
         public GuildPermission? GuildPermission { get; }
         public ChannelPermission? ChannelPermission { get; }
+        public override string ErrorMessage { get; set; }
+        public string NotAGuild { get; set; }
```
Chris-Johnston: I think that `NotAGuildMessage` or `NotAGuildErrorMessage` would be a more fitting name.

PR #1126, opened 2018-08-18
Add == support on Color

Joe4evr:


src/Discord.Net.Core/Entities/Roles/Color.cs
```diff
@@ -100,6 +100,17 @@ public Color(float r, float g, float b)
                 (uint)(b * 255.0f);
         }
 
+        public static bool operator ==(Color lhs, Color rhs)
+            => lhs.RawValue == rhs.RawValue;
+
+        public static bool operator !=(Color lhs, Color rhs)
+            => lhs.RawValue != rhs.RawValue;
```
FiniteReality, line 107: Wouldn't this make more sense to implement as `!(lhs == rhs)`?

PR #1127, opened 2018-08-19
Code Cleanup

ghost:
Normalizing code, improving performance and simplifying it


FiniteReality:
Most of this is highly opinionated formatting changes. There's no need to change them.

JohanLarsson:
@SleepyBoyy before putting in the work to do something like this it is a good idea to discuss it for example in an issue.

Still34:
Not every piece of project should be put through something like R# and "cleaned up" without prior discussion and respecting the project's contribution guide.

PR #1128, opened 2018-08-19
Add validation to bot tokens based on string length

Chris-Johnston:
Before generating the authorization header to the prefixed token, this change will check the length of bot tokens to see if they are at least 50 characters long. This is longer than the length of a bearer token or oauth client secret, so it throws an exception if the user tries to use the wrong token here.

I've checked against several of my tokens of various ages, and they all appear to be 59 characters long. [Even this example on the Discord api docs is 59 characters long.](https://discordapp.com/developers/docs/reference#authentication-example-bot-token-authorization-header) But, because this value isn't explicitly stated to always be of this length, I don't think that's a safe assumption to make.

Would it make sense to add this validation for all of the token types?

### DIFF
--- a/src/Discord.Net.Core/Utils/TokenUtils.cs
+++ b/src/Discord.Net.Core/Utils/TokenUtils.cs
@@ -1,0 +1,46 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Discord
+{
+    public static class TokenUtils
+    {
+        /// <summary>
+        ///     Checks the validity of the supplied token of a specific type.
+        /// </summary>
+        /// <param name="tokenType"> The type of token to validate. </param>
+        /// <param name="token"> The token value to validate. </param>
+        /// <exception cref="ArgumentNullException"> Thrown when the supplied token string is null, empty, or contains only whitespace.</exception>
+        /// <exception cref="ArgumentException"> Thrown when the supplied TokenType or token value is invalid. </exception>
+        public static void ValidateToken(TokenType tokenType, string token)
+        {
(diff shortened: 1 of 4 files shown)


Joe4evr:
There's no guarantee from Discord that tokens will *stay* inside some range of length, far as I know.

But if we *are* going with this, aren't client secrets way longer than the auth token?

Chris-Johnston:
> aren't client secrets way longer than the auth token?

[Going off of what I can see from the docs](https://discordapp.com/developers/docs/topics/oauth2#authorization-code-grant-access-token-exchange-example) and my own applications, client secrets look to be significantly shorter. Discord's example is 25 characters long, mine was 32. So it looks like the bot tokens are longer.

Chris-Johnston:
[There's a `validateToken` parameter of `LoginAsync` that doesn't appear to be used?](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Rest/BaseDiscordClient.cs#L53)

Seems like it would be more appropriate to handle it there.

Update: As of 86afe76c91d8851d18b8a4b678efec94083824f8 , I've moved this logic to the `BaseDiscordClient#LoginAsync` method. Because of this optional `validateToken` parameter, should the assumptions being made about Discord tokens have some breaking change, setting this value to `false` will get around this validation.

Update 2: As of 04199b2bf802078d372838b5f24081f3cc640370 , invalid tokens will log a warning if `validateToken` is set to `true`, but will not stop any calls to the API (where it should later throw an exception).

Joe4evr:
LGTM, I guess.

AntiTcb:
Was brought up earlier in the channel, it might be better to log a warning to the user and let the API throw the error.

FiniteReality (approved):

PR #1129, opened 2018-08-20
Make SocketVoiceChannel.ConnectAsync arguments optional.

AntiTcb:


FiniteReality (commented):
It would make sense to make sure we've provided default parameters wherever we can, since the interface states they are optional.

AntiTcb:
While not a compile time error, a warning is logged when you add default values to parameters in an explicit interface implementation.

As long as the interface defines the arguments as optional, then explicit implementations *technically* follow that. As it currently stands this is technically valid without also making the explicit interface implemenation's arguments also have default values:
```cs
IVoiceChannel foo = new SocketGroupChannel();
foo.ConnectAsync();
```
<https://stackoverflow.com/questions/4922714/why-are-c-sharp-4-optional-parameters-defined-on-interface-not-enforced-on-imple>
<https://stackoverflow.com/questions/5683111/warning-from-explicitly-implementing-an-interface-with-optional-parameters>

PR #1130, opened 2018-08-22
Add suffix support

JustNrik:
I'm sure some people would like this library to support suffixes so I suggest these changes to implement it.

### DIFF
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -348,6 +348,21 @@ namespace Discord.Commands
 
         public Task<IResult> ExecuteAsync(ICommandContext context, int argPos, IServiceProvider services, MultiMatchHandling multiMatchHandling = MultiMatchHandling.Exception)
             => ExecuteAsync(context, context.Message.Content.Substring(argPos), services, multiMatchHandling);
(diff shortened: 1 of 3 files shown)


FiniteReality:
I dislike the idea of supporting suffixes. It makes reasoning about a command harder (e.g. "is this part of the command, or is it just the suffix?") and doesn't really have any valid use-cases.

Furthermore, the way this is implemented is not useful at all: why didn't you just copy the existing prefixes implementation but tweak it for suffixes?

EDIT: Just because some people may want it doesn't mean the library should support it. I'm sure some people want Discord.Net to support user tokens again, but that's not allowed by Discord so we're not supporting it.

JustNrik:
Current 'work-around' to perform successful suffixed commands just make code to look uglier, anyway no one forces you to use suffixes and this doesn't affect how prefixed commands work, so no one needs to make the slightest change in their code if this is implemented. I'm suggesting this for that minority of users who would like to execute their commands with suffixes.

For example:

Suffix = " pls"
Commands:
ban @someuser pls
kick @someuser pls
mute @someuser pls
say Hello, world! pls
createchannel Test pls
fun randomcat pls

It's all about how the bot owner decides to use it and explains to.

PR #1131, opened 2018-08-22
TimeSpan parser

XtremeOwnageDotCom:
This change will greatly increase the success rate for the TimeSpan parser to actually work, by leveraging the built-in TimeSpan.Parse, with no additional parameters.

### DIFF
--- a/src/Discord.Net.Commands/Readers/TimeSpanTypeReader.cs
+++ b/src/Discord.Net.Commands/Readers/TimeSpanTypeReader.cs
@@ -34,13 +34,12 @@ namespace Discord.Commands
             //First, try using the built-in TimeSpan.Parse. (Default Implementation)
             if (TimeSpan.TryParseExact(data, _formats, CultureInfo.InvariantCulture, out var timeSpan))
                 return Task.FromResult(TypeReaderResult.FromSuccess(timeSpan));
-            //@ericmck2000 - Try using the regular TimeSpan.TryPrase. Greatly improves success rate.
-            else if (TimeSpan.TryParse(data, out var time))
-                return Task.FromResult(TypeReaderResult.FromSuccess(time));
+            //Try using the regular TimeSpan.TryPrase. Greatly improves success rate.
+            else if (TimeSpan.TryParse(data, out timeSpan))
+                return Task.FromResult(TypeReaderResult.FromSuccess(timeSpan));
+            //Else, return parse failed error.
             else
-            {
                 return Task.FromResult(TypeReaderResult.FromError(CommandError.ParseFailed, "Failed to parse TimeSpan"));
-            }
         }
     }
 }
--- a/src/Discord.Net.Commands/Readers/TimeSpanTypeReader.cs
+++ b/src/Discord.Net.Commands/Readers/TimeSpanTypeReader.cs
@@ -25,11 +25,22 @@ namespace Discord.Commands
             "%s's'",                //      1s
         };
 
+
         public override Task<TypeReaderResult> ReadAsync(ICommandContext context, string input, IServiceProvider services)
         {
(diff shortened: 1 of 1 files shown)


XtremeOwnageDotCom:
Related Issues:

https://github.com/RogueException/Discord.Net/issues/817

src/Discord.Net.Commands/Readers/TimeSpanTypeReader.cs
```diff
@@ -25,11 +25,22 @@ internal class TimeSpanTypeReader : TypeReader
...
-                ? Task.FromResult(TypeReaderResult.FromSuccess(timeSpan))
-                : Task.FromResult(TypeReaderResult.FromError(CommandError.ParseFailed, "Failed to parse TimeSpan"));
+            //Store as variable, to prevent multiple calls to this method.
+            string data = input.ToLowerInvariant();
+
+            //First, try using the built-in TimeSpan.Parse. (Default Implementation)
+            if (TimeSpan.TryParseExact(data, _formats, CultureInfo.InvariantCulture, out TimeSpan timeSpan))
+                return Task.FromResult(TypeReaderResult.FromSuccess(timeSpan));
+            //@ericmck2000 - Try using the regular TimeSpan.TryPrase. Greatly improves success rate.
+            else if (TimeSpan.TryParse(data, out TimeSpan Time))
```
FiniteReality: Use `out var time` instead

FiniteReality (changes requested):

src/Discord.Net.Commands/Readers/TimeSpanTypeReader.cs
XtremeOwnageDotCom: Updated.

src/Discord.Net.Commands/Readers/TimeSpanTypeReader.cs
```diff
@@ -25,11 +25,22 @@ internal class TimeSpanTypeReader : TypeReader
...
-                ? Task.FromResult(TypeReaderResult.FromSuccess(timeSpan))
-                : Task.FromResult(TypeReaderResult.FromError(CommandError.ParseFailed, "Failed to parse TimeSpan"));
+            //Store as variable, to prevent multiple calls to this method.
+            string data = input.ToLowerInvariant();
+
+            //First, try using the built-in TimeSpan.Parse. (Default Implementation)
+            if (TimeSpan.TryParseExact(data, _formats, CultureInfo.InvariantCulture, out var timeSpan))
+                return Task.FromResult(TypeReaderResult.FromSuccess(timeSpan));
+            //@ericmck2000 - Try using the regular TimeSpan.TryPrase. Greatly improves success rate.
+            else if (TimeSpan.TryParse(data, out var time))
```
k-boyle: Forgive me if I'm wrong, but couldn't you just do, `out timeSpan` because of the previous `out`?

XtremeOwnageDotCom: You are absolutely correct. I didn't think about it. Updated.

FiniteReality:
I've had more thought about this, and I think it would be better in the long term to add additional formats to the `_formats` array. For example, TimeSpan.TryParse uses this format string: `[ws][-]{ d | d.hh:mm[:ss[.ff]] | hh:mm[:ss[.ff]] }[ws]`, which we could easily replicate if we wrote a parser ourselves.
In the long term, I think a span-based typereader for reading dates and times would actually be superior to using TimeSpan.TryParse or TryParseExact

XtremeOwnageDotCom:
As another method, There is an issue over at Humanizer to basically De-Humanize a string into a timespan too.

https://github.com/Humanizr/Humanizer/issues/691

If that issue gets traction, can piggy-back the efforts.

XtremeOwnageDotCom:
Closing this.

PR #1132, opened 2018-08-23
Add AddBlankField to embed + Allow empty name/values.

XtremeOwnageDotCom:
Updated Field builder, to allow for "Empty" fields and values.

If you send "_ _" via the API, it will properly display as a blank field. I have tested this locally, and are currently utilizing it in production.

As well,
I Updated the Value setting to use String.IsNullOrWhiteSpace instead of String.IsNullOrEmpty, because passing a string containing nothing but spaces through the API will cause errors as well. However, "_ _" works perfectly.

### DIFF
--- a/src/Discord.Net.Core/Entities/Messages/EmbedBuilder.cs
+++ b/src/Discord.Net.Core/Entities/Messages/EmbedBuilder.cs
@@ -196,6 +196,17 @@ namespace Discord
             AddField(field);
             return this;
         }
+        
+        public EmbedBuilder AddBlankField(bool inline = false)
+        {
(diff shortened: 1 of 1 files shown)


XtremeOwnageDotCom:
To note, there is an alternative method to using "_ _".

I reviewed the source code from discord.js @ https://github.com/discordjs/discord.js/blob/master/src/structures/MessageEmbed.js

They utilize '\u200B' instead.

FiniteReality:
It's not in our best interest to support this in the library; it relies on a behaviour which could easily change in the near future without warning. If you need this, it would be best to make an extension method.

PR #1134, opened 2018-08-27
Update CommandAttribute.cs

JustNrik:
Nullable<T> is not valid for Attributes, this is my suggested fix.


src/Discord.Net.Commands/Attributes/CommandAttribute.cs
```diff
@@ -7,7 +7,7 @@ public class CommandAttribute : Attribute
     {
         public string Text { get; }
         public RunMode RunMode { get; set; } = RunMode.Default;
-        public bool? IgnoreExtraArgs { get; set; }
+        public bool? IgnoreExtraArgs { get; private set; }
```
Joe4evr: If it's only set via constructor, then the setter should be removed entirely.

Joe4evr (changes requested):

PR #1139, opened 2018-08-31
Fix swapped parameters in ArgumentException, ArgumentNullException cstrs

Chris-Johnston:
The constructors of `ArgumentException` and `ArgumentNullException` are inconsistent with each other, which sometimes results in their parameters being swapped. I've looked for all occurrences of this and fixed their constructors. Named parameters would have worked too.

For reference, here's the constructor definitions I'm referring to:
```csharp
public ArgumentNullException(string paramName, string message);
public ArgumentException(string message, string paramName);
```

### DIFF
--- a/src/Discord.Net.Core/Entities/Emotes/Emote.cs
+++ b/src/Discord.Net.Core/Entities/Emotes/Emote.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Globalization;
 
 namespace Discord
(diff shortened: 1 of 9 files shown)


Joe4evr:
I think we should probably go for named arguments regardless, just so that it's clear which parameter they're associated with down the line.

Chris-Johnston:
My only reservation with this is that to be consistent with the rest of the codebase, we should then also update all of the 70+ other constructors of `ArgumentException` and the few of `ArgumentNullException`, since they don't use named parameters.

AntiTcb:
Make it so, number one.

Chris-Johnston:
Added named parameters to (what I believe to be) all of the `ArgumentException` and `ArgumentNullException` constructors. Any other types come to mind that could named parameters?

foxbot (approved):

PR #1140, opened 2018-08-31
Add invite maxAge check before attempting to create the invite

Still34:
# Summary

The maximum age of an invite at the moment cannot exceed `86400` (1 day); this PR adds a check and throws prior to the creation of the invite should the `maxAge` exceed such value to prevent wasting a request and have it return HTTP400.

This PR also improves the maxAge and maxUses param by swapping `if else` with simple null coalescence operator.


src/Discord.Net.Rest/Entities/Channels/ChannelHelper.cs
```diff
@@ -76,15 +76,15 @@ internal static class ChannelHelper
...
-            if (maxAge.HasValue)
-                args.MaxAge = maxAge.Value;
-            else
-                args.MaxAge = 0;
-            if (maxUses.HasValue)
-                args.MaxUses = maxUses.Value;
-            else
-                args.MaxUses = 0;
+            if (maxAge.HasValue && maxAge > 86400)
+                throw new InvalidOperationException("The maximum age of an invite must be less than or equal to a day (86400).");
```
FiniteReality: IMO this would be better as an ArgumentOutOfRangeException

FiniteReality (changes requested):

src/Discord.Net.Rest/Entities/Channels/ChannelHelper.cs
Still34: Good idea, changed.

src/Discord.Net.Rest/Entities/Channels/ChannelHelper.cs
```diff
@@ -76,15 +76,16 @@ internal static class ChannelHelper
...
-            var args = new CreateChannelInviteParams { IsTemporary = isTemporary, IsUnique = isUnique };
-            if (maxAge.HasValue)
-                args.MaxAge = maxAge.Value;
-            else
-                args.MaxAge = 0;
-            if (maxUses.HasValue)
-                args.MaxUses = maxUses.Value;
-            else
-                args.MaxUses = 0;
+            if (maxAge.HasValue && maxAge > 86400)
```
Joe4evr: I'm pretty sure you don't have to check `.HasValue` since the lifting of the operator already does that.

FiniteReality: Just checked, it does work as expected: https://sharplab.io/#v2:D4AQDABCCMDcCwAoEBmKAmCBhCBvJEhUaIALBALIAUAlHgUYwJYB2ALgPwQAeEAvBDAJEjRg1EQmAMwhVeAPggBWGuIlEYATioAiNgFMAzmx01hjAL5ILQA=

Joe4evr: I'm surprised the compiler doesn't put the `HasValue` check first there. :eyes:

src/Discord.Net.Rest/Entities/Channels/ChannelHelper.cs
```diff
@@ -76,15 +76,16 @@ internal static class ChannelHelper
...
-            if (maxAge.HasValue)
-                args.MaxAge = maxAge.Value;
-            else
-                args.MaxAge = 0;
-            if (maxUses.HasValue)
-                args.MaxUses = maxUses.Value;
-            else
-                args.MaxUses = 0;
+            if (maxAge > 86400)
+                throw new ArgumentOutOfRangeException(nameof(maxAge), maxAge,
```
Chris-Johnston: Could possibly use named parameters here?

src/Discord.Net.Rest/Entities/Channels/ChannelHelper.cs
```diff
@@ -76,15 +76,16 @@ internal static class ChannelHelper
...
+                    "The maximum age of an invite must be less than or equal to a day (86400).");
```
Chris-Johnston: What do you think about the wording:
"... less than or equal to a day's length in seconds (86400)."

So that it can include units?

FiniteReality: I'd say `less than or equal to a day (86400 seconds)`

Chris-Johnston: I think that reads better.

Chris-Johnston (approved):

src/Discord.Net.Rest/Entities/Channels/ChannelHelper.cs
```diff
@@ -76,15 +76,16 @@ internal static class ChannelHelper
...
-            var args = new CreateChannelInviteParams { IsTemporary = isTemporary, IsUnique = isUnique };
-            if (maxAge.HasValue)
-                args.MaxAge = maxAge.Value;
-            else
-                args.MaxAge = 0;
-            if (maxUses.HasValue)
-                args.MaxUses = maxUses.Value;
-            else
-                args.MaxUses = 0;
+            if (maxAge > 86400)
```
foxbot: per consistency, usually these checks are made in the API client using the static `Preconditions` class; see https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Rest/DiscordRestApiClient.cs#L327

foxbot (approved):

FiniteReality (approved):

PR #1143, opened 2018-09-05
Added commands to sample bot

patrykr5:


JustNrik:
Have you even tested if these commands actually work? For instance, unban command.

samples/02_commands_framework/Modules/PublicModule.cs
```diff
@@ -49,6 +49,32 @@ public async Task BanUserAsync(IGuildUser user, [Remainder] string reason = null
...
         }
 
+        // Unban a user
+        [Command("unban")]
+        [RequireContext(ContextType.Guild)]
+        // make sure the user invoking the coomand can unban
+        [RequireUserPermission(GuildPermission.BanMembers)]
+        // make sure the bot itself can unban
+        [RequireBotPermission(GuildPermission.BanMembers)]
+        public async Task UnbanUserAsync(IGuildUser user)
```
FiniteReality: This won't work - a user not in the guild cannot have an IGuildUser instance associated with them.

samples/02_commands_framework/Modules/PublicModule.cs
```diff
@@ -59,5 +85,18 @@ public Task EchoAsync([Remainder] string text)
         [Command("list")]
         public Task ListAsync(params string[] objects)
             => ReplyAsync("You listed: " + string.Join("; ", objects));
+
+        // Can be used to example put information
+        [Command("info")]
+        public async Task InformationAsync()
```
FiniteReality: IMO this command is rather pointless

FiniteReality (changes requested):

samples/02_commands_framework/Modules/PublicModule.cs
patrykr5: Right, this is pointless command - only i wanted show how to use embedBuilder; now i created embedBuilder in kick command.

patrykr5:
I would like to add :
- fixed command unban
- warn user and after crossing limit - mute user. 
I'll do it tomorrow or day after tomorrow, please don't close this pull request.

samples/02_commands_framework/Modules/PublicModule.cs
```diff
@@ -49,6 +50,29 @@ public async Task BanUserAsync(IGuildUser user, [Remainder] string reason = null
             await ReplyAsync("ok!");
         }
 
+        // Kick a user
+        [Command("kick")]
```
Still34, line 54: Adding only one sample command does not seem to really contribute to the existing project in a helpful manner; perhaps add more or something more helpful?

samples/02_commands_framework/Modules/PublicModule.cs
```diff
@@ -49,6 +50,29 @@ public async Task BanUserAsync(IGuildUser user, [Remainder] string reason = null
...
+        public async Task KickUserAsync(IGuildUser user, [Remainder] string reason = null)
+        {
+            // Creates the pattern TextBlock
+            EmbedBuilder builder = new EmbedBuilder();
+
+            builder.WithTitle(user.Username + " kicked from server!")
+                .WithColor(Color.Orange)
+                .WithDescription("by " +
+                                $"{Context.User.Username}" +
+                                Environment.NewLine +
```
Still34, line 69: What's with the strange string formatting choice? Interpolating and concat?

samples/02_commands_framework/Modules/PublicModule.cs
```diff
@@ -49,6 +50,29 @@ public async Task BanUserAsync(IGuildUser user, [Remainder] string reason = null
...
+                                "Reason: " + reason);
+
+            await user.KickAsync(reason, null);
```
Still34, line 72: `RequestOptions` is already `null` by default. If anything, show off how a `RequestOptions` can be used.

Still34 (changes requested):

samples/02_commands_framework/Modules/PublicModule.cs
FiniteReality: Instead of using Environment.Newline you should just use \n instead.

This whole thing should have probably been written as:
```
$"By {context.User.Username}\nReason: {reason}"
```

samples/02_commands_framework/Modules/PublicModule.cs
FiniteReality: RequestOptions is only really useful for setting audit log reasons, which is provided by the reason param anyway, IIRC.

PR #1145, opened 2018-09-10
Fix for incomplete structure sample

Still34:
# Summary

This short PR fixes the incomplete sample that was written for previous pre-release version(s) of Discord.Net or coding errors that would not allow the sample to compile; this fix also suggests people use env vars instead of hardcoding the token.

### DIFF
--- a/docs/guides/getting_started/samples/intro/structure.cs
+++ b/docs/guides/getting_started/samples/intro/structure.cs
@@ -36,7 +36,7 @@ class Program
             // you must set the MessageCacheSize. You may adjust the number as needed.
             //MessageCacheSize = 50,
 
-            // If your platform doesn't have native websockets,
+            // If your platform doesn't have native WebSockets,
             // add Discord.Net.Providers.WS4Net from NuGet,
             // add the `using` at the top, and uncomment this line:
             //WebSocketProvider = WS4NetProvider.Instance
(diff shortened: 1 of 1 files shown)


docs/guides/getting_started/samples/intro/structure.cs
```diff
@@ -36,7 +36,7 @@ private Program()
             // you must set the MessageCacheSize. You may adjust the number as needed.
             //MessageCacheSize = 50,
 
-            // If your platform doesn't have native websockets,
+            // If your platform doesn't have native WebSockets,
```
FiniteReality, line 39: IMO there was no need for this change

FiniteReality (approved):

docs/guides/getting_started/samples/intro/structure.cs
Still34: I'm actually not sure how that got there; I didn't explicitly make that change in my IDE 🤔

foxbot: this is a good change and I appreciate that you fixed a grammar error in a PR which has meaningful code changes instead of allowing someone to snipe it as a hacktoberfest PR, thanks :)

foxbot (approved):

PR #1159, opened 2018-09-30
Add SyncPermissionsAsync to Sync Child Channels with its Parent

### LINKED ISSUE #1144: No way to change the category of a channel
There is currently no way I'm aware of to modify the `parent_id` property of a channel, which is needed to change its category.

Still34:
# Summary

This resolves #1144 and attempts to improve from the previous attempt of #1016. Instead of forcefully implementing a property that does not exist on the API, this creates a new method called `SyncPermissionsAsync` under all `INestedChannel`.

### DIFF
--- a/src/Discord.Net.Rest/API/Common/Overwrite.cs
+++ b/src/Discord.Net.Rest/API/Common/Overwrite.cs
@@ -13,13 +13,5 @@ namespace Discord.API
         public ulong Deny { get; set; }
         [JsonProperty("allow"), Int53]
         public ulong Allow { get; set; }
-
-        public Overwrite(ulong targetId, PermissionTarget targetType, ulong allowValue, ulong denyValue)
-        {
-            TargetId = targetId;
-            TargetType = targetType;
-            Allow = allowValue;
-            Deny = denyValue;
-        }
     }
 }
--- a/src/Discord.Net.Rest/API/Common/Overwrite.cs
+++ b/src/Discord.Net.Rest/API/Common/Overwrite.cs
@@ -1,4 +1,4 @@
-﻿#pragma warning disable CS1591
+#pragma warning disable CS1591
 using Newtonsoft.Json;
 
 namespace Discord.API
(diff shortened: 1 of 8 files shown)


src/Discord.Net.Rest/Entities/Channels/ChannelHelper.cs
```diff
@@ -326,6 +326,20 @@ public static async Task<ICategoryChannel> GetCategoryAsync(INestedChannel chann
             var model = await client.ApiClient.GetChannelAsync(channel.CategoryId.Value, options).ConfigureAwait(false);
             return RestCategoryChannel.Create(client, model) as ICategoryChannel;
         }
+        public static async Task SyncPermissionsAsync(INestedChannel channel, BaseDiscordClient client, RequestOptions options)
+        {
+            var category = await GetCategoryAsync(channel, client, options).ConfigureAwait(false);
+            if (category == null) return;
```
foxbot: This should probably throw an `InvalidOperation` or something; I'm more keen on throwing when unexpected behavior happens than just dropping the error. No-op errors like this can be a lot harder to debug.

foxbot (changes requested):

src/Discord.Net.Rest/API/Common/Overwrite.cs
```diff
@@ -13,5 +13,13 @@ internal class Overwrite
         public ulong Deny { get; set; }
         [JsonProperty("allow"), Int53]
         public ulong Allow { get; set; }
+
+        public Overwrite(ulong targetId, PermissionTarget targetType, ulong allowValue, ulong denyValue)
```
FiniteReality: IMO this is unnecessary; these values can easily be set using brace initialization:
```cs
new Overwrite
{
   TargetId = 1234,
    // ...
    DenyValue = 1234
}
```

FiniteReality (changes requested):

PR #1163, opened 2018-09-30
fix: strip trailing slash from ratelimit bucket IDs

### LINKED ISSUE #1125: Rate limits when using RemoveReactionAsync and RemoveAllReactionsAsync
Using version 2.0.0-beta, I am sometimes receiving a "Rate limit triggered" warning when using RemoveReactionAsync and RemoveAllReactionsAsync. [This](https://hastebin.com/ocepuqanow.cs) command (`!reaction 1`) yields these results:

```
[Debug - 8/13/2018 5:54:30 PM] Script (Reaction): Creating message
[Debug - 8/13/2018 5:54:31 PM] Script (Reaction): Starting iteration 1
[Debug - 8/13/2018 5:54:31 PM] Script (Reaction): Adding reaction
[Debug - 8/13/2018 5:54:31 PM] Script (Reaction): Removing reaction
[Warning - 8/13/2018 5:54:31 PM] Rest: Rate limit triggered: channels/(id)/messages/reactions/
[Debug - 8/13/2018 5:54:31 PM] Script (Reaction): Adding reaction
[Debug - 8/13/2018 5:54:32 PM] Script (Reaction): Removing all reactions
[Warning - 8/13/2018 5:54:32 PM] Rest: Rate limit triggered: channels/(id)/messages/reactions
[Debug - 8/13/2018 5:54:33 PM] Script (Reaction): Finished iteration 1
[Debug - 8/13/2018 5:54:33 PM] Script (Reaction): Deleting message
```

On other tests of the same command, it only triggered the rate limit on RemoveAllReactionsAsync. Most tests of it result in at least one rate limit warning.

foxbot:
This resolves #1125.

fixes a bug where some ratelimit buckets would include a trailing slash,
while others wouldn't; this would cause them to be treated as separate
ratelimits, even though they are the same

ideally this fix should change the ratelimit generator, but that code is
pretty complicated and this was an easier fix that seems less likely to
break things in the future.

tested against normal bot function, all routes are assigned the proper
buckets from my testing, so this should be good to go.

### DIFF
--- a/src/Discord.Net.Rest/DiscordRestApiClient.cs
+++ b/src/Discord.Net.Rest/DiscordRestApiClient.cs
@@ -1415,6 +1415,9 @@ namespace Discord.API
                 }
 
                 format = builder.ToString();
+                if (format.LastIndexOf('/') == format.Length-1)
+                    format = format.Substring(0, format.Length - 1); // perf: change the code above so this isn't necessary
+
                 return x => string.Format(format, x.ToArray());
             }
             catch (Exception ex)
--- a/src/Discord.Net.Rest/DiscordRestApiClient.cs
+++ b/src/Discord.Net.Rest/DiscordRestApiClient.cs
@@ -1,3 +1,4 @@
+
 #pragma warning disable CS1591
 using Discord.API.Rest;
 using Discord.Net;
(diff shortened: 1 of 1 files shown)


src/Discord.Net.Rest/DiscordRestApiClient.cs
```diff
@@ -1415,6 +1415,9 @@ private static string GetBucketId(BucketIds ids, Expression<Func<string>> endpoi
                 }
 
                 format = builder.ToString();
+                if (format.LastIndexOf('/') == format.Length-1)
```
Chris-Johnston: How about checking like this? Might be (insignificantly) faster.

```suggestion
                if (format.Length > 0 && format[format.Length - 1] == '/')
```

Joe4evr: Better alternative is performing the above check on the *builder* and then doing `builder.Remove(builder.Length - 1, 1)` to avoid the intermediate string allocation.

PR #1166, opened 2018-10-01
Add GetVoiceRegionsAsync to IGuild.

natekford:
Already mostly implemented from [DiscordRestAPIClient](https://github.com/RogueException/Discord.Net/blob/ff0fea98a65d907fbce07856f1a9ef4aebb9108b/src/Discord.Net.Rest/DiscordRestApiClient.cs#L1228), but there was no way to publicly access it.

This is more useful than [IDiscordClient#GetVoiceRegionsAsync](https://github.com/RogueException/Discord.Net/blob/ff0fea98a65d907fbce07856f1a9ef4aebb9108b/src/Discord.Net.Core/IDiscordClient.cs#L245) because it will return the VIP voice channels if the guild is partnered.


foxbot (approved):
Looks good, thanks for your help!

PR #1167, opened 2018-10-02
fix: don't assume the member will always be included on MESSAGE_CREATE

### LINKED ISSUE #1153: System.NullReferenceException at AddOrUpdateUser(GuildMember model)
Bot just gave off this message and stack trace makes this seem like an internal library issue. Not sure what caused it or even remotely know how it can be reproduced.

> 21:39:58 Gateway     Error handling Dispatch (MESSAGE_CREATE):
   at Discord.WebSocket.SocketGuild.AddOrUpdateUser(GuildMember model)
   at Discord.WebSocket.DiscordSocketClient.<ProcessMessageAsync>d__107.MoveNext()

Posting here in case you can take a quick gander and pinpoint where it's stemming from. This has only occurred twice in the time that I've been using this version of Discord.Net, and both times happening within minutes of each other. I will be keeping an eye on it in the coming hours / days to see if it occurs again.

foxbot:
This resolves #1153.

Member objects are only included on a message when the user has
transitioned from an offline state to an online state (i think?), so
this change will fall back to the prior behavior, where we just create
an incomplete member object for these states.

### DIFF
--- a/src/Discord.Net.WebSocket/DiscordSocketClient.cs
+++ b/src/Discord.Net.WebSocket/DiscordSocketClient.cs
@@ -1159,10 +1159,9 @@ namespace Discord.WebSocket
                                         {
                                             if (guild != null)
                                             {
-                                                if (data.Member.IsSpecified) // member isn't always specified...
-                                                    author = guild.AddOrUpdateUser(data.Member.Value); //per g250k, we can create an entire member now
-                                                else
-                                                    author = guild.AddOrUpdateUser(data.Author.Value); // user has no guild-specific data
+                                                author = data.Member.IsSpecified // member isn't always included, but use it when we can
+                                                    ? guild.AddOrUpdateUser(data.Member.Value)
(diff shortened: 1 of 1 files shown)


src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -1158,7 +1158,12 @@ private async Task ProcessMessageAsync(GatewayOpCode opCode, int? seq, string ty
                                         if (author == null)
                                         {
                                             if (guild != null)
-                                                author = guild.AddOrUpdateUser(data.Member.Value); //per g250k, we can create an entire member now
+                                            {
+                                                if (data.Member.IsSpecified) // member isn't always specified...
+                                                    author = guild.AddOrUpdateUser(data.Member.Value); //per g250k, we can create an entire member now
+                                                else
+                                                    author = guild.AddOrUpdateUser(data.Author.Value); // user has no guild-specific data
```
Joe4evr: Wouldn't it be nicer to use a ternary here?

Issue #1168, opened 2018-10-05
High CPU Usage on playing music [with fix]


InkyQuill:
There are a lot of bots whose icons start flashing in voice chat on second music track. This flashing triggers a high CPU usage on web-based Discord clients.

Seems that the problem lies in not disposing underlying AudioStream when OpusEncoderStream is disposed.

Here's a little patch to fix this:
https://github.com/InkyQuill/Discord.Net/commit/e7061bff413c2ae7c8ff9327af23ac69856d08cd.patch

Joe4evr:
Sounds reasonable. Have you verified this fix yourself first? If so, you can submit a PR, though the comment isn't necessary.

FiniteReality:
I'm looking through the project with IDisposable analyzers now. I'll create a PR when I have more work ready.

ComputerMaster1st:
Hi There,

Are you using "CreatePCMStream" per chance? If so, are you using that command each time you start up a new song to play?

If you're answer is yes, then that explains the high CPU usage. I've encountered this long ago and only creating the stream once solved this problem. Constantly recreating the stream does also affect the browser version of Discord by making it appear like heavy stutter.

For example...

```
foreach (song in queue) {
    // create new pcm stream
    // stream song to discord
}
```
^^ That causes high CPU & stutter within the browser from my experience.

THis is how I solved it.
```
using (var discordStream = audioClient.CreatePCMStream())
{
    foreach (song in queue) {
        // stream song to discord
        // Flush the pcm stream
    }
}
```

As a reminder, this is from my own experience with the "High CPU Usage" & the stutter in the browser.

InkyQuill:
I've analyzed the merged PR, looks like my fix wasn't implemented, and underlying AudioStream still isn't disposed. So, even with a big Dispose PR clients are still suffering from high CPU load and blinking voice icon when bot is playing music. The issue is not closed.

InkyQuill:
@ComputerMaster1st I've used your approach. It is fine if you have a small number of songs to play, but our guild has a big playlist and playback can be several hours long. And Discord API is not so stable, so sometimes UDP connection will be broken and it can't restart. Yep, after that the only choice is to restart the bot for music to play. That's why I chose a one PCM stream per song approach. And the high load/stuttering is gone with a one line fix to OpusEncoder Dispose section.

PR #1169, opened 2018-10-05
Fix for Audio CPU Usage

InkyQuill:
PR for #1168


FiniteReality:
I'm going to look into this further and see if there's anywhere else we're not disposing of things we should be. Is that okay?

InkyQuill:
Sure, there's always something that could be optimized. Take your time.
Also, I've tried disposing other underlying streams in BufferedWriteStream, RTPWriteStream, SodiumEncryptStream. It doesn't help much, nothing changes. But still, those are not disposed correctly and left for GC.

FiniteReality:
Closing since I have a work-in-progress branch with IDisposable analyzers being ran on it.

Issue #1084, opened 2018-06-05
CancellationTokenSource memory leak


Twistplay:
Discord.Net is not calling Dispose after using CancellationTokenSource.CreateLinkedTokenSource.    According to MSDN, this is necessary.  See remarks in:

[https://msdn.microsoft.com/en-us/library/system.threading.cancellationtokensource(v=vs.110).aspx](url)

This therefore creates a memory leak on each web call, which becomes quite significant over time.

See DefaultRestClient.SendInternalAsync for example.   It calls:

`cancelToken = CancellationTokenSource.CreateLinkedTokenSource(_cancelToken, cancelToken).Token;`

But provides no means for the caller to dispose of the CancellationTokenSource that is created.   Note that it is the source that needs to be disposed, not just the token itself.

Auralytical:
Did a profiler show that this was actually memory leaking?

While I agree that disposing it would be the correct pattern, I'd be surprised if the GC wasn't cleaning up linked tokens.

Twistplay:
It did come from the Visual Studio memory profiler, yes.   Taking memory snapshots from diagnostics periodically and using it's diff'ing capabilities clearly showed a continuous growth in the number of CancellationTokenSource objects and associated structures.   This was in proportion to the number of calls made to the API.

My understanding is that they do not get automatically GC'ed according to everything I've read online, hence the need to Dispose.

PR #1171, opened 2018-10-05
Implement Dispose for types which have disposable data

### LINKED ISSUE #1168: High CPU Usage on playing music [with fix]

FiniteReality:
This is likely a **large breaking change** for some people, so should be merged before 2.0 is released.

I still need to handle Dispose warnings for these:
- [x] Discord.Net.Websocket/Entities/SocketGuild
- [x] Discord.Net.Tests

None of this has been thoroughly tested, either. I'm just opening the PR to gather feedback on things I might be missing as I work on the PR.

This should fix #1168 and #1084.


src/Discord.Net.WebSocket/DiscordShardedClient.cs
```diff
@@ -369,5 +375,20 @@ Task<IReadOnlyCollection<IVoiceRegion>> IDiscordClient.GetVoiceRegionsAsync(Requ
         /// <inheritdoc />
         Task<IVoiceRegion> IDiscordClient.GetVoiceRegionAsync(string id, RequestOptions options)
             => Task.FromResult<IVoiceRegion>(GetVoiceRegion(id));
+
+        internal override void Dispose(bool disposing)
```
AntiTcb, line 379: Should this go through each shard and dispose those as well?

FiniteReality: Yeah, it should. I missed another thing :sweat_smile:

src/Discord.Net.WebSocket/DiscordShardedClient.cs
```diff
@@ -369,5 +375,22 @@ Task<IReadOnlyCollection<IVoiceRegion>> IDiscordClient.GetVoiceRegionsAsync(Requ
...
```
FiniteReality, line 379: Would it be a good idea to call StopAsync in Dispose here? We do it in other places (e.g. DiscordSocketClient)

AntiTcb: Sharded client's stop async just calls stop async for each shard, so if their disposers call StopAsync, it's enough just to dispose them.

samples/01_basic_ping_bot/Program.cs
```diff
@@ -14,23 +14,31 @@ namespace _01_basic_ping_bot
     // - Here, under the 02_commands_framework sample
     // - https://github.com/foxbot/DiscordBotBase - a bare-bones bot template
     // - https://github.com/foxbot/patek - a more feature-filled bot, utilizing more aspects of the library
-    class Program
+    class Program : IDisposable
```
brianebeling: Is that really necessary? On termination of the program that is already taken care of.

samples/01_basic_ping_bot/Program.cs
```diff
@@ -14,23 +14,31 @@ namespace _01_basic_ping_bot
...
     {
-        private DiscordSocketClient _client;
+        private readonly DiscordSocketClient _client;
 
         // Discord.Net heavily utilizes TAP for async, so we create
         // an asynchronous context from the beginning.
         static void Main(string[] args)
-            => new Program().MainAsync().GetAwaiter().GetResult();
+        {
+            using (var program = new Program())
```
brianebeling: Then we also would not need this anymore and could go back to an async Main method or get rid of the using at least.

samples/01_basic_ping_bot/Program.cs
```diff
@@ -14,23 +14,31 @@ namespace _01_basic_ping_bot
...
+                program.MainAsync().GetAwaiter().GetResult();
+        }
 
-        public async Task MainAsync()
+        public Program()
```
RethoughtSolutions: In addition I would advice against doing this kind of work in the constructor. Despite it causing confusion or rather being "untypically". Complex work in constructors should be avoided because of awkward behaviour when throwing exceptions. That is exactly why creational patterns exist. 
In this case there (probably) will never be an exception thrown. But for readability reasons I would just move that into MainAsync and be done with it.

brianebeling, line 28: In addition I would advice against doing this kind of work in the constructor. Despite it causing confusion or rather being "untypically". Complex work in constructors should be avoided because of awkward behaviour when throwing exceptions. That is exactly why creational patterns exist. 
In this case there (probably) will never be an exception thrown. But for readability reasons I would just move that into MainAsync and be done with it.

samples/01_basic_ping_bot/Program.cs
FiniteReality: This is simply not true. For example, if you use a Timer and do not dispose it, your application may not close properly if it is ran using `dotnet run`.

samples/01_basic_ping_bot/Program.cs
FiniteReality: The socket client should be disposed to prevent memory leaks.

samples/01_basic_ping_bot/Program.cs
FiniteReality: As far as I'm aware, lots of initializer work has been done in constructors, so this is fine. And if the constructor throws (which I doubt it will), we're likely in a state where we cannot logically continue anyway, so for a sample, this is fine.

FiniteReality:
Okay, so I've finished the first pass I believe. This should grab everything that had disposable data properly, but I would like some help testing it

FiniteReality:
This should be more or less ready to review now. cc @foxbot

foxbot (approved):

PR #1175, opened 2018-10-13
Added check to ignore bot commands.

Rithari:
Added the usual check to ignore bot commands in this sample CommandHandler as seen and approved in #1174.

### DIFF
--- a/docs/guides/commands/samples/post-execution/command_executed_demo.cs
+++ b/docs/guides/commands/samples/post-execution/command_executed_demo.cs
@@ -27,7 +27,7 @@ public async Task HandleCommandAsync(SocketMessage msg)
     var message = messageParam as SocketUserMessage;
     if (message == null) return;
     int argPos = 0;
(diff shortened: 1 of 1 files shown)


Herdo (commented):
In the sample, some additional comments would be great, explaining the conditions in the `if`-statement.

Still34:
@Herdo that would fit more in line with the tutorial snippet of `Introduction to Command Service` instead of the `Post-execution Handling` snippet, which is what this PR changes.

Rithari:
Hi, this being a more advanced snippet treating something different than the initial set up, I think that the suggested comments are redundant. 

Sincerely, 
Rithari

Herdo (approved):

Still34:
With that being said, adding comments to the files mentioned would still be appreciated, as that was a valid suggestion - just on a different file.

Rithari:
I'm pretty sure there were comments added on my behalf as well - should there be none I'll do a PR.

PR #1183, opened 2018-10-25
Add Add Guild Member Endpoint

AntiTcb:
Could use some assistance with testing this one if anyone already has an Oauth workflow setup.

Nitpick for later, we should probably revise our naming scheme for GuildUser to be GuildMember in all cases to be consistent with Discord.

### DIFF
--- a/src/Discord.Net.Rest/Entities/Guilds/GuildHelper.cs
+++ b/src/Discord.Net.Rest/Entities/Guilds/GuildHelper.cs
@@ -264,8 +264,14 @@ namespace Discord.Rest
             func?.Invoke(args);
 
             if (args.Roles.IsSpecified)
-                args.RoleIds = Optional.Create(args.Roles.Value.Select(r => r.Id));
+            {
+                var ids = args.Roles.Value.Select(r => r.Id);
 
+                if (args.RoleIds.IsSpecified)
+                    args.RoleIds.Value.Concat(args.Roles.Value.Select(r => r.Id));
+                else
+                    args.RoleIds = Optional.Create(args.Roles.Value.Select(r => r.Id));
+            }
             var apiArgs = new AddGuildMemberParams
             {
                 AccessToken = accessToken,
--- a/src/Discord.Net.Rest/Entities/Guilds/GuildHelper.cs
+++ b/src/Discord.Net.Rest/Entities/Guilds/GuildHelper.cs
@@ -257,6 +257,28 @@ namespace Discord.Rest
         }
 
         //Users
+        public static async Task<RestGuildUser> AddGuildUserAsync(IGuild guild, BaseDiscordClient client, ulong userId, string accessToken,
+            Action<AddGuildUserProperties> func, RequestOptions options)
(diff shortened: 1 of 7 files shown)


Chris-Johnston (approved):
After testing with the following steps, these changes work for me:

1. Use the OAuth2 URL generator on a bot application, with the `guilds.join` scope.
2. POST the code in the querystring, [as defined in steps here](https://discordapp.com/developers/docs/topics/oauth2#authorization-code-grant)
3. Have a test bot create a new guild, then call `AddGuildUserAsync` with the token from step 2. `AddGuildUserProperties` properties all appear to work as expected\*, including roles.

~~\*I'm not sure if Deafen/Mute should do anything if that's a client-side feature? Toggling these values didn't appear to make a difference.~~ EDIT: These work as expected, and can be seen when actually joining a voice channel, and not from the corner of the client.

AntiTcb:
It's not necessary on the library's end to catch that. The client's the one that should handle that case.

src/Discord.Net.Rest/Entities/Guilds/GuildHelper.cs
```diff
@@ -257,6 +257,28 @@ public static async Task<RestBan> GetBanAsync(IGuild guild, BaseDiscordClient cl
...
 
         //Users
+        public static async Task<RestGuildUser> AddGuildUserAsync(IGuild guild, BaseDiscordClient client, ulong userId, string accessToken,
+            Action<AddGuildUserProperties> func, RequestOptions options)
+        {
+            var args = new AddGuildUserProperties();
+            func?.Invoke(args);
+
+            if (args.Roles.IsSpecified)
+                args.RoleIds = Optional.Create(args.Roles.Value.Select(r => r.Id));
```
foxbot: this should either append to the RoleIds collection, or throw if the RoleIds is defined. otherwise, the user could specify both RoleIds and Roles, and would lose half of their request.

foxbot (changes requested):

src/Discord.Net.Rest/Entities/Guilds/GuildHelper.cs
```diff
@@ -257,6 +257,34 @@ public static async Task<RestBan> GetBanAsync(IGuild guild, BaseDiscordClient cl
...
         //Users
+        public static async Task<RestGuildUser> AddGuildUserAsync(IGuild guild, BaseDiscordClient client, ulong userId, string accessToken,
+            Action<AddGuildUserProperties> func, RequestOptions options)
+        {
+            var args = new AddGuildUserProperties();
+            func?.Invoke(args);
+
+            if (args.Roles.IsSpecified)
+            {
+                var ids = args.Roles.Value.Select(r => r.Id);
```
Chris-Johnston, line 268: `ids` isn't used here?

Chris-Johnston (changes requested):

Chris-Johnston (approved):

foxbot (approved):

PR #1192, opened 2018-11-08
Fixed CommandExecuted firing twice for failed RuntimeResults

k-boyle:
The issue was because of [line 247](https://github.com/RogueException/Discord.Net/blob/dev/src/Discord.Net.Commands/Info/CommandInfo.cs#L247) of CommandInfo also triggering the event.
As far as I can tell this fixes the specific issue. Not too sure of any unintended consequences of removing these two lines, but from my testing, I don't see any but would appreciate others testing.

### DIFF
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -603,6 +603,8 @@ namespace Discord.Commands
             //If we get this far, at least one parse was successful. Execute the most likely overload.
             var chosenOverload = successfulParses[0];
             var result = await chosenOverload.Key.ExecuteAsync(context, chosenOverload.Value, services).ConfigureAwait(false);
+            if (!result.IsSuccess && result is ExecuteResult)
+                await _commandExecutedEvent.InvokeAsync(chosenOverload.Key.Command, context, result);
             return result;
         }
     }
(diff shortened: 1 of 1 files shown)


k-boyle:
Just realised this would cause an issue with Line 181 of CommandInfo, will look for a better solution to the issue later.

Kuromu (commented):
@TheCasino would it be a "simple" solution to check if the result is an `ExecuteResult` as well, instead of just omitting the line(s) entirely? it would keep `RuntimeResult`s from being sent out twice, while still allowing for the `ExecuteResult`s to still make their way through? A shoddy mockup, but:
```cs
if (!result.IsSuccess && result is ExecuteResult _)
    await _commandExecutedEvent.InvokeAsync(chosenOverload.Key.Command, context, result);
```

k-boyle:
Yeah, that's exactly what I was thinking. Will push changes shortly.

src/Discord.Net.Commands/CommandService.cs
```diff
@@ -603,7 +603,7 @@ float CalculateScore(CommandMatch match, ParseResult parseResult)
             //If we get this far, at least one parse was successful. Execute the most likely overload.
             var chosenOverload = successfulParses[0];
             var result = await chosenOverload.Key.ExecuteAsync(context, chosenOverload.Value, services).ConfigureAwait(false);
-            if (!result.IsSuccess) // succesful results raise the event in CommandInfo#ExecuteInternalAsync (have to raise it there b/c deffered execution)
```
foxbot, line 606: retain the comment here, and invert this so that we ensure the result is _not_ a `RuntimeResult`, since there are other failed results that the command can return prior to execution.

foxbot (changes requested):

foxbot (approved):

PR #1194, opened 2018-11-13
Move Name property from IChannel to IGuildChannel

AntiTcb:


AntiTcb:
And apparently Group channels can have names and I missed that entirely. >_>

PR #1198, opened 2018-11-27
feature: add DiscordSocketRestClient

### LINKED ISSUE #803: The SocketClient should allow you to force an HTTP request
Currently, the only way to obtain an uncached member is by calling `SocketGuild#DownloadUsersAsync()`.

The current implementation should probably be modified to mirror how cached vs uncached messages are handled. 
https://github.com/RogueException/Discord.Net/blob/3c1e76615e1ef37cf3deb9aa9ab3e3b2f2064e41/src/Discord.Net.WebSocket/Entities/Channels/SocketTextChannel.cs#L52-L60

foxbot:
this resolves #803.

Users can access a DiscordSocketRestClient from the new
`DiscordSocketClient.Rest` property.

DiscordSocketRestClient is a wrapper over DiscordRestClient with certain
state-modifying methods, such as Login/Logout disabled, to prevent users
from breaking the client state.

DiscordSocketRestClient uses the same API client as the
DiscordSocketClient, allowing for shared ratelimiting - meaning users
can now force HTTP requests without needing to wory about running into
429s.


Joe4evr:
We should probably also modify some methods that return entities (primarily `GetUserAsync`) to make use of this?

PR #1199, opened 2018-11-27
List of many changes

ComputerMaster1st:
Hi There,

Please note, the fork I use is a few commits behind and does contain merge commits due to possible mistakes/failures and I've only been using the fork for bot operations.

Changes made...

Solved the AudioClient Disconnect Deadlock << I believe this is already pulled but not sure.
Solves the ObjectDisposedException w/ UdpClient as "Unobserved Task Handler" kept picking them up.
Fixed the "Disconnecting" locking bug. Symptoms of this was 100% core usage meaning it was stuck in an infinite loop.
Added msg.content null check as the msg may have only an embed.

For reasons unknown, all this works and has been running on a live bot for some time. Please accept my apologies for the many mess ups (which I believe are many).

Regards,

CM1


FiniteReality:
Split your changes off (use cherry picking) to make it easier to review.

PR #1200, opened 2018-11-28
(Part 1 of 4) Added Msg.Content Null Check For Prefixes

ComputerMaster1st:
Hi There,

This PR solves a NullRef. exception when checking for prefixes as some messages may only contain an embed.

Tested & Non-Breaking Change.

Regards,

CM1

### DIFF
--- a/src/Discord.Net.Commands/Extensions/MessageExtensions.cs
+++ b/src/Discord.Net.Commands/Extensions/MessageExtensions.cs
@@ -19,7 +19,7 @@ namespace Discord.Commands
         public static bool HasCharPrefix(this IUserMessage msg, char c, ref int argPos)
         {
             var text = msg.Content;
-            if (!string.IsNullOrEmpty(text) && text.Length > 0 && text[0] == c)
+            if (!string.IsNullOrEmpty(text) && text[0] == c)
             {
                 argPos = 1;
                 return true;
@@ -45,7 +45,7 @@ namespace Discord.Commands
         public static bool HasMentionPrefix(this IUserMessage msg, IUser user, ref int argPos)
         {
             var text = msg.Content;
-            if (!string.IsNullOrEmpty(text) && text.Length <= 3 || text[0] != '<' || text[1] != '@') return false;
+            if ((!string.IsNullOrEmpty(text) && text.Length <= 3) || text[0] != '<' || text[1] != '@') return false;
 
             int endPos = text.IndexOf('>');
             if (endPos == -1) return false;
--- a/src/Discord.Net.Commands/Extensions/MessageExtensions.cs
+++ b/src/Discord.Net.Commands/Extensions/MessageExtensions.cs
@@ -19,7 +19,7 @@ namespace Discord.Commands
         public static bool HasCharPrefix(this IUserMessage msg, char c, ref int argPos)
         {
             var text = msg.Content;
-            if (text.Length > 0 && text[0] == c)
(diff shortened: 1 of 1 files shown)


src/Discord.Net.Commands/Extensions/MessageExtensions.cs
```diff
@@ -19,7 +19,7 @@ public static class MessageExtensions
         public static bool HasCharPrefix(this IUserMessage msg, char c, ref int argPos)
         {
             var text = msg.Content;
-            if (text.Length > 0 && text[0] == c)
+            if (!string.IsNullOrEmpty(text) && text.Length > 0 && text[0] == c)
```
Chris-Johnston: `string.IsNullOrEmpty` makes the `text.Length > 0` check redundant

src/Discord.Net.Commands/Extensions/MessageExtensions.cs
```diff
@@ -45,7 +45,7 @@ public static bool HasStringPrefix(this IUserMessage msg, string str, ref int ar
         public static bool HasMentionPrefix(this IUserMessage msg, IUser user, ref int argPos)
         {
             var text = msg.Content;
-            if (text.Length <= 3 || text[0] != '<' || text[1] != '@') return false;
+            if (!string.IsNullOrEmpty(text) && text.Length <= 3 || text[0] != '<' || text[1] != '@') return false;
```
Chris-Johnston: These parameters should be grouped together. **EDIT: This is incorrect, see latest suggestion.**
```suggestion
            if (!string.IsNullOrEmpty(text) && (text.Length <= 3 || text[0] != '<' || text[1] != '@')) return false;
```

Chris-Johnston (changes requested):

src/Discord.Net.Commands/Extensions/MessageExtensions.cs
ComputerMaster1st: Removed length check.

src/Discord.Net.Commands/Extensions/MessageExtensions.cs
ComputerMaster1st: Done.

src/Discord.Net.Commands/Extensions/MessageExtensions.cs
```diff
@@ -45,7 +45,7 @@ public static bool HasStringPrefix(this IUserMessage msg, string str, ref int ar
         public static bool HasMentionPrefix(this IUserMessage msg, IUser user, ref int argPos)
         {
             var text = msg.Content;
-            if (text.Length <= 3 || text[0] != '<' || text[1] != '@') return false;
+            if (!string.IsNullOrEmpty(text) && (text.Length <= 3 || text[0] != '<' || text[1] != '@')) return false;
```
Chris-Johnston: My mistake, the logic for this is still incorrect. This returns `false` if any of these conditions are true, which is unlike the others, so `!string.IsNullOrEmpty(text)` isn't correct.
```suggestion
            if (string.IsNullOrEmpty(text) || text.Length <= 3 || text[0] != '<' || text[1] != '@') return false;
```

Chris-Johnston (changes requested):

src/Discord.Net.Commands/Extensions/MessageExtensions.cs
ComputerMaster1st: Change Made.

Chris-Johnston (approved):

foxbot (approved):

PR #1203, opened 2018-11-28
(Part 4 of 4) Solves AudioClient Lockup On Disconnect

ComputerMaster1st:
Hi There,

This PR solves the AudioClient disconnect lockup. This issue caused the audio client to not always disconnect properly and indirectly led to memory leaks (objects not being disposed correctly). In other words, the "_task" variable was waiting on itself to complete which caused the lockup.

The "disconnected" event was moved up as it made logical sense to do so. That way, any further faults could be found a little bit better (in my mind at least).

Tested & Non-Breaking.

Regards,

CM1

### DIFF
--- a/src/Discord.Net.WebSocket/ConnectionManager.cs
+++ b/src/Discord.Net.WebSocket/ConnectionManager.cs
@@ -157,9 +157,9 @@ namespace Discord
 
             await _onDisconnecting(ex).ConfigureAwait(false);
 
-            await _logger.InfoAsync("Disconnected").ConfigureAwait(false);
-            State = ConnectionState.Disconnected;
             await _disconnectedEvent.InvokeAsync(ex, isReconnecting).ConfigureAwait(false);
+            State = ConnectionState.Disconnected;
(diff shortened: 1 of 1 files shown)


src/Discord.Net.WebSocket/ConnectionManager.cs
```diff
@@ -103,12 +103,10 @@ public virtual async Task StartAsync()
                 finally { _stateLock.Release(); }
             });
         }
-        public virtual async Task StopAsync()
+        public virtual Task StopAsync()
         {
             Cancel();
-            var task = _task;
```
Joe4evr, line 109: What happens to `_task` now?

ComputerMaster1st: It'll complete as normal without blocking. Refer to https://github.com/RogueException/Discord.Net/pull/1095#issue-196925154 for better explanation.

Joe4evr: Gotcha.

PR #1206, opened 2018-12-01
Improve validation of Bot Tokens

Chris-Johnston:
Per feedback from @Emzi0767 received right before #1204 was merged, this PR improves the methods used to validate Bot Tokens.

In addition to checking that tokens satisfy a minimum length requirement, this method will also try to decode the user id from the first part of the Bot Token. Test cases have been updated as well.

Exception messages have been updated to include a hint to check if the supplied Bot Token is an OAuth client secret.

### DIFF
--- a/src/Discord.Net.Core/Utils/TokenUtils.cs
+++ b/src/Discord.Net.Core/Utils/TokenUtils.cs
@@ -28,6 +28,9 @@ namespace Discord
         /// </returns>
         internal static bool CheckBotTokenValidity(string message)
         {
+            if (string.IsNullOrWhiteSpace(message))
+                return false;
+
             // split each component of the JWT
             var segments = message.Split('.');
 
--- a/src/Discord.Net.Core/Utils/TokenUtils.cs
+++ b/src/Discord.Net.Core/Utils/TokenUtils.cs
@@ -18,6 +18,40 @@ namespace Discord
         internal const int MinBotTokenLength = 58;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Discord.Net.Core/Utils/TokenUtils.cs
```diff
@@ -16,6 +16,48 @@ public static class TokenUtils
...
+
+            // ensure that there are three parts
+            if (segments.Length != 3)
+                return false;
+
+            try
+            {
+                // decode the first segment as base64
+                var v = Convert.FromBase64String(segments[0]);
+                BitConverter.ToUInt64(v, 0);
```
Emzi0767: You misinterpreted what I said, again. The base64 contents result in a **string representation of the ulong**, i.e. it decodes to a string.

BitConverter is for converting binary representations to respective types (so if it decoded to 8 bytes, that'd be the ulong you'd need to convert).

What you actually need to do is something akin to `Encoding.ASCII.GetString(v)` and run the result of that through `ulong.Parse()` or `ulong.TryParse()`.

Emzi0767 (changes requested):

src/Discord.Net.Core/Utils/TokenUtils.cs
Chris-Johnston: After checking against one of my own tokens (id = 428477944009195520), I was able to fix this result. The examples from Discord docs looked correct, despite following the wrong procedure. Fixed now.

foxbot (approved):

PR #1209, opened 2018-12-02
Fix after message remaining in MessageUpdated if message author wasn't in the payload

### LINKED ISSUE #1208: MessageUpdated sometimes returns a null SocketMessage if MessageCacheSize == 0
I was seeing an issue where `MessageUpdated` was occasionally giving me a null arg2 (the `SocketMessage`), and was able to reproduce it (on the latest version) by sending a message with a URL for which Discord generates an embed automatically - such as https://github.com/RogueException/Discord.Net - but, notably, only when `MessageCacheSize` is `0`; MessageUpdated's arg2 will never be null if the cache is enabled.

This seems to due to the fact that the `MESSAGE_UPDATE` gateway event, if updating a message to add an embed, doesn't contain author data - and the check on `DiscordSocketClient.cs#1221` will only create a "detached" `SocketMessage` if the author is present. The `SocketMessage` passed to the event on `DiscordSocketClient.cs#1212`, `after`, remains null - as without `MessageCacheSize` being set, the call to `channel.GetCachedMessage` will always return null.

According to the documentation (and messages on DAPI), the presence of `MessageCacheSize` should only affect arg1 of `MessageUpdated`, but it seems that it indirectly affects both. I'm not sure what the best solution is in this case, but perhaps the `"MESSAGE_UPDATE"` case could be modified to return a SocketMessage with missing properties (or a less derived type that matches better). This would at least match the functionality described in Discord's API docs: 

> Unlike creates, message updates may contain only a subset of the full message object payload (but will always contain an id and channel_id). 
https://discordapp.com/developers/docs/topics/gateway#message-update

Of course, the solution for now is just "Set `MessageCacheSize`", but I still think the event should return, at minimum, what the Discord API does (message and channel IDs).

AntiTcb:
Resolves #1208

### DIFF
--- a/src/Discord.Net.WebSocket/DiscordSocketClient.cs
+++ b/src/Discord.Net.WebSocket/DiscordSocketClient.cs
@@ -1233,7 +1233,7 @@ namespace Discord.WebSocket
                                             }
                                             else
                                                 // Message author wasn't specified in the payload, so create a completely anonymous unknown user
-                                                author = new SocketUnknownUser(this, 0);
+                                                author = new SocketUnknownUser(this, id: 0);
 
                                             after = SocketMessage.Create(this, State, author, channel, data);
                                         }
--- a/src/Discord.Net.WebSocket/DiscordSocketClient.cs
+++ b/src/Discord.Net.WebSocket/DiscordSocketClient.cs
@@ -1218,16 +1218,22 @@ namespace Discord.WebSocket
                                             cachedMsg.Update(State, data);
                                             after = cachedMsg;
                                         }
-                                        else if (data.Author.IsSpecified)
+                                        else
                                         {
                                             //Edited message isnt in cache, create a detached one
                                             SocketUser author;
-                                            if (guild != null)
-                                                author = guild.GetUser(data.Author.Value.Id);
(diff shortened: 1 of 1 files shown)


src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -1218,16 +1218,22 @@ private async Task ProcessMessageAsync(GatewayOpCode opCode, int? seq, string ty
...
+                                                    author = (channel as SocketChannel).GetUser(data.Author.Value.Id);
+                                                if (author == null)
+                                                    author = SocketUnknownUser.Create(this, State, data.Author.Value);
+                                            }
                                             else
-                                                author = (channel as SocketChannel).GetUser(data.Author.Value.Id);
-                                            if (author == null)
-                                                author = SocketUnknownUser.Create(this, State, data.Author.Value);
+                                                // Message author wasn't specified in the payload, so create a completely anonymous unknown user
+                                                author = new SocketUnknownUser(this, 0);
```
moiph: nitpick: named params are nice when values like this are passed in

moiph (approved):

PR #1214, opened 2018-12-15
Fix #1213, Add GuildEmote Creator Property

### LINKED ISSUE #1213: Add Creator property to (Guild?)Emote
Discord's API returns [the user who created an emote as part of the emote payload](https://discordapp.com/developers/docs/resources/emoji#emoji-object). We should give users the ability to access that if possible.

Proposed addition:
```cs
public partial class GuildEmote
{
    public Optional<IUser> Creator { get; }
}
```

*Update:* Adjust to feedback.

Chris-Johnston:
Fix #1213 . 

Updates the API Model for Emoji to include a User, [from the API specifications](https://discordapp.com/developers/docs/resources/emoji#emoji-object).

Adds the Creator property to GuildEmote, which is optional. This property is not guaranteed to be set by the API whenever Emotes are used. For example, the Guild Create event does not supply the creators of emotes, but explicitly `GET`ting the Emote does.

```cs
private async Task Client_GuildAvailable(SocketGuild arg)
{
    foreach (var e in arg.Emotes)
    {
        // e is supplied as part of the Guild Create payload, Creator will typically **not** be supplied
        // <:lgtm:523410039948443649> by
        Console.WriteLine($"{e} by {e.Creator}");
        // when explicitly GETting the emote, the Creator is supplied
        var x = await arg.GetEmoteAsync(e.Id);
        // 01:14:05 Rest        GET guilds/428476681519497218/emojis/523410039948443649: 511.47 ms
        // <:lgtm:523410039948443649> by ChrisJ#8703
        var c = x.Creator.IsSpecified ? await x.Creator.Value.GetOrDownloadAsync() : null;
        Console.WriteLine($"{x} by {(object)c ?? "null"}");
    }
}
```

Because an IUser entity is used as part of this model and not a User Id, modifications to the `ToEntity` extension method were made, and all references of this method were updated.

### DIFF
--- a/src/Discord.Net.Core/Entities/Emotes/GuildEmote.cs
+++ b/src/Discord.Net.Core/Entities/Emotes/GuildEmote.cs
@@ -34,7 +34,7 @@ namespace Discord
         ///     Gets the user Id that created this emoji.
         /// </summary>
         /// <returns>
-        ///     A User Id who created this emoji, which may be null.
+        ///     A user Id of the user who created this emoji, which may be null. A null value only indicates that the creator was not supplied as part of the API response.
         /// </returns>
         public ulong? CreatorId { get; }
 
--- a/src/Discord.Net.Core/Entities/Emotes/GuildEmote.cs
+++ b/src/Discord.Net.Core/Entities/Emotes/GuildEmote.cs
@@ -30,12 +30,20 @@ namespace Discord
         ///     A read-only list containing snowflake identifiers for roles that are allowed to use this emoji.
         /// </returns>
         public IReadOnlyList<ulong> RoleIds { get; }
+        /// <summary>
+        ///     Gets the user Id that created this emoji.
+        /// </summary>
+        /// <returns>
+        ///     A User Id who created this emoji, which may be null.
+        /// </returns>
+        public ulong? CreatorId { get; }
 
-        internal GuildEmote(ulong id, string name, bool animated, bool isManaged, bool requireColons, IReadOnlyList<ulong> roleIds) : base(id, name, animated)
+        internal GuildEmote(ulong id, string name, bool animated, bool isManaged, bool requireColons, IReadOnlyList<ulong> roleIds, ulong? userId) : base(id, name, animated)
         {
             IsManaged = isManaged;
             RequireColons = requireColons;
             RoleIds = roleIds;
+            CreatorId = userId;
         }
 
         private string DebuggerDisplay => $"{Name} ({Id})";
--- a/src/Discord.Net.Core/Entities/Emotes/GuildEmote.cs
+++ b/src/Discord.Net.Core/Entities/Emotes/GuildEmote.cs
@@ -31,19 +31,19 @@ namespace Discord
         /// </returns>
         public IReadOnlyList<ulong> RoleIds { get; }
         /// <summary>
-        ///     Gets the user Id that created this emoji.
+        ///     Gets the User that created this emoji.
         /// </summary>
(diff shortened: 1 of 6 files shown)


src/Discord.Net.Rest/Extensions/EntityExtensions.cs
```diff
@@ -1,13 +1,20 @@
...
         {
-            return new GuildEmote(model.Id.Value, model.Name, model.Animated.GetValueOrDefault(), model.Managed, model.RequireColons, ImmutableArray.Create(model.Roles));
+            return new GuildEmote(model.Id.Value, model.Name, model.Animated.GetValueOrDefault(), model.Managed, model.RequireColons, ImmutableArray.Create(model.Roles), GetEmoteAuthor(model, guild));
+        }
+
+        internal static Optional<IUser> GetEmoteAuthor(API.Emoji model, IGuild guild)
+        {
+            if (!model.User.IsSpecified || guild == null)
+                return new Optional<IUser>();
+            return new Optional<IUser>(guild.GetUserAsync(model.User.Value.Id).Result);
```
foxbot: this seems unideal, I would rather we use a `Cacheable` here than force a request in a non-asynchronous context.

src/Discord.Net.Core/Entities/Emotes/GuildEmote.cs
```diff
@@ -30,12 +30,20 @@ public class GuildEmote : Emote
         ///     A read-only list containing snowflake identifiers for roles that are allowed to use this emoji.
         /// </returns>
         public IReadOnlyList<ulong> RoleIds { get; }
+        /// <summary>
+        ///     Gets the User that created this emoji.
+        /// </summary>
+        /// <returns>
+        ///     An optional <see cref="IUser"/> who created this emoji. An unspecified value only indicates that the creator was not supplied as part of the API response.
+        /// </returns>        
+        public Optional<IUser> Creator { get; }
```
foxbot: as per below comment, drop a Cacheable here?

foxbot (changes requested):

src/Discord.Net.Rest/Extensions/EntityExtensions.cs
```diff
@@ -1,13 +1,26 @@
...
+using System.Collections.Immutable;
 using System.Linq;
+using System.Threading.Tasks;
 
 namespace Discord.Rest
 {
     internal static class EntityExtensions
     {
-        public static GuildEmote ToEntity(this API.Emoji model)
+        public static GuildEmote ToEntity(this API.Emoji model, IGuild guild)
```
Joe4evr, line 9: Since `ToEntity()` is internal and all the public callsites are already async (I think), it doesn't seem like a breaking change to make this async as well.

Chris-Johnston: https://github.com/Chris-Johnston/Discord.Net/blob/322be8a31a6bb67f2d6451a25a73634ce7fe11fa/src/Discord.Net.WebSocket/Entities/Guilds/SocketGuild.cs#L340

https://github.com/Chris-Johnston/Discord.Net/blob/322be8a31a6bb67f2d6451a25a73634ce7fe11fa/src/Discord.Net.WebSocket/Entities/Guilds/SocketGuild.cs#L412

I've found these two usages that aren't within an async function. Given that I'd have to update all occurrences if I made `ToEntity` async anyways, maybe for these two cases I could use the `Result` property.

I'd like to get your opinion on this.

PR #1219, opened 2018-12-23
Add Discord.Net logo

brianebeling:
This PR "solves" https://github.com/RogueException/Discord.Net/issues/1006

I wasn't quite sure what structure you imagined, so I just did it like I usually do. Feel free to remove any of the resolutions you would deem as unnecessary. I would recommend at least including the SVG, because that one is scaleable and very lightweight in size. Technically you could remove all others and have people resize the SVG as needed, in case they want a bitmap with a fixed resolution.


Chris-Johnston:
I think we definitely should keep the `.ai` file around, since that was the original source. It might be useful to keep a copy of the pngs, maybe zipped, just for ease of use?

PR #1231, opened 2019-01-04
Allow the user to define guild icon's format and size

sudo-gab:
The default ``IconUrl`` property returns the image in a jpg format with no ability to define the size. This PR is a simple change following ``IUser``'s ``GetAvatarUrl`` method, adding the ability to define both the format and the size.

### DIFF
--- a/src/Discord.Net.Core/CDN.cs
+++ b/src/Discord.Net.Core/CDN.cs
@@ -56,10 +56,13 @@ namespace Discord
         /// <returns>
         ///     A URL pointing to the guild's icon in the specified size.
         /// </returns>
-        public static string GetGuildIconUrl(ulong guildId, string iconId, ushort size, ImageFormat format)
+        public static string GetGuildIconUrl(ulong guildId, string iconId, ImageFormat format, ushort size)
         {
             if (string.IsNullOrWhiteSpace(iconId))
                 return null;
+            if (format == ImageFormat.Gif)
+                throw new ArgumentException("Requested image format mustn't be a gif.");
+
             string extension = FormatToExtension(format, iconId);
             return $"{DiscordConfig.CDNUrl}icons/{guildId}/{iconId}.{extension}?size={size}";
         }
--- a/src/Discord.Net.Core/CDN.cs
+++ b/src/Discord.Net.Core/CDN.cs
@@ -58,7 +58,7 @@ namespace Discord
         /// </returns>
         public static string GetGuildIconUrl(ulong guildId, string iconId, ushort size, ImageFormat format)
         {
-            if (iconId == null)
+            if (string.IsNullOrWhiteSpace(iconId))
                 return null;
             string extension = FormatToExtension(format, iconId);
             return $"{DiscordConfig.CDNUrl}icons/{guildId}/{iconId}.{extension}?size={size}";
--- a/src/Discord.Net.Core/CDN.cs
+++ b/src/Discord.Net.Core/CDN.cs
@@ -51,11 +51,18 @@ namespace Discord
         /// </summary>
         /// <param name="guildId">The guild snowflake identifier.</param>
         /// <param name="iconId">The icon identifier.</param>
+        /// <param name="size">The size of the image to return in. This can be any power of two between 16 and 2048.</param>
+        /// <param name="format">The format to return.</param>
         /// <returns>
-        ///     A URL pointing to the guild's icon.
+        ///     A URL pointing to the guild's icon in the specified size.
         /// </returns>
-        public static string GetGuildIconUrl(ulong guildId, string iconId)
-            => iconId != null ? $"{DiscordConfig.CDNUrl}icons/{guildId}/{iconId}.jpg" : null;
+        public static string GetGuildIconUrl(ulong guildId, string iconId, ushort size, ImageFormat format)
+        {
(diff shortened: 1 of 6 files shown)


src/Discord.Net.Core/CDN.cs
```diff
@@ -51,11 +51,18 @@ public static string GetDefaultUserAvatarUrl(ushort discriminator)
...
+        /// <param name="size">The size of the image to return in. This can be any power of two between 16 and 2048.</param>
+        /// <param name="format">The format to return.</param>
         /// <returns>
-        ///     A URL pointing to the guild's icon.
+        ///     A URL pointing to the guild's icon in the specified size.
         /// </returns>
-        public static string GetGuildIconUrl(ulong guildId, string iconId)
-            => iconId != null ? $"{DiscordConfig.CDNUrl}icons/{guildId}/{iconId}.jpg" : null;
+        public static string GetGuildIconUrl(ulong guildId, string iconId, ushort size, ImageFormat format)
+        {
```
Chris-Johnston, line 60: We should check that size is a power of two ([this looks like a good solution to do so](https://stackoverflow.com/a/600306)), and that it's in the range [16-2048].

src/Discord.Net.Core/CDN.cs
```diff
@@ -51,11 +51,18 @@ public static string GetDefaultUserAvatarUrl(ushort discriminator)
...
+            if (iconId == null)
+                return null;
+            string extension = FormatToExtension(format, iconId);
```
Chris-Johnston, line 66: I seem to get a HTTP 415 error when using the gif format, so we may have to verify that the format is not a gif.

Chris-Johnston (changes requested):
Looks like the build is breaking for unrelated reasons because of the version string.

src/Discord.Net.Core/CDN.cs
```diff
@@ -51,11 +51,18 @@ public static string GetDefaultUserAvatarUrl(ushort discriminator)
...
         /// <param name="iconId">The icon identifier.</param>
+        /// <param name="size">The size of the image to return in. This can be any power of two between 16 and 2048.</param>
+        /// <param name="format">The format to return.</param>
         /// <returns>
-        ///     A URL pointing to the guild's icon.
+        ///     A URL pointing to the guild's icon in the specified size.
         /// </returns>
-        public static string GetGuildIconUrl(ulong guildId, string iconId)
-            => iconId != null ? $"{DiscordConfig.CDNUrl}icons/{guildId}/{iconId}.jpg" : null;
+        public static string GetGuildIconUrl(ulong guildId, string iconId, ushort size, ImageFormat format)
```
Joe4evr: > `ushort size, ImageFormat format`

All your other changes put `ImageFormat` before the `ushort`; it'd be better to be consistent with parameter ordering here.

src/Discord.Net.Rest/Entities/Guilds/RestGuild.cs
```diff
@@ -59,7 +59,8 @@ public class RestGuild : RestEntity<ulong>, IGuild, IUpdateable
         [Obsolete("DefaultChannelId is deprecated, use GetDefaultChannelAsync")]
         public ulong DefaultChannelId => Id;
         /// <inheritdoc />
-        public string IconUrl => CDN.GetGuildIconUrl(Id, IconId);
```
Joe4evr, line 62: It's probably better to keep the property and have it simply call the new method (same in the other cases).

src/Discord.Net.Core/CDN.cs
```diff
@@ -51,11 +51,18 @@ public static string GetDefaultUserAvatarUrl(ushort discriminator)
...
+        {
+            if (iconId == null)
```
Chris-Johnston:
```suggestion
            if (string.IsNullOrWhiteSpace(iconId))
```

I don't think allowing whitespace strings makes sense either

Joe4evr (changes requested):
Can do with a couple tweaks.

src/Discord.Net.Core/CDN.cs
Joe4evr: This *may* be better represented as an enum:
```cs
public enum ImageSize : ushort
{
    Sixteen = 16,
    ThirtyTwo = 32,
    SixtyFour = 64,
    OneTwentyEight = 128,
    TwoFiftySix = 256,
    FiveTwelve = 512,
    TenTwentyFour = 1024,
    TwentyFourtyEight = 2048
}
```

src/Discord.Net.Rest/Entities/Guilds/RestGuild.cs
sudo-gab: I feel like it would be a naming inconsistency, since it's a method that takes optional parameters.

src/Discord.Net.Core/CDN.cs
sudo-gab: I'm waiting on more opinions on this. To me, joe's enum solution seems a bit better. Anyways if such change would have been done, we would have to change the arguments we take for both the user's avatar url and the guild's icon url.

sudo-gab:
Closing this PR for a bigger one, since I want to adjust a few things other than guild icons in the CDN file.

PR #1234, opened 2019-01-09
CDN Rehaul

sudo-gab:
This PR rehauls the CDN class. Contains changes from #1231 
Current changes:

 - All CDN methods have been updated to take a format and a size argument for endpoints which allow to  do so.
 - Methods for endpoints which do not support a certain format now throw an ``ArgumentException``  (tested).
 - Every method that takes a size argument now enforces a range between 16 and 2048 and for it to be a power of 2.
 - Arguments have been changed from ``uint, ImageFormat`` to ``ImageFormat, uint`` for fluid consistency.
 - Every entity using a method from the CDN is now prefixed with ``Get`` and takes an ``ImageFormat`` and ``uint`` to suit CDN's changes.

Note: The build may be breaking due to versioning, I have no effect on it.

### DIFF
--- a/src/Discord.Net.Core/CDN.cs
+++ b/src/Discord.Net.Core/CDN.cs
@@ -56,10 +56,13 @@ namespace Discord
         /// <returns>
         ///     A URL pointing to the guild's icon in the specified size.
         /// </returns>
-        public static string GetGuildIconUrl(ulong guildId, string iconId, ushort size, ImageFormat format)
+        public static string GetGuildIconUrl(ulong guildId, string iconId, ImageFormat format, ushort size)
         {
             if (string.IsNullOrWhiteSpace(iconId))
                 return null;
+            if (format == ImageFormat.Gif)
+                throw new ArgumentException("Requested image format mustn't be a gif.");
+
             string extension = FormatToExtension(format, iconId);
             return $"{DiscordConfig.CDNUrl}icons/{guildId}/{iconId}.{extension}?size={size}";
         }
--- a/src/Discord.Net.Core/CDN.cs
+++ b/src/Discord.Net.Core/CDN.cs
@@ -12,26 +12,45 @@ namespace Discord
         /// </summary>
         /// <param name="appId">The application identifier.</param>
         /// <param name="iconId">The icon identifier.</param>
+        /// <param name="format">The format to return. Mustn't be a gif.</param>
+        /// <param name="size">The size of the image to return in. This can be any power of two between 16 and 2048.</param>
         /// <returns>
         ///     A URL pointing to the application's icon.
         /// </returns>
-        public static string GetApplicationIconUrl(ulong appId, string iconId)
-            => iconId != null ? $"{DiscordConfig.CDNUrl}app-icons/{appId}/{iconId}.jpg" : null;
+        public static string GetApplicationIconUrl(ulong appId, string iconId, ImageFormat format, ushort size)
+        {
+            if (string.IsNullOrWhiteSpace(iconId))
+                return null;
+            if (format == ImageFormat.Gif)
+                throw new ArgumentException("Requested image format mustn't be a gif.");
+            if (!(size >= 16 && size <= 2048))
+                throw new ArgumentOutOfRangeException("Size must be a power of two in a range between 16 and 2048.");
+            if ((size & (size - 1)) != 0)
+                throw new ArgumentException("Size must be a power of two.");
+
+            string extension = FormatToExtension(format, iconId);
+            return $"{DiscordConfig.CDNUrl}app-icons/{appId}/{iconId}.{extension}?size={size}";
+        }
 
         /// <summary>
         ///     Returns a user avatar URL.
         /// </summary>
         /// <param name="userId">The user snowflake identifier.</param>
         /// <param name="avatarId">The avatar identifier.</param>
-        /// <param name="size">The size of the image to return in. This can be any power of two between 16 and 2048.</param>
         /// <param name="format">The format to return.</param>
+        /// <param name="size">The size of the image to return in. This can be any power of two between 16 and 2048.</param>
         /// <returns>
         ///     A URL pointing to the user's avatar in the specified size.
         /// </returns>
-        public static string GetUserAvatarUrl(ulong userId, string avatarId, ushort size, ImageFormat format)
+        public static string GetUserAvatarUrl(ulong userId, string avatarId, ImageFormat format, ushort size)
         {
-            if (avatarId == null)
+            if (string.IsNullOrWhiteSpace(avatarId))
                 return null;
+            if (!(size >= 16 && size <= 2048))
+                throw new ArgumentOutOfRangeException("Size must be a power of two in a range between 16 and 2048.");
+            if ((size & (size - 1)) != 0)
+                throw new ArgumentException("Size must be a power of two.");
+
             string extension = FormatToExtension(format, avatarId);
             return $"{DiscordConfig.CDNUrl}avatars/{userId}/{avatarId}.{extension}?size={size}";
         }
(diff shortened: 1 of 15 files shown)


src/Discord.Net.Core/Entities/Guilds/IGuild.cs
```diff
@@ -69,10 +69,12 @@ public interface IGuild : IDeletable, ISnowflakeEntity
         /// <summary>
         ///     Gets the URL of this guild's icon.
         /// </summary>
+        /// <param name="format">The format to return. Mustn't be a gif.</param>
+        /// <param name="size">The size of the image to return in. This can be any power of two between 16 and 2048.</param>
         /// <returns>
         ///     A URL pointing to the guild's icon; <c>null</c> if none is set.
         /// </returns>
-        string IconUrl { get; }
+        string GetIconUrl(ImageFormat format = ImageFormat.Auto, ushort size = 128);
```
k-boyle, line 77: Would probably be better as an extension method

src/Discord.Net.Core/Entities/Guilds/IGuild.cs
```diff
@@ -83,10 +85,12 @@ public interface IGuild : IDeletable, ISnowflakeEntity
         /// <summary>
         ///     Gets the URL of this guild's splash image.
         /// </summary>
+        /// <param name="format">The format to return. Mustn't be a gif.</param>
+        /// <param name="size">The size of the image to return in. This can be any power of two between 16 and 2048.</param>
         /// <returns>
         ///     A URL pointing to the guild's splash image; <c>null</c> if none is set.
         /// </returns>
-        string SplashUrl { get; }
+        string GetSplashUrl(ImageFormat format = ImageFormat.Auto, ushort size = 128);
```
k-boyle, line 93: Would probably be better as an extension method

src/Discord.Net.Core/Entities/Guilds/IUserGuild.cs
```diff
@@ -9,7 +9,7 @@ public interface IUserGuild : IDeletable, ISnowflakeEntity
         /// <summary>
         ///     Gets the icon URL associated with this guild, or <c>null</c> if one is not set.
         /// </summary>
-        string IconUrl { get; }
+        string GetIconUrl(ImageFormat format = ImageFormat.Auto, ushort size = 128);
```
TheCasino: Would probably be better as an extension method

k-boyle (commented):
The implementation of a lot of these get methods can just be thrown in as an extension method instead since the implementation is the same

src/Discord.Net.Core/CDN.cs
```diff
@@ -12,26 +12,45 @@ public static class CDN
...
         ///     A URL pointing to the application's icon.
         /// </returns>
-        public static string GetApplicationIconUrl(ulong appId, string iconId)
-            => iconId != null ? $"{DiscordConfig.CDNUrl}app-icons/{appId}/{iconId}.jpg" : null;
+        public static string GetApplicationIconUrl(ulong appId, string iconId, ImageFormat format, ushort size)
+        {
+            if (string.IsNullOrWhiteSpace(iconId))
+                return null;
+            if (format == ImageFormat.Gif)
+                throw new ArgumentException("Requested image format mustn't be a gif.");
```
Chris-Johnston, line 25:
```suggestion
                throw new ArgumentException(message: "Requested image format mustn't be a gif.", paramName: nameof(format));
```
I think we should include `paramName` in `ArgumentException`s when possible. I'm using named parameters because I know that these parameters are sometimes switched around.

src/Discord.Net.Core/CDN.cs
```diff
@@ -12,26 +12,45 @@ public static class CDN
...
+            if (!(size >= 16 && size <= 2048))
+                throw new ArgumentOutOfRangeException("Size must be a power of two in a range between 16 and 2048.");
```
Chris-Johnston, line 27:
```suggestion
                throw new ArgumentOutOfRangeException(message: "Size must be a power of two in a range between 16 and 2048.", paramName: nameof(size));
```

src/Discord.Net.Core/CDN.cs
```diff
@@ -12,26 +12,45 @@ public static class CDN
...
+            if ((size & (size - 1)) != 0)
+                throw new ArgumentException("Size must be a power of two.");
```
Chris-Johnston, line 29:
```suggestion
                throw new ArgumentException(message: "Size must be a power of two.", paramName: nameof(size));
```

src/Discord.Net.Core/CDN.cs
```diff
@@ -12,26 +12,45 @@ public static class CDN
...
+        /// <param name="size">The size of the image to return in. This can be any power of two between 16 and 2048.</param>
         /// <returns>
         ///     A URL pointing to the application's icon.
         /// </returns>
-        public static string GetApplicationIconUrl(ulong appId, string iconId)
-            => iconId != null ? $"{DiscordConfig.CDNUrl}app-icons/{appId}/{iconId}.jpg" : null;
+        public static string GetApplicationIconUrl(ulong appId, string iconId, ImageFormat format, ushort size)
+        {
+            if (string.IsNullOrWhiteSpace(iconId))
+                return null;
```
Chris-Johnston, line 23: I think it seems odd that we mix throwing exceptions and returning a null value here. Maybe this could return a `ArgumentNullException`?

src/Discord.Net.Core/CDN.cs
```diff
@@ -51,54 +70,87 @@ public static string GetDefaultUserAvatarUrl(ushort discriminator)
...
+        /// <param name="size">The size of the image to return in. This can be any power of two between 16 and 2048.</param>
         /// <returns>
-        ///     A URL pointing to the guild's icon.
+        ///     A URL pointing to the guild's icon in the specified size.
         /// </returns>
-        public static string GetGuildIconUrl(ulong guildId, string iconId)
-            => iconId != null ? $"{DiscordConfig.CDNUrl}icons/{guildId}/{iconId}.jpg" : null;
+        public static string GetGuildIconUrl(ulong guildId, string iconId, ImageFormat format, ushort size)
+        {
+            if (string.IsNullOrWhiteSpace(iconId))
```
Chris-Johnston, line 80: These checks are repeated a lot. Maybe have a method just for validating these parameters so we don't repeat this over and over again?

src/Discord.Net.Core/Entities/IApplication.cs
```diff
@@ -21,7 +21,9 @@ public interface IApplication : ISnowflakeEntity
         /// <summary>
         ///     Gets the icon URL of the application.
         /// </summary>
-        string IconUrl  { get; }
+        /// <param name="format">The format to return. Mustn't be a gif.</param>
```
Chris-Johnston, line 24:
```suggestion
        /// <param name="format">The format to return. Must not be of type <see cref="ImageFormat.Gif"/></param>
```
Refers to the type, which should have a link in docs.

Chris-Johnston (changes requested):

Chris-Johnston:
Just so you know, for future PR's you may want to checkout a new branch for the feature that you are working on, rather than the default branch. Not a big deal though.

src/Discord.Net.Core/Entities/Guilds/IGuild.cs
AntiTcb: Agreed. Preserve the property definitions to what they're already doing now, and just expose the method publically for people that want to explicitly control what they get back.

Additionally, reverse the parameter order so that it's consistent with already existing methods and we don't need to break those.

sudo-gab: Alright, noted. Should I keep both the properties and the methods or reverse the method changes and just let them be in the CDN file? Methods are already public there.

src/Discord.Net.Core/CDN.cs
sudo-gab: Probably, following the old structure, it returned null instead of an exception.

src/Discord.Net.Core/CDN.cs
sudo-gab: Yeah, I'd only see a use in having a method to validate the size. Validating the id would be calling a method in any way possible.

PR #1236, opened 2019-01-10
Implement messages bulk deleted event (from previous PR #1120)

NovusTheory:
## About
This re-implemented MessagesBulkDeleted event from an old PR that was closed by the author and forgotten about.

I give credit to the original author of the code in #1120 as I referenced them to rewrite it in.

This event would tremendously help with logging bots so as to possibly ignore bulk deletes as they get quite spammy or to perform a different action, though this isn't limited to that scope and is just useful to know that a set of messages were bulk deleted.

## Changes
- A new BaseSocketClient#MessagesBulkDeleted event
- MESSAGE_DELETE_BULK gateway event no longer fires individual BaseSocketClient#MessageDeleted events


Still34:
A suggested documentation fix for this implementation has been added @ https://github.com/Still34/Discord.Net/commit/af338565a5ef5ad04bb1eb0eee665282e881fac1
However, I do not have push permission since I am not a maintainer. Perhaps @foxbot could take a look?

foxbot (approved):

PR #1240, opened 2019-01-18
Add AddGuildUserAsync guild ID extension method.

AntiTcb:
Ext. method forgoes returning a guild user as a full guild object won't available to create a RestGuildUser for.

### DIFF
--- a/src/Discord.Net.Rest/Extensions/ClientExtensions.cs
+++ b/src/Discord.Net.Rest/Extensions/ClientExtensions.cs
@@ -1,0 +1,23 @@
+using System;
+using System.Threading.Tasks;
+
+namespace Discord.Rest.Extensions
+{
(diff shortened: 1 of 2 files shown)


src/Discord.Net.Rest/Extensions/ClientExtensions.cs
```diff
@@ -0,0 +1,23 @@
+using System;
+using System.Threading.Tasks;
+
+namespace Discord.Rest.Extensions
```
Chris-Johnston, line 4:
```suggestion
namespace Discord.Rest
```
Namespace is inconsistent with the other extension class in this directory.

Chris-Johnston (changes requested):

foxbot (approved):

PR #1241, opened 2019-01-18
Add Webhook example.

AntiTcb:


Chris-Johnston:
Depending on if/when #1260 is merged, should we update this example to pass the entire webhook URL? Might be easier to explain as users won't have to pick out the id and token from the string.

samples/04_webhook_client/Program.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        static void Main(string[] args)
+            => new Program().MainAsync().GetAwaiter().GetResult();
+
+        public async Task MainAsync()
+        {
+            // The webhook url follows the format https://discordapp.com/api/webhooks/{id}/{token}
+            // Because anyone with the webhook URL can use your webhook
+            // you should NOT hard code the URL or ID + token into your application.
+            using (var client = new DiscordWebhookClient(webhookId: 0, webhookToken: ""))
```
Chris-Johnston:
```suggestion
            using (var client = new DiscordWebhookClient(Environment.GetEnvironmentVariable("WEBHOOK_URL")))
```
We can update this to use the new constructor. Comments should be updated to say that the env var needs to be set, or we can just let this example have a hardcoded string (since that's what it already was, effectively)

Chris-Johnston (changes requested):

samples/04_webhook_client/Program.cs
AntiTcb: Enough people get tripped up over the use of environment variables in the samples, I opted to just throw a hardcoded dummy url into it instead.

Chris-Johnston (approved):

PR #1244, opened 2019-01-25
Fix DM permissions: AddReactions is missing.

hollowstrawberry:
AddReactions is missing from the library's DM permissions. I noticed this months ago and brought it up in the Discord API channel, but I thought it was so obvious that someone else would PR it. I'm sorry about the bystander effect.

This aligns with the actual behavior in Discord DM channels.

Changes:
-Switched one bit from 0 to 1 (0x40)


Chris-Johnston (changes requested):
Could you please update the tests to include the `AddReactions` flag? As of now, the tests only run on non-pull request branches, which is why the build hasn't failed in CI. (Unrelated, but down the line I think we should fix that)

Edit, all tests that are affected by this:
  - [ChanelPermissionsTests.TestChannelPermission (add missing flag)](https://github.com/RogueException/Discord.Net/blob/db50badcc4306b4cdbc49e29d15b9ffe1b8984dc/test/Discord.Net.Tests/Tests.ChannelPermissions.cs#L94)
  - PermissionsTests.TestOverwritePermissionsDM (change from inherit to allow)
  - PermissionsTests.TestPermissionsHasChannelPermissionDM (change false to true)

SubZero0:
Since permissions tests were changed since then, that isn't needed anymore.

PR #1247, opened 2019-01-27
Implement Client Status Support

Chris-Johnston:
Adds support for using the client_status as sent as part of the Presence model. This value can be used to determine if a user is active on the native desktop app, the mobile app, or the website.

From my testing, I only found that this value was supplied over sockets, and not REST. The API sends a dictionary that contains a set of clients and the status on each of them. Given that you cannot have a unique status per device, these values are all the same. I'm speculating that maybe this was done because you cannot really have a Set in JSON (otherwise it'd be an array).

```json
"client_status": { "desktop": "dnd", "mobile": "dnd" }
```

Given that there's already a `Status` property, I ignore the value for each of these keys (assumed to all be the same) and expose an `IImmutableSet<ClientType>`.

### DIFF
--- a/src/Discord.Net.Core/Entities/Users/IPresence.cs
+++ b/src/Discord.Net.Core/Entities/Users/IPresence.cs
@@ -18,7 +18,6 @@ namespace Discord
         /// <summary>
         ///     Gets the set of clients where this user is currently active.
         /// </summary>
-        // TODO: Un-comment ActiveClients definition in IPresence for next breaking update
-        // IImmutableSet<ClientType> ActiveClients { get; }
+        IImmutableSet<ClientType> ActiveClients { get; }
     }
 }
--- a/src/Discord.Net.Core/Entities/Users/IPresence.cs
+++ b/src/Discord.Net.Core/Entities/Users/IPresence.cs
@@ -1,3 +1,6 @@
+
+using System.Collections.Immutable;
+
 namespace Discord
 {
     /// <summary>
@@ -13,5 +16,9 @@ namespace Discord
         ///     Gets the current status of this user.
         /// </summary>
         UserStatus Status { get; }
+        /// <summary>
+        ///     Gets the set of clients where this user is currently active.
+        /// </summary>
+        IImmutableSet<ClientType> ActiveClients { get; }
     }
 }
--- a/src/Discord.Net.Core/Entities/Users/IPresence.cs
+++ b/src/Discord.Net.Core/Entities/Users/IPresence.cs
@@ -18,6 +18,7 @@ namespace Discord
         /// <summary>
         ///     Gets the set of clients where this user is currently active.
         /// </summary>
-        IImmutableSet<ClientType> ActiveClients { get; }
+        // TODO: Un-comment ActiveClients definition in IPresence for 2.1.x breaking update
+        // IImmutableSet<ClientType> ActiveClients { get; }
     }
 }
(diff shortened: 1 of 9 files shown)


src/Discord.Net.Core/Entities/Users/IPresence.cs
```diff
@@ -13,5 +15,9 @@ public interface IPresence
         ///     Gets the current status of this user.
         /// </summary>
         UserStatus Status { get; }
+        /// <summary>
+        ///     Gets the set of clients where this user is currently active.
+        /// </summary>
+        IImmutableSet<ClientType> ActiveClients { get; }
```
foxbot, line 21: Maybe we should avoid adding an interface member here? Otherwise this will have to be a breaking update.

edit: commenting it out with a note for future reference is probably a better option than removing outright

Chris-Johnston: I left a `TODO: Un-comment ActiveClients definition in IPresence for 2.1.x breaking update` comment. I'm assuming that a breaking revision this small wouldn't require a major release, given that I'm not sure it would affect much other than addons, maybe?

foxbot: As per semver, any breaking change (API or ABI) requires a major version bump. Adding an interface member is an API breaking change (since anyone implementing the IPresence interface would need to implement this method after an upgrade).

AntiTcb: Wait till C#8 then we could throw a default interface implementation and it wouldn't be breaking anymore to add to the interface.

Chris-Johnston: Is there any way we could use preprocessor directives for the C# version?

AntiTcb: Not that I'm aware of.

Joe4evr: > Wait till C#8 then we could throw a default interface implementation and it wouldn't be breaking anymore to add to the interface.

That would force us to target only .NET Core instead of .NET Standard.

Moreover, the interfaces are meant only for consumption, not re-implementation, so semver can take a hike here.

src/Discord.Net.Core/Entities/Users/IPresence.cs
```diff
@@ -13,5 +15,10 @@ public interface IPresence
         ///     Gets the current status of this user.
         /// </summary>
         UserStatus Status { get; }
+        /// <summary>
+        ///     Gets the set of clients where this user is currently active.
+        /// </summary>
+        // TODO: Un-comment ActiveClients definition in IPresence for next breaking update
```
foxbot: you can un-comment this since we've changed our stance on SemVer

src/Discord.Net.Rest/API/Common/Presence.cs
```diff
@@ -18,5 +19,7 @@ internal class Presence
         public Optional<ulong[]> Roles { get; set; }
         [JsonProperty("nick")]
         public Optional<string> Nick { get; set; }
+        [JsonProperty("client_status")]
+        public Optional<Dictionary<string, string>> ClientStatus { get; set; }
```
foxbot, line 28: could you include a comment with the json payload here? would make it easier to understand why this is implemented as a Dict<string, string> (besides the obvious answer that "that's just what works here")

foxbot (changes requested):

foxbot (approved):

Issue #816, opened 2017-09-16
Inconsistency between Emoji and Emote


Still34:
At the time of writing, to create an `Emoji` object, you'd have to pass the unicode into its public constructor; to create an `Emote` object, you'd have to do it through either the `Parse` or the `TryParse` method, and the object's constructor is `internal` only.

This feels extremely inconsistent and can sometimes be confusing.

khionu:
Does Discord even make a distinction between Emoji and Emotes in the API?

Edit: outside of Discord, Emoji and Emotes are used interchangeably, so if the distinction is not concrete in the API, I would argue we shouldn't make it concrete, either.

foxbot:
With #793, since we're going to be importing a list of emojis from Unicode, we may as well expose Parse/TryParse on Emoji to accept an alias, while still retaining the constructor for pre-encoded emoji strings.

As for keeping a concrete distinction, Discord does provide an Emote structure in the API (at the guild-level). Removing Emoji would put us back at the original behavior, where `AddReactionAsync` takes either an `Emote` or a `string`, opening up more opportunities for users to throw invalid data at the API and get a 400 back.

BinkanSalaryman:
Emote: To give expression to emotion.
Emoji: Japanese for picture-character(s). (Yes, 絵文字 has no aspect of time like any jp. noun; "emojis" hurts to read. 🌵)

Mostly they can be used interchangeably in this case. But 🆎 is definitly not an emote. ;)

Still34:
@BinkanSalaryman did you read the issue? The definition has nothing to do with the issue. The point of the issue was to point out the inconsistent implementation. 

Edit: Also, the terms used in this lib all follow the terms defined by Discord. Emote is specifically for a custom guild emote, while Emoji is for standard Unicode ones.

BinkanSalaryman:
@Still34 I did, why so pricky? I hope I lifted your confusion that drove you to submit the naming as an issue!

> Also, the terms used in this lib all follow the terms defined by Discord. Emote is specifically for a custom guild emote, while Emoji is for standard Unicode ones.

Discord only defines [emoji](https://discordapp.com/developers/docs/resources/emoji), desktop application has written emoji too in server/guild settings. Where did emote come from?

Still34:
Apologies if I sounded hostile; ~~however, I still fail to understand why you brought up the definition when the issue itself was questioning the implementation between the current Emote and Emoji.~~

[...]

PR #1253, opened 2019-01-29
v3-review: core emote implementation

foxbot:
Requesting API review for this pattern.

Notable points of contention might include `EmoteBuilder`, not sure how I feel about a static class+method to construct something like emotes, though static builders seem best to use with an interface-only pattern.

Implementation of `IGuildEmote` is a little wonky as well; makes sense for `AttachedGuildEmote`, but then `Emote` itself is a bit odd... it can't fit into `IGuildEmote`, and I don't quite like making an `IEmote`->`IGuildEmote`->`ICachedGuildEmote` pattern... maybe this would be most ideal though? Move emote manipulation methods to IGuildEmote, and keep properties on ICachedGuildEmote?

### DIFF
--- a/src/Discord.Net/Entities/Emotes/IEmote.cs
+++ b/src/Discord.Net/Entities/Emotes/IEmote.cs
@@ -1,7 +1,19 @@
 namespace Discord
 {
-    public interface IEmote : IMentionable // TODO: Is `Mention` the correct verbage here?
+    /// <summary>
+    ///     An emote which may be used as a reaction or embedded in chat.
+    ///
+    ///     This includes Unicode emoji as well as unattached guild emotes.
+    /// </summary>
+    public interface IEmote : ITaggable
     {
+        /// <summary>
+        ///     The display-name of the emote.
+        /// </summary>
+        /// <remarks>
+        ///     For Unicode emoji, this is the raw value of the character, not its
+        ///     Unicode display name.
+        /// </remarks>
         string Name { get; }
     }
 }
--- a/src/Discord.Net/Entities/Emotes/IEmote.cs
+++ b/src/Discord.Net/Entities/Emotes/IEmote.cs
@@ -1,10 +1,6 @@
-using System;
-using System.Collections.Generic;
-using System.Text;
-
 namespace Discord
 {
-    public interface IEmote
+    public interface IEmote : IMentionable // TODO: Is `Mention` the correct verbage here?
     {
         string Name { get; }
     }
--- a/src/Discord.Net/Entities/Emotes/IGuildEmote.cs
+++ b/src/Discord.Net/Entities/Emotes/IGuildEmote.cs
@@ -3,13 +3,17 @@ using System.Threading.Tasks;
 
 namespace Discord
 {
+    /// <summary>
+    ///     An emote attached to a guild. This differs from an <see cref="IEmote"/> in that it contains
+    ///     information relevant to the source guild.
+    /// </summary>
     public interface IGuildEmote : IEmote, ISnowflakeEntity, IDeletable
     {
         /// <summary>
         ///     Gets whether this emoji is managed by an integration.
         /// </summary>
         /// <returns>
-        ///     A boolean that determines whether or not this emote is managed by a Twitch integration.
+        ///     A boolean that determines whether or not this emote is managed by an external integration, such as Twitch.
         /// </returns>
         bool IsManaged { get; }
         /// <summary>
--- a/src/Discord.Net/Entities/Emotes/IGuildEmote.cs
+++ b/src/Discord.Net/Entities/Emotes/IGuildEmote.cs
@@ -1,7 +1,44 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
 namespace Discord
 {
-    public interface IGuildEmote : IEmote
+    public interface IGuildEmote : IEmote, ISnowflakeEntity, IDeletable
     {
-        // TODO
+        /// <summary>
+        ///     Gets whether this emoji is managed by an integration.
+        /// </summary>
+        /// <returns>
+        ///     A boolean that determines whether or not this emote is managed by a Twitch integration.
+        /// </returns>
+        bool IsManaged { get; }
+        /// <summary>
+        ///     Gets whether this emoji must be wrapped in colons.
+        /// </summary>
+        /// <returns>
+        ///     A boolean that determines whether or not this emote requires the use of colons in chat to be used.
+        /// </returns>
+        bool RequireColons { get; }
+        /// <summary>
+        ///     Gets the roles that are allowed to use this emoji.
+        /// </summary>
+        /// <returns>
+        ///     A read-only list containing snowflake identifiers for roles that are allowed to use this emoji.
+        /// </returns>
+        IReadOnlyList<IRole> Roles { get; }
+        /// <summary>
+        ///     Gets the user ID associated with the creation of this emoji.
+        /// </summary>
+        /// <returns>
+        ///     An <see cref="ulong"/> snowflake identifier representing the user who created this emoji; 
+        ///     <c>null</c> if unknown.
+        /// </returns>
+        ulong? CreatorId { get; }
+        /// <summary>
+        ///     Gets the guild this emote sourced from.
+        /// </summary>
+        IGuild Guild { get; }
+
+        Task ModifyAsync(); // TODO
     }
 }
--- a/src/Discord.Net/Models/Emotes/Emote.cs
+++ b/src/Discord.Net/Models/Emotes/Emote.cs
@@ -13,6 +13,6 @@ namespace Discord
         public ulong Id { get; set; }
         public string Name { get; set; }
 
-        public string Mention => EmoteUtilities.FormatGuildEmote(Id, Name);
+        public string Tag => EmoteUtilities.FormatGuildEmote(Id, Name);
     }
 }
--- a/src/Discord.Net/Models/Emotes/Emote.cs
+++ b/src/Discord.Net/Models/Emotes/Emote.cs
@@ -1,0 +1,18 @@
+namespace Discord
+{
+    // placeholder for user-constructed guild emotes
+    // TODO: naming - should be called GuildEmote? but does not impl IGuildEmote, so maybe not...
+    internal class Emote : IEmote
+    {
+        public Emote(ulong id, string name)
+        {
+            Id = id;
+            Name = name;
+        }
+
+        public ulong Id { get; set; }
+        public string Name { get; set; }
+
+        public string Mention => throw new System.NotImplementedException(); // TODO: EmojiUtils
+    }
+}
--- a/src/Discord.Net/Models/Emotes/Emote.cs
+++ b/src/Discord.Net/Models/Emotes/Emote.cs
@@ -13,6 +13,6 @@ namespace Discord
         public ulong Id { get; set; }
         public string Name { get; set; }
 
-        public string Mention => throw new System.NotImplementedException(); // TODO: EmojiUtils
+        public string Mention => EmoteUtilities.FormatGuildEmote(Id, Name);
     }
 }
--- a/src/Discord.Net/Utilities/EmoteUtilities.cs
+++ b/src/Discord.Net/Utilities/EmoteUtilities.cs
@@ -7,21 +7,28 @@ namespace Discord
         public static string FormatGuildEmote(ulong id, string name)
             => $"<:{name}:{id}>";
 
-        public static (ulong, string) ParseGuildEmote(string formatted)
+        // TODO: perf: bench whether this should be passed by ref (in)
+        public static bool TryParseGuildEmote(ReadOnlySpan<char> formatted, out (ulong, string) result)
         {
-            if (formatted.IndexOf('<') != 0 || formatted.IndexOf(':') != 1 || formatted.IndexOf('>') != formatted.Length-1)
-                throw new ArgumentException("passed string does not match a guild emote format", nameof(formatted)); // TODO: grammar
+            result = default;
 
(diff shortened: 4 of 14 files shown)


test/Discord.Tests.Unit/Utilities/EmoteTests.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            string input = "<:gopher:243902586946715658>";
+            var (resultId, resultName) = EmoteUtilities.ParseGuildEmote(input);
+            Assert.Equal(243902586946715658UL, resultId);
+            Assert.Equal("gopher", resultName);
+
+            Assert.Throws<ArgumentException>(() => EmoteUtilities.ParseGuildEmote("foo"));
+            Assert.Throws<ArgumentException>(() => EmoteUtilities.ParseGuildEmote("<foo"));
+            Assert.Throws<ArgumentException>(() => EmoteUtilities.ParseGuildEmote("<:foo"));
+            Assert.Throws<ArgumentException>(() => EmoteUtilities.ParseGuildEmote("<:foo>"));
```
Chris-Johnston: I think a `Theory` should be used here instead:

```cs
[Theory]
[InlineData("foo")]
[InlineData("<foo")]
[InlineData("<:foo")]
[InlineData("<:foo>")]
public void ParseInvalid(string input)
{
    Assert.Throws<ArgumentException>(() => EmoteUtilities.ParseGuildEmote(input));
}
```

src/Discord.Net/Utilities/EmoteUtilities.cs
```diff
@@ -0,0 +1,27 @@
...
+                throw new ArgumentException("passed string does not match a guild emote format", nameof(formatted)); // TODO: grammar
+
+            int closingIndex = formatted.IndexOf(':', 2);
+            if (closingIndex < 0)
+                throw new ArgumentException("passed string does not match a guild emote format", nameof(formatted));
+
+            string name = formatted.Substring(2, closingIndex-2);
+            string idStr = formatted.Substring(closingIndex + 1);
+            idStr = idStr.Substring(0, idStr.Length - 1); // ignore closing >
+            ulong id = ulong.Parse(idStr); // TODO: TryParse here?
```
Chris-Johnston: If `TryParse` is used (and an exception is thrown on the error case), I think an extension to the current tests should be added.

src/Discord.Net/Entities/Emotes/IEmote.cs
```diff
@@ -1,10 +1,6 @@
-using System;
-using System.Collections.Generic;
-using System.Text;
-
 namespace Discord
 {
-    public interface IEmote
+    public interface IEmote : IMentionable // TODO: Is `Mention` the correct verbage here?
```
FiniteReality: I'd say ITag instead of IMentionable; as it stays mostly in line with the current naming.

src/Discord.Net/Utilities/EmoteUtilities.cs
FiniteReality: Personally, I'd suggest creating a TryParseGuildEmote method which uses out params for the id/name. In addition, I'd suggest using Span-based parsing.

FiniteReality (changes requested):
For parsing, We should definitely prefer a TryParse overload, so I'd like to see one.

src/Discord.Net/Utilities/EmoteUtilities.cs
Chris-Johnston: we could also avoid repeating the same `throw new ArgumentException ...` line if `ParseGuildEmote` called `TryParseGuildEmote`

src/Discord.Net/Entities/Emotes/IGuildEmote.cs
```diff
@@ -1,7 +1,44 @@
...
 {
-    public interface IGuildEmote : IEmote
+    public interface IGuildEmote : IEmote, ISnowflakeEntity, IDeletable
     {
-        // TODO
+        /// <summary>
+        ///     Gets whether this emoji is managed by an integration.
+        /// </summary>
+        /// <returns>
+        ///     A boolean that determines whether or not this emote is managed by a Twitch integration.
```
AntiTcb: Minor nit, but this doc probably shouldn't exclusively mention Twitch. I'm sure other integrations that can create managed emoji will come in the future, most likely YouTube being ntext.

src/Discord.Net/Entities/Emotes/IGuildEmote.cs
```diff
@@ -1,7 +1,44 @@
...
+        /// </summary>
+        /// <returns>
+        ///     An <see cref="ulong"/> snowflake identifier representing the user who created this emoji; 
+        ///     <c>null</c> if unknown.
+        /// </returns>
+        ulong? CreatorId { get; }
+        /// <summary>
+        ///     Gets the guild this emote sourced from.
+        /// </summary>
+        IGuild Guild { get; }
```
AntiTcb, line 44: Depending on the decision of the above comment for CreatorId, a GuildId prop should probably be added as well.

src/Discord.Net/Entities/Emotes/IGuildEmote.cs
```diff
@@ -1,7 +1,44 @@
...
+        /// </returns>
+        IReadOnlyList<IRole> Roles { get; }
+        /// <summary>
+        ///     Gets the user ID associated with the creation of this emoji.
+        /// </summary>
+        /// <returns>
+        ///     An <see cref="ulong"/> snowflake identifier representing the user who created this emoji; 
+        ///     <c>null</c> if unknown.
+        /// </returns>
+        ulong? CreatorId { get; }
```
AntiTcb, line 40: The API provides a partial user object for this when it can. 

How do we want to handle cases like this? I see a few possibilities in my head.

- Providing both the parent id and parent object props. If the parent object is partial, the prop is null.
- Same as above, but a partial parent object is provided with whatever data was passed from the API. (This may be a little hairy if we also decide to utilize nullable reference types in the lib).
- Only one of the props is provided, most likely Id.

src/Discord.Net/Models/Emotes/Emote.cs
```diff
@@ -0,0 +1,18 @@
+namespace Discord
+{
+    // placeholder for user-constructed guild emotes
+    // TODO: naming - should be called GuildEmote? but does not impl IGuildEmote, so maybe not...
+    internal class Emote : IEmote
+    {
+        public Emote(ulong id, string name)
```
AntiTcb, line 7: I'd swap the parameter order here to keep consistent with the `<:name:id>` format.

src/Discord.Net/Entities/Emotes/IGuildEmote.cs
```diff
@@ -1,7 +1,44 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
 namespace Discord
 {
-    public interface IGuildEmote : IEmote
+    public interface IGuildEmote : IEmote, ISnowflakeEntity, IDeletable
```
AntiTcb, line 10: Now that I think about it, (and while the concretes are classes and not structs for the moment), is there any pressing reason to keep the name for this as `IGuildEmote` and not `IGuildEmoji`? The API docs don't give credence to the existence of an "emote," and while we could use that as a shared interface for `GuildEmote/GuildEmoji` and `Emoji`, I'd like to cut down on one more nuance (Emote v Emoji) that users would need to be aware of that isn't immediately obvious. 

Ideally, it'd flow as `IEmoji` -> `Emoji` -> `GuildEmoji`. I feel like I'm missing something for why we presently have `Emote` and `GuildEmote`. Is the former really necessary for representing user created guild emoji?

Still34:
I think with the design of v3 emoji we should really reconsider the naming of all related objects. As pointed out in https://github.com/discord-net/Discord.Net/pull/1253#discussion_r252939808, ideally the naming scheme should be something like `UnicodeEmoji` and or `XEmoji` in general to avoid the confusion of Emote vs Emoji, which happens very very often with v2.x and prior as evident in https://github.com/discord-net/Discord.Net/issues/816.

src/Discord.Net/Entities/Emotes/IGuildEmote.cs
foxbot: keeping this as-is; IGuildEmote is only for guild-attached emotes, which means the user will have direct access to the guild it sourced from.

easier to keep the interface here light

src/Discord.Net/Entities/Emotes/IGuildEmote.cs
foxbot: user-created emoji cannot be forced into the same class as guild-attached emoji, because the user will never be able to provide as much information as the guild does

these are kept separately so as not to constrain the data available in the guild attached class

src/Discord.Net/Utilities/EmoteUtilities.cs
```diff
@@ -7,21 +7,28 @@ public static class EmoteUtilities
...
-        public static (ulong, string) ParseGuildEmote(string formatted)
+        // TODO: perf: bench whether this should be passed by ref (in)
+        public static bool TryParseGuildEmote(ReadOnlySpan<char> formatted, out (ulong, string) result)
         {
-            if (formatted.IndexOf('<') != 0 || formatted.IndexOf(':') != 1 || formatted.IndexOf('>') != formatted.Length-1)
-                throw new ArgumentException("passed string does not match a guild emote format", nameof(formatted)); // TODO: grammar
+            result = default;
 
-            int closingIndex = formatted.IndexOf(':', 2);
+            if (formatted.IndexOf('<') != 0 || formatted.IndexOf(':') != 1 || formatted.IndexOf('>') != formatted.Length - 1)
```
FiniteReality, line 15: We should use something like this instead here:
```cs
if (formatted.Length < 7)
    return false;
if (formatted[0] != '<' && formatted[1] == ':' && formatted[formatted.Length-1] != '>')
    return false;
```
This avoids a O(n) lookup for each character, since we know their positions anyway. The second colon will still have to use IndexOf, however.

src/Discord.Net/Utilities/EmoteUtilities.cs
```diff
@@ -7,21 +7,28 @@ public static class EmoteUtilities
...
+                return false;
+
+            int closingIndex = formatted.LastIndexOf(':');
             if (closingIndex < 0)
-                throw new ArgumentException("passed string does not match a guild emote format", nameof(formatted));
+                return false;
+
+            ReadOnlySpan<char> name = formatted.Slice(2, closingIndex-2);
+            ReadOnlySpan<char> idStr = formatted.Slice(closingIndex + 1, formatted.Length - (name.Length + 4));
```
FiniteReality, line 23: We can probably simplify this to `Slice(closingIndex + 1, formatted.Length - closingIndex - 2)` and then avoid the later slice to remove the `>`

src/Discord.Net/Utilities/EmoteUtilities.cs
```diff
@@ -7,21 +7,28 @@ public static class EmoteUtilities
         public static string FormatGuildEmote(ulong id, string name)
             => $"<:{name}:{id}>";
 
-        public static (ulong, string) ParseGuildEmote(string formatted)
+        // TODO: perf: bench whether this should be passed by ref (in)
+        public static bool TryParseGuildEmote(ReadOnlySpan<char> formatted, out (ulong, string) result)
```
FiniteReality, line 11: I'd suggest returning a ReadOnlySpan<char> instead of a string to avoid allocating another string here. If the user needs a string, they can easily allocate one themselves.

src/Discord.Net/Utilities/EmoteUtilities.cs
```diff
@@ -7,21 +7,28 @@ public static class EmoteUtilities
...
+            idStr = idStr.Slice(0, idStr.Length - 1); // ignore closing >
+
+            if (!ulong.TryParse(idStr.ToString(), out ulong id))
+                return false;
 
-            string name = formatted.Substring(2, closingIndex-2);
-            string idStr = formatted.Substring(closingIndex + 1);
-            idStr = idStr.Substring(0, idStr.Length - 1); // ignore closing >
-            ulong id = ulong.Parse(idStr); // TODO: TryParse here?
+            result = (id, name.ToString());
```
FiniteReality, line 29: Likewise, I'd avoid allocating here and simply return the ROS for the name.

FiniteReality (commented):
I didn't get to point these out when you made the commit - sorry about that.

src/Discord.Net/Utilities/EmoteUtilities.cs
FiniteReality: The 7 here is *AFAIK* the minimum length of a valid emote: `<:aa:0>` I personally don't know of any emote which is shorter than this.

foxbot: not sure what good the `length < 7` check does, but checking known indices makes sense, done.

FiniteReality: The span indexer throws if you try to index an out-of-range value - the length check ensures the indexers work later.

foxbot: yep that's definitely something that could happen

Joe4evr: > ```cs
> if (formatted[0] != '<' && formatted[1] == ':' && formatted[formatted.Length-1] != '>')
> ```

Is this logic right, like shouldn't there be `||`'s in there instead?

FiniteReality: Err, Probably. That `==` should likely also be a `!=` too.

foxbot: yeah that was changed at some point, not sure why the diff isn't updated

PR #1255, opened 2019-01-31
Fix #1216, Add support for spoiler formatting & attachments

### LINKED ISSUE #1216: Add support for spoilers
Spoilers are on canary and they apply to text (just need to add a markdown), and images

Chris-Johnston:
Fix #1216 

Implements support for formatting text as spoilers, and spoiler attachments.
- Adds a method to convert text to spoiler text in the `Format` class.
- Adds `|` to the list of characters to escape, since this is used for formatting spoilers.
- Adds `isSpoiler` optional parameter to `SendFileAsync`.
    - Adds a field to `UploadFileParams`. When this is set to true, the file name will be prefixed with the text `"SPOILER_"`. Will also send the `hasSpoiler` value in the upload request body.
    - From my observations, the client only cares that attachments begin with `SPOILER_`, and not that the `hasSpoiler` property is supplied.
- Adds an extension method for `IAttachment` which checks if an attachment is a spoiler. Spoilers attachments can be of any type, not just images.

### DIFF
--- a/src/Discord.Net.Core/Entities/Messages/IAttachment.cs
+++ b/src/Discord.Net.Core/Entities/Messages/IAttachment.cs
@@ -6,11 +6,6 @@ namespace Discord
     public interface IAttachment
     {
         /// <summary>
-        ///     The prefix applied to files to indicate that it is a spoiler.
-        /// </summary>
-        public const string SpoilerPrefix = "SPOILER_";
-
-        /// <summary>
         ///     Gets the ID of this attachment.
         /// </summary>
         /// <returns>
--- a/src/Discord.Net.Core/Entities/Messages/IAttachment.cs
+++ b/src/Discord.Net.Core/Entities/Messages/IAttachment.cs
@@ -6,6 +6,11 @@ namespace Discord
     public interface IAttachment
     {
         /// <summary>
+        ///     The prefix applied to files to indicate that it is a spoiler.
+        /// </summary>
+        public const string SpoilerPrefix = "SPOILER_";
+
+        /// <summary>
         ///     Gets the ID of this attachment.
         /// </summary>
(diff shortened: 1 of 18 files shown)


src/Discord.Net.Core/Entities/Messages/IAttachment.cs
```diff
@@ -5,6 +5,11 @@ namespace Discord
     /// </summary>
     public interface IAttachment
     {
+        /// <summary>
+        ///     The prefix applied to files to indicate that it is a spoiler.
+        /// </summary>
+        public const string SpoilerPrefix = "SPOILER_";
```
Joe4evr: :eyes:? We're not building with preview compiler versions, are we?
How does this not produce a CI failure?

Joe4evr: [SharpLab, for reference.](https://sharplab.io/#v2:C4LgbgpgTglgZgTwDQBMQGoA+ABADAAmwEYBuAWAChsBmfGAO2GjgEMBjCfASQDEB7PpQDelfGMK02fegGdghIgQAa+ALz4ARPz4byFAL5A=)

Chris-Johnston: Nope, this was just my mistake. I didn't build before committing. Also, it seems that CI hasn't been working since we moved to the org.

FiniteReality: Yeah, we need to get it moved over from Volt's personal repo. cc @RogueException

AntiTcb:
Should also be added to `DiscordWebhookClient.SendFileAsync` if those support spoiler attachments as well.

PR #1256, opened 2019-02-05
Fix NRE when getting audit logs

SubZero0:
A NullReferenceException is thrown when getting audit log data that contains a deleted webhook.

### DIFF
--- a/src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/WebhookCreateAuditLogData.cs
+++ b/src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/WebhookCreateAuditLogData.cs
@@ -43,7 +43,7 @@ namespace Discord.Rest
         ///     Gets the webhook that was created if it still exists.
         /// </summary>
         /// <returns>
-        ///     A webhook object representing the webhook that was created if it still exists.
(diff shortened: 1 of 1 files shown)


src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/WebhookCreateAuditLogData.cs
```diff
@@ -31,23 +32,31 @@ internal static WebhookCreateAuditLogData Create(BaseDiscordClient discord, Mode
...
 
         // Doc Note: Corresponds to the *current* data
 
         /// <summary>
-        ///     Gets the webhook that was created.
+        ///     Gets the webhook that was created if it still exists.
         /// </summary>
         /// <returns>
-        ///     A webhook object representing the webhook that was created.
+        ///     A webhook object representing the webhook that was created if it still exists.
```
Chris-Johnston:
```suggestion
        ///     A webhook object representing the webhook that was created if it still exists, otherwise returns <c>null</c>.
```

Chris-Johnston (commented):
Just one small nitpick about xmldocs. I haven't worked with the audit logs myself, do we know that `entry.TargetId.Value` is always specified?

SubZero0:
That's my only concern, if discord always provides the webhook id since it's the only way to reliably identify it.

PR #1257, opened 2019-02-05
Clarify command samples 

Still34:
# Summary

This PR adds several additional notes answering questions a newcomer may have for the command framework.

### DIFF
--- a/samples/02_commands_framework/Modules/PublicModule.cs
+++ b/samples/02_commands_framework/Modules/PublicModule.cs
@@ -60,6 +60,7 @@ namespace _02_commands_framework.Modules
         public Task ListAsync(params string[] objects)
             => ReplyAsync("You listed: " + string.Join("; ", objects));
 
+        // Setting a custom ErrorMessage property will help clarify the precondition error
         [Command("guild_only")]
         [RequireContext(ContextType.Guild, ErrorMessage = "Sorry, this command must be ran from within a server, not a DM!")]
         public Task GuildOnlyCommand()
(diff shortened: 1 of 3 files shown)


samples/02_commands_framework/Services/CommandHandlingService.cs
```diff
@@ -37,10 +41,18 @@ public async Task MessageReceivedAsync(SocketMessage rawMessage)
 
             // This value holds the offset where the prefix ends
             var argPos = 0;
+            // Perform prefix check. You may want to replace this with
+            // (!message.HasCharPrefix(_prefix, ref argPos))
```
Chris-Johnston:
```suggestion
            // (!message.HasCharPrefix('!', ref argPos))
```
I think we should provide an example that matches the format "!help" as described on the next line.

Chris-Johnston (commented):
Looks good, just one thing I would change.

samples/02_commands_framework/Modules/PublicModule.cs
```diff
@@ -60,6 +60,7 @@ public Task EchoAsync([Remainder] string text)
         public Task ListAsync(params string[] objects)
             => ReplyAsync("You listed: " + string.Join("; ", objects));
 
+        // Setting a custom ErrorMessage property will help clarify the precondition error
```
Joe4evr, line 63: This comment could be moved one line down so that it's more obvious what it's referring to.

foxbot (approved):

PR #1260, opened 2019-02-08
Add webhook url overload for DiscordWebhookClient

Chris-Johnston:
Adds an overloaded constructor for `DiscordWebhookClient` which accepts the webhook URL. This URL is parsed using a regex for the id and token. If the token is invalid an `ArgumentException` or `ArgumentNullException` is thrown.


Yucked:
Why not RegexOptions.Compiled | IgnoreCase

test/Discord.Net.Tests/Tests.DiscordWebhookClient.cs
```diff
@@ -0,0 +1,62 @@
...
+        [InlineData("https://canary.discordapp.com/api/webhooks/123412347732897802/_abcde123456789-ABCDEFGHIJKLMNOP12345678-abcdefghijklmnopABCDEFGHIJK")]
+        // don't care about https
+        [InlineData("http://canary.discordapp.com/api/webhooks/123412347732897802/_abcde123456789-ABCDEFGHIJKLMNOP12345678-abcdefghijklmnopABCDEFGHIJK")]
+        // this is the minimum that the regex cares about
+        [InlineData("p.com/api/webhooks/123412347732897802/_abcde123456789-ABCDEFGHIJKLMNOP12345678-abcdefghijklmnopABCDEFGHIJK")]
+        public void TestWebhook_Valid(string webhookurl)
+        {
+            try
+            {
+                _ = new DiscordWebhookClient(webhookurl);
```
AntiTcb: These tests should do the same format check that the client's constructor does, without using the constructor.

The Webhook client is a special snowflake and actually makes an API request in its ctor currently, so this is technically an integration test, not a unit test.

src/Discord.Net.Webhook/DiscordWebhookClient.cs
```diff
@@ -26,6 +27,12 @@ public DiscordWebhookClient(IWebhook webhook)
         /// <summary> Creates a new Webhook Discord client. </summary>
         public DiscordWebhookClient(ulong webhookId, string webhookToken)
             : this(webhookId, webhookToken, new DiscordRestConfig()) { }
+        /// <summary> Creates a new Webhook Discord client. </summary>
+        public DiscordWebhookClient(string webhookUrl)
+            : this(webhookUrl, new DiscordRestConfig()) { }
+
+        // regex pattern to match webhook urls
+        private const string WebhookUrlRegex = @"^.+.com\/api\/webhooks\/([\d]+)\/([A-Za-z0-9_-]+)$";
```
AntiTcb: This should validate that the domain is discordapp.com as well.

src/Discord.Net.Webhook/DiscordWebhookClient.cs
```diff
@@ -43,6 +50,41 @@ public DiscordWebhookClient(IWebhook webhook, DiscordRestConfig config)
...
+        /// <exception cref="ArgumentException">Thrown if the <paramref name="webhookUrl"/> is an invalid format.</exception>
+        /// <exception cref="ArgumentNullException">Thrown if the <paramref name="webhookUrl"/> is null or whitespace.</exception>
+        public DiscordWebhookClient(string webhookUrl, DiscordRestConfig config) : this(config)
+        {
+            if (string.IsNullOrWhiteSpace(webhookUrl))
+                throw new ArgumentNullException(paramName: nameof(webhookUrl), message:
+                    "The given webhook Url cannot be null or whitespace.");
+
+            // thrown when groups are not populated/valid, or when there is no match
+            ArgumentException ex()
```
AntiTcb: Add an optional argument to this location function to specify whether the id or the token was what didn't pass the format check.

src/Discord.Net.Webhook/DiscordWebhookClient.cs
foxbot: this should be a `static Regex`, rather than a `const string`, so it only needs to be compiled once

Chris-Johnston:
According to the docs for the Compiled flag: ["Specifies that the regular expression is compiled to an assembly. This yields faster execution but increases startup time. "](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regexoptions?view=netcore-2.2)
Given that I'm only using this regex once, I don't think that flag should be used.

Webhooks are case sensitive, so I wouldn't expect someone to `.Upper()` the entire string, or something. The expression already handles both upper and lower case for the token group.

src/Discord.Net.Webhook/DiscordWebhookClient.cs
Chris-Johnston: re: the other comment, this would be a good fit for this flag since the regex can be reused.

AntiTcb (changes requested):

test/Discord.Net.Tests/Tests.DiscordWebhookClient.cs
Chris-Johnston: Maybe move the parsing logic into some static internal function (since tests have internal access)?

AntiTcb: Sounds good to me

Chris-Johnston: I really wish I could have used a ValueTuple here https://github.com/discord-net/Discord.Net/pull/1260/files#diff-062b7bba813fb6aac2adf2a6f980eb7fR120

Chris-Johnston:
@foxbot were we going to start having PRs update lines in the changelog? if so I can start doing this here

Update: nope

foxbot (approved):

PR #1261, opened 2019-02-08
Add GetCategoryChannel method

Still34:


Chris-Johnston (approved):
Maybe instead of just "if none is found" we could also explain null is returned if the channel id is not a category channel.

Otherwise lgtm

foxbot (approved):

bloodwiing:
So... Is a merge happening anytime?
~~I've built the WebSocket library from Still34's forked repo and am using that, but would still like to use the Nuget one. 🙂~~
I'd need to rewrite everything to make that work. So will wait for this...

foxbot (approved):

PR #1276, opened 2019-02-28
Use a relative path for docs index page logo

Chris-Johnston:
Changes the paths for the logo on the index page of the documentation
to use relative links instead of absolute ones.

Before:
before image

After:
after image

The current absolute path is just fine as long as the site is not hosted under another directory. When opening files locally (without serving them in docfx) these images will fail to load.
In addition, if these files are served under a directory that is not the root endpoint (like: `docs.com/stable/`) the images will also break.

I tested these changes locally using these steps:

- Build & serve docs: `docfx docs/docfx.json --serve`
- Verify that logo shows up on index page using all 3 themes
- Open `docs/_site/index.html`
- Verify logo works on all 3 themes

Also includes a minor grammar fix.

### DIFF
--- a/docs/_template/light-dark-theme/styles/dark.css
+++ b/docs/_template/light-dark-theme/styles/dark.css
@@ -304,5 +304,5 @@ span.arrow-r{
 }
 
 .logo-switcher {
-    background: url("/marketing/logo/SVG/Combinationmark White.svg") no-repeat;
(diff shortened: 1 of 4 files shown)


Chris-Johnston:
@foxbot this will likely affect you, since you were planning on updating your documentation, and your site hosts the latest docs under `/latest/` and not `/`

foxbot (approved):

PR #1278, opened 2019-03-04
Fix false invalidation when decoding token User Ids

Chris-Johnston:
This fixes an issue when trying to validate older bot tokens. The first segment is a base64-encoded string that contains a ulong. The format that discord provides will not contain any padding, while `Convert.FromBase64String` will throw a `FormatException` when trying to decode from a string that requires it (string length is not a multiple of 4).

This PR:
- Adds a util method for padding base64 strings + tests
- Pads the first segment of the token before trying to decode it
- Adds a test for tokens that require base64 padding

[Original conversation can be found here.](https://discordapp.com/channels/81384788765712384/381889909113225237/552000399234695169)

### DIFF
--- a/src/Discord.Net.Core/Utils/TokenUtils.cs
+++ b/src/Discord.Net.Core/Utils/TokenUtils.cs
@@ -17,6 +17,49 @@ namespace Discord
         /// </remarks>
         internal const int MinBotTokenLength = 58;
 
+        internal const char Base64Padding = '=';
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


foxbot (approved):
i'm beginning to question whether or not our token invalidation logic is growing too complicated; however, this is probably a reasonable fix to have on the current codebase, so we'll accept now, reconsider later.

Chris-Johnston:
Given that it's run into a few problems I agree that it might be best to reconsider, but it seems like it's been somewhat effective at helping people identify when they are passing in invalid tokens.

PR #1283, opened 2019-03-13
Make the client public on socket entities

k-boyle:
My use-case for this is (for example) defining an extension method on `SocketGuild` without having to pass the client manually. I also changed the name of the property from `Discord` to `Client` so the naming is a bit more obvious.

```
async Task<IGuildUser> GetOrFetchUserAsync(this SocketGuild guild, ulong userId)
{
    var user = guild.GetUser(userId) as IGuildUser ?? await guild.Client.Rest.GetGuildUserAsync(guild.Id, userId);
    return user;
}
```

Whereas currently, you'd need to pass the client (imo this shouldn't be necessary) or in my case, use reflection to get the client from the entity.


AntiTcb:
@RogueException Can't remember even though I know I've asked you at least twice: why was SocketEntity.Discord made internal ages ago anyway?

FiniteReality:
IIRC it was because of the sharded client - the multiple shards were intended to be internal and hidden all along, from what I remember

k-boyle:
@FiniteReality considering that GetShard is a thing isn't that argument void?

FiniteReality:
¯\\\_(ツ)\_/¯

PR #1294, opened 2019-04-10
EmbedBuilderExtensions no longer has an extension for WithAuthor(IGuildUser)

GreemDev:
The embed should not display nickname#discriminator as that's incorrect.
So the extension was scrapped altogether to use the IUser (and by inheritance, IGuildUser) one.


Chris-Johnston:
Considering that this could break, what about marking it Deprecated for 2.x, and removing it in 3.x?

PR #1308, opened 2019-05-15
Fix: AudioClient Disconnect Race Condition

ComputerMaster1st:
Hi There,

This solves a race condition which occurs when disconnecting from voice. Some users may have noticed it, some might not. All this does is await on `_task` found inside the `ConnectionManager` when calling `StopAsync`.

This has only been locally tested, however has been installed onto a public/production bot for real-world testing. It is recommended to test this further before merging in-case of any unexpected issues.

### DIFF
--- a/src/Discord.Net.WebSocket/ConnectionManager.cs
+++ b/src/Discord.Net.WebSocket/ConnectionManager.cs
@@ -106,11 +106,13 @@ namespace Discord
                 finally { _stateLock.Release(); }
             });
         }
-        public virtual Task StopAsync()
+
+        public virtual async Task StopAsync()
         {
             Cancel();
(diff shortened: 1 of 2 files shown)


Emzi0767:
This is unawaited because if you do, it will deadlock.

Furthermore, when awaiting tasks, you should await them properly.

ComputerMaster1st:
> This is unawaited because if you do, it will deadlock.
> 
> Furthermore, when awaiting tasks, you should await them properly.

~~Can you confirm this deadlock's please? Since I'm "unreliable". I'm currently not experiencing this deadlock.~~ You were right, didn't think getawaiter would cause blocking to that extent.

EDIT: I've changed it to be async/await now.

PR #1310, opened 2019-05-23
Added id overload for RemoveReactionAsync

k-boyle:
When working in the `ReactionAdded` event if the user wasn't in cache you'd need to use a REST request to fetch them to remove their reaction, doubling your request amount even though all that's required to remove their reaction is the users id (which is provided through the event). 

If there are any other methods that should have an id overload let me know and I'll add it to this PR.

### DIFF
--- a/src/Discord.Net.Core/Entities/Messages/IUserMessage.cs
+++ b/src/Discord.Net.Core/Entities/Messages/IUserMessage.cs
@@ -84,6 +84,23 @@ namespace Discord
         /// <seealso cref="IEmote"/>
         Task RemoveReactionAsync(IEmote emote, IUser user, RequestOptions options = null);
         /// <summary>
+        ///     Removes a reaction from message.
+        /// </summary>
+        /// <example>
+        ///     The following example removes the reaction, <c>💕</c>, added by the message author from the message.
+        ///     <code language="cs">
+        ///     await msg.RemoveReactionAsync(new Emoji("\U0001f495"), 84291986575613952);
+        ///     </code>
+        /// </example>
+        /// <param name="emote">The emoji used to react to this message.</param>
+        /// <param name="userId">The id of the user that added the emoji.</param>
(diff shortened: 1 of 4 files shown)


src/Discord.Net.Core/Entities/Messages/IUserMessage.cs
```diff
@@ -84,6 +84,23 @@ public interface IUserMessage : IMessage
         /// <seealso cref="IEmote"/>
         Task RemoveReactionAsync(IEmote emote, IUser user, RequestOptions options = null);
         /// <summary>
+        ///     Removes a reaction from message.
+        /// </summary>
+        /// <example>
+        ///     The following example removes the reaction, <c>💕</c>, added by the message author from the message.
```
Chris-Johnston:
```suggestion
        ///     The following example removes the reaction, <c>💕</c>, added by the user with ID 84291986575613952 from the message.
```

src/Discord.Net.Core/Entities/Messages/IUserMessage.cs
```diff
@@ -84,6 +84,23 @@ public interface IUserMessage : IMessage
...
+        ///     <code language="cs">
+        ///     await msg.RemoveReactionAsync(new Emoji("\U0001f495"), 84291986575613952);
+        ///     </code>
+        /// </example>
+        /// <param name="emote">The emoji used to react to this message.</param>
+        /// <param name="userId">The id of the user that added the emoji.</param>
```
Chris-Johnston:
```suggestion
        /// <param name="userId">The ID of the user that added the emoji.</param>
```
keeping this consistent with other docs

Chris-Johnston (changes requested):
Just a few things to update in xmldocs.

Chris-Johnston (approved):

foxbot (approved):
i guess

PR #1312, opened 2019-05-30
Allow for custom default RequestOptions

LukeCSamuel:
This change would allow the consumer of the library to set the default `RequestOptions` if they desire.  Invoking a method that takes `RequestOptions` as a parameter without specifying the argument will use a copy of the default.  Behavior will be unchanged without modifying the `RequestOptions.Default` property.

Additionally, this change adds 4 `With...` methods that allow for fluent construction of new `RequestOptions` objects.

### DIFF
--- a/src/Discord.Net.Core/RequestOptions.cs
+++ b/src/Discord.Net.Core/RequestOptions.cs
@@ -8,9 +8,9 @@ namespace Discord
     public class RequestOptions
     {
         /// <summary>
-        ///     Creates a new <see cref="RequestOptions" /> class with its default settings.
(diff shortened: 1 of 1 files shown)


FiniteReality:
:-1: Default should always be the "most plain old simple" request options you can get. Even though it would have been more effort, I would have plumbed a RequestOptions property through the clients which is used by default.

However, the `With...` methods are :ok_hand: - I'd be fine with merging those.

LukeCSamuel:
I appreciate the suggestion; I'll take some time to work on that this weekend.

SubZero0:
Since no changes were done since then, I'll close this PR.

PR #1317, opened 2019-06-03
fix: Changed Emote GetHashCode to not include Name

cabral-tiago:
Continuation of commit abf3e90 attempt to fix Emote comparisons, as Hash comparisons are done before Equals in Comparers

### DIFF
--- a/src/Discord.Net.Core/Entities/Emotes/Emote.cs
+++ b/src/Discord.Net.Core/Entities/Emotes/Emote.cs
@@ -54,13 +54,7 @@ namespace Discord
         }
 
         /// <inheritdoc />
-        public override int GetHashCode()
-        {
-            unchecked
-            {
-                return (Name.GetHashCode() * 397) ^ Id.GetHashCode();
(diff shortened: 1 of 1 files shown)


src/Discord.Net.Core/Entities/Emotes/Emote.cs
```diff
@@ -54,13 +54,7 @@ public override bool Equals(object other)
...
 
         /// <inheritdoc />
-        public override int GetHashCode()
-        {
-            unchecked
-            {
-                return (Name.GetHashCode() * 397) ^ Id.GetHashCode();
-            }
-        }
+        public override int GetHashCode() => Id.GetHashCode();
```
AntiTcb, line 57:
```cs
public override int GetHashCode()
    => Id.GetHashCode();
```
Style change to be consistent with the rest of the lib.

AntiTcb (changes requested):

foxbot (approved):

PR #1318, opened 2019-06-04
Fix #1314 Don't parse tags within code blocks

### LINKED ISSUE #1314: Role mentions inside of codeblocks get parsed in the Tags collection of messages
If you have a message that has (for example) `@everyone` then this lib will parse this as an everyone mention (when it is not). With roles doing `@role` won't get parsed, however `<@&id>` will, which again isn't actually a mention.

Chris-Johnston:
Fix #1314 

- Prevents `MessageHelper` from parsing tags only when they are contained within inline code or a code block.
- Adds some unit tests for `MessageHelper.ParseTags` to check that it works.

In addition, one test that was left behind from a rebase has been re-added to the unit tests project. In df68b7e I exposed the internals of the projects to the new test project to fix CI tests. This should be afterwards to [remove the stale references to old projects.](https://github.com/discord-net/Discord.Net/pull/1318/files#diff-bc3bdc81fc66b1eb12b433ae205e7f3bL3)


foxbot (approved):
This looks good, and your tests seem thorough enough that the edge cases should be ironed out here. Thanks!

PR #1328, opened 2019-06-20
Fix #1327 Color.ToString returns wrong value

### LINKED ISSUE #1327: Color.ToString returns wrong value
For example `new Discord.Color(0x0080ff).ToString()` returns `#80ff`, which results in this color: https://www.color-hex.com/color/8800ff

PassiveModding:
Fix #1327 

- Pads Color#ToString to 6 characters long so it returns a valid hexadecimal color value.

### DIFF
--- a/src/Discord.Net.Core/Entities/Roles/Color.cs
+++ b/src/Discord.Net.Core/Entities/Roles/Color.cs
@@ -204,8 +204,8 @@ namespace Discord
         ///     A hexadecimal string of the color.
         /// </returns>
         public override string ToString() =>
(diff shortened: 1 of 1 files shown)


FiniteReality (changes requested):
Instead of using Convert.ToString, we may as well use String.Format.

FiniteReality (changes requested):
Use the format parameter provided in formattable strings instead of calling string.Format twice like this.

FiniteReality (approved):
I wasn't really expecting the entire thing to become string.Format, but it's close enough.

PR #1329, opened 2019-06-20
Fix invalid cref values in docs

Still34:
This PR fixes the invalid cref values complained graciously by docfx.


foxbot (approved):
thanks for you kind words as always

PR #1330, opened 2019-06-21
Fix spelling mistake in ExclusiveBulkDelete warning

PassiveModding:


foxbot (approved):
the word suppress is from hereafter banned

PR #1340, opened 2019-07-01
v3: API design review

foxbot:
Hi all,

Please use this PR to leave any comments, feedback, and concerns about the new API design for v3. You are welcome to leave a comment directly on this thread, or use a GitHub suggestion on the relevant file. 

Design feedback may also be left in the Discord.Net channel in [Discord API](https://discord.gg/discord-api).

Note that this review is focused primarily around the public API only; none of the interfaces exposed (sans the IDiscordClient) have any actual implementations yet. Implementation of the interfaces will come during a later pass.


src/Discord.Net/Entities/Channels/IGuildChannel.cs
```diff
@@ -0,0 +1,29 @@
...
+    public interface IGuildChannel : IChannel, IDeletable
+    {
+        ValueTask<IGuild> GetGuildAsync(StateBehavior stateBehavior = default, RequestOptions? options = default);
+        ulong GuildId { get; }
+
+        int Position { get; } // todo: validate that this can be populated
+
+        Task ModifyAsync(GuildChannelProperties properties, RequestOptions? options = default);
+
+        // todo: use Overwrite directly or create a wrapper?
```
k-boyle: Use them directly, the wrapper that is currently in v2 causes a lot of issues with newer users, and also looks very ugly.

src/Discord.Net/Entities/Channels/IGuildChannel.cs
```diff
@@ -0,0 +1,29 @@
...
+        IReadOnlyCollection<Overwrite> PermissionOverwrites { get; }
+        Optional<Overwrite> GetPermissionOverwrite(SnowflakeOrEntities<IUser, IRole> entity);
+
+        Task AddPermissionOverwriteAsync(SnowflakeOrEntities<IUser, IRole> entity, Overwrite overwrite, RequestOptions? options = default);
+        Task RemovePermissionOverwriteAsync(SnowflakeOrEntities<IUser, IRole> entity, RequestOptions? options = default);
+
+        // review: separate GetMember methods here, or shadow GetUser
```
k-boyle: I think a separate method is better since it's more verbose and better reflects the API's naming of guild members.  Also shadowing methods unless there are no other ways seems like a code-smell.

src/Discord.Net/Entities/Channels/IMessageChannel.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.IO;
+using System.Threading.Tasks;
+using Wumpus;
+
+namespace Discord
+{
+    public interface IMessageChannel : IChannel
+    {
+        Task<IUserMessage> SendMessageAsync(string? text = null, bool isTTS = false, IEmbed? embed = null, RequestOptions? options = default);
+        Task<IUserMessage> SendFileAsync(Stream fileStream, string filename, string? text = null, bool isTTS = false, IEmbed? embed = null, RequestOptions? options = default);
```
k-boyle: Could maybe considering a `FileProperties` class? This would mean you wouldn't need the `SendFileAsync` overloads/extensions.

src/Discord.Net/Entities/Channels/INestedChannel.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public interface INestedChannel : IGuildChannel
+    {
+        ValueTask<ICategoryChannel> GetCategoryAsync(StateBehavior stateBehavior = default, RequestOptions? options = default);
+        ulong? CategoryId { get; }
+
+        Task SyncPermissionsAsync(RequestOptions? options = default);
+
+        // review: do we want long, paramaterized methods like this, or would we rather a CreateInviteProperties
+        Task CreateInviteAsync(int? maxAge = 86400,
```
k-boyle: I think a properties class is much cleaner.

src/Discord.Net/Entities/Channels/INewsChannel.cs
```diff
@@ -0,0 +1,9 @@
+// todo: docs
+namespace Discord
+{
+    // review: marker interfaces suck! should we just have an "IsNews" property on ITextChannel?
```
k-boyle: `IsNews` property, yes.

src/Discord.Net/Entities/Channels/ICategoryChannel.cs
```diff
@@ -0,0 +1,8 @@
+// todo: docs
+namespace Discord
+{
+    // review: another marker interface 😕
```
k-boyle: Similar to the `IsNews` property, I think this should also just be a property unless it gets something added to it down the line.

src/Discord.Net/Entities/Channels/ITextChannel.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Discord
+{
+    public interface ITextChannel : IMessageChannel, INestedChannel, ITaggable
+    {
+        string Topic { get; }
+        int SlowModeInterval { get; } // review: int vs int? vs TimeSpan
```
k-boyle: I think TimeSpan is easier to understand and more user-friendly.

src/Discord.Net/Entities/Channels/ITextChannel.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        bool IsNsfw { get; }
+
+        Task ModifyAsync(TextChannelProperties properties, RequestOptions? options = default);
+
+        Task DeleteMessagesAsync(IEnumerable<SnowflakeOrEntity<IMessage>> messages, RequestOptions? options = default);
+
+        Task<IWebhook> CreateWebhookAsync(string name, Stream avatar = default, RequestOptions? options = default);
+        // review: allow webhooks to be held in state? (no gateway event will populate them though)
+        ValueTask<IWebhook> GetWebhookAsync(ulong id, StateBehavior stateBehavior, RequestOptions? options = default);
```
k-boyle: Since there won't be a gateway event to update any changes it's probably better to leave it to the end-user.

src/Discord.Net/Entities/Channels/TextChannelProperties.cs
```diff
@@ -0,0 +1,24 @@
...
+using Voltaic;
+using Model = Wumpus.Requests.ModifyChannelParams;
+
+namespace Discord
+{
+    public class TextChannelProperties : GuildChannelProperties
+    {
+        public Optional<string> Topic { get; set; }
+        public Optional<bool> IsNsfw { get; set; }
+        public Optional<int> SlowModeInterval { get; set; } // review: int vs TimeSpan
```
k-boyle: TimeSpan

src/Discord.Net/Entities/Emotes/EmoteProperties.cs
```diff
@@ -0,0 +1,20 @@
+// TODO: docs
+using Voltaic;
+using Model = Wumpus.Requests.ModifyGuildEmojiParams;
+
+namespace Discord
+{
+    public class EmoteProperties
+    {
+        public Optional<string> Name { get; set; }
+        // review: do we care about RoleIds? can bots even do anything with them?
```
k-boyle: I think bots can assign them?

src/Discord.Net/Entities/Emotes/IAttachedEmote.cs
```diff
@@ -0,0 +1,15 @@
+// TODO: docs
+// TODO: docs
+using System.Threading.Tasks;
+
+namespace Discord
+{
+    // review: is this naming clear?
```
k-boyle: I have no idea what this is for, so no.

src/Discord.Net/Entities/Emotes/IGuildEmote.cs
```diff
@@ -0,0 +1,8 @@
+// TODO: docs
+namespace Discord
+{ 
+    public interface IGuildEmote : IEmote, ISnowflakeEntity
+    {
+        // this is purposefully empty! Name is satisfied by IEmote, Id is satisfied by Snowflake
```
k-boyle: Maybe ditch this interface in preference for a property?

src/Discord.Net/Entities/Guilds/GuildProperties.cs
```diff
@@ -0,0 +1,48 @@
...
+using Wumpus;
+using Wumpus.Entities;
+using Model = Wumpus.Requests.ModifyGuildParams;
+
+namespace Discord
+{
+    public class GuildProperties
+    {
+        public Optional<string> Name { get; set; }
+        public Optional<string> RegionId { get; set; } // review: do we need to wrap VoiceRegion
```
k-boyle: End-users are stupid, so probably.

src/Discord.Net/Entities/Guilds/IGuild.cs
```diff
@@ -0,0 +1,90 @@
...
+    public interface IGuild : ISnowflakeEntity, IDeletable
+    {
+        // todo: IAudioClient
+        string Name { get; }
+        ulong? ApplicationId { get; }
+        ulong EveryoneRoleId { get; }
+        ulong OwnerId { get; }
+
+        int AFKTimeout { get; }
+        bool IsEmbeddable { get; } //review: is this still applicable
```
k-boyle: This is still a thing, yes.

src/Discord.Net/Entities/Guilds/IGuild.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        string? IconId { get; }
+        string? IconUrl { get; }
+        string? SplashId { get; }
+        string? SplashUrl { get; }
+
+        bool Attached { get; } // review: how to make this more consumable
+
+        ulong? AFKChannelId { get; }
+        ulong? DefaultChannelId { get; } // todo: can we expose this synchronously? needs access to channels
```
k-boyle: Since this has been scrapped from the client should probably just leave this up to the user to implement/have an extension method for it.

src/Discord.Net/Entities/Invites/IInvite.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        ValueTask<IChannel> GetChannelAsync(StateBehavior stateBehavior, RequestOptions? options = default);
+
+        string Code => Id; // review: is this worth having
+        string Url { get; }
+
+        ulong ChannelId { get; }
+        string ChannelName { get; }
+
+        ulong? GuildId { get; } // review: do we need to support group DM invites
```
k-boyle: Not unless bots in the future are allowed to join them.

src/Discord.Net/Entities/Invites/IInvite.cs
```diff
@@ -0,0 +1,22 @@
+// todo: docs
+using System.Threading.Tasks;
+
+namespace Discord
+{
+    public interface IInvite : IEntity<string>, IDeletable
+    {
+        ValueTask<IChannel> GetChannelAsync(StateBehavior stateBehavior, RequestOptions? options = default);
+
+        string Code => Id; // review: is this worth having
```
k-boyle: Seems needlessly verbose.

src/Discord.Net/Entities/Messages/ITag.cs
```diff
@@ -0,0 +1,14 @@
+// todo: docs
+using System;
+
+namespace Discord
+{
+    public interface ITag
+    {
+        Index Index { get; } // review: is this big brained? would it be bigger brained to just use a ReadOnlySpan?
```
k-boyle: Spans aren't really user-friendly (imo) so for user-experience probably better to leave as an Index.

src/Discord.Net/Entities/Messages/IUserMessage.cs
```diff
@@ -0,0 +1,24 @@
+// todo: docs
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Discord
+{
+    public interface IUserMessage : IMessage
+    {
+        string JumpUrl { get; } // review: use System.Uri for this?
```
k-boyle: Is there any benefit for having it as a Uri?

src/Discord.Net/Entities/Messages/MessageProperties.cs
```diff
@@ -0,0 +1,22 @@
...
+using Model = Wumpus.Requests.ModifyMessageParams;
+
+namespace Discord
+{
+    public class MessageProperties
+    {
+        public Optional<string> Content { get; set; }
+        public Optional<IEmbed> Embed { get; set; }
+
+        internal Model ToWumpus() // review: internal or external ToWumpus meths?
```
k-boyle: Unless wumpus stuff is exposed I don't see the point.

src/Discord.Net/Entities/Roles/RoleProperties.cs
```diff
@@ -0,0 +1,33 @@
+// todo: docs
+using Voltaic;
+using Wumpus.Entities;
+using Model = Wumpus.Requests.ModifyGuildRoleParams;
+
+namespace Discord
+{
+    public class RoleProperties
```
k-boyle: Is there anything about `IsBoosted` or does that just come under managed?

src/Discord.Net/Entities/Users/IMember.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        DateTimeOffset? JoinedAt { get; }
+        string Nickname { get; }
+        GuildPermissions GuildPermissions { get; }
+
+        ValueTask<ChannelPermissions> GetPermissionsAsync(IGuildChannel channel);
+
+        Task ModifyAsync(MemberProperties properties, RequestOptions? options = default);
+
+        // review: should these two be extensions on IGuild?
```
k-boyle: There should at least be an id version of it on IGuild.

src/Discord.Net/Entities/Users/IUser.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Threading.Tasks;
+using Voltaic;
+
+namespace Discord
+{
+    public interface IUser : ISnowflakeEntity
+    {
+        string Username { get; }
+        string Discriminator { get; }
+        // review: does anyone use DiscriminatorValue? seems like a waste of memory
```
k-boyle: It is a waste, yes.

src/Discord.Net/Entities/Users/MemberProperties.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    public class MemberProperties
+    {
+        public Optional<bool> IsDeafened { get; set; }
+        public Optional<bool> IsMuted { get; set; }
+        public Optional<string> Nickname { get; set; }
+        public Optional<IEnumerable<SnowflakeOrEntity<IRole>>> Roles { get; set; }
+        public Optional<SnowflakeOrEntity<IVoiceChannel>?> VoiceChannel { get; set; }
+
+        public Model ToWumpus()
```
k-boyle: Isn't this meant to be internal?

src/Discord.Net/IDiscordClient.cs
```diff
@@ -0,0 +1,28 @@
...
+        event Action Connected;
+        event Action Disconnected;
+        event Action Ready;
+
+        event Action<IChannel> ChannelCreated;
+        event Action<IChannel> ChannelDeleted;
+
+        event Action<IMessage> MessageCreated;
+        // todo: implement the rest of the events
+        // review: event classes or just Action<A,B,C> types
```
k-boyle: the EventArgs pattern is probably better.

src/Discord.Net/State/IStateProvider.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Discord
+{
+    public interface IStateProvider
+    {
+        // we'll need to extend this as the client is implemented
+
+        ValueTask<IUser> GetUserAsync(ulong id, StateBehavior stateBehavior, RequestOptions? options);
+        // review: IReadOnlyCollection is proper collection type here?
+        ValueTask<IReadOnlyCollection<IUser>> GetUsersAsync(StateBehavior stateBehavior, RequestOptions? options);
```
k-boyle: Yes. As long as the immutable collections aren't used (they're very slow and memory heavy).

src/Discord.Net/Entities/Roles/RoleProperties.cs
foxbot: No, boosted roles don't come with wheels, they are just managed.

examples/0_Using_the_Client/Program.cs
```diff
@@ -0,0 +1,53 @@
+using System;
+using System.Threading.Tasks;
+using Discord;
+
+namespace Sample
+{
+    public class Program
+    {
+        static void Main()
```
monoclex: why not `static async Task Main`

src/Discord.Net/Entities/Channels/ICategoryChannel.cs
JakenVeina: I think it's appropriate to split the different types of entities via the typing system, rather than simple properties, when the set of functionality for that entity diverges. I.E. when there are operations that can be performed only upon a category channel, and operations that cannot be performed upon a category channel, even if they don't appear directly upon the interface. Such operations can then be defined to operate specifically upon `ICategoryChannel`, which provides a more self-documented, foolproof API.

If no such operations exist for `ICategoryChannel` then it probably makes more sense as just a property.

src/Discord.Net/Entities/Channels/IMessageChannel.cs
JakenVeina: Or as simple as `SendFileOptions`. Make it a `struct` if you're terribly worried about allocations.

src/Discord.Net/Entities/Channels/INestedChannel.cs
JakenVeina: I also lean towards `CreateInviteProperties` or `CreateInviteOptions`.

src/Discord.Net/Entities/Channels/INewsChannel.cs
JakenVeina: Same comment as above. If there is no functional/operational difference in the API, a property is simpler.

src/Discord.Net/Entities/Channels/ITextChannel.cs
JakenVeina: Seconded.

src/Discord.Net/Entities/Emotes/EmoteProperties.cs
JakenVeina: https://discordapp.com/developers/docs/resources/emoji#modify-guild-emoji

Unless this PATCH call is specifically restricted from bot users.

src/Discord.Net/Entities/Emotes/IAttachedEmote.cs
JakenVeina: Would this be a custom emoji added to the guild, as opposed to a built-in one? If so, maybe `ICustomEmote` might work better?

src/Discord.Net/Entities/Emotes/IGuildEmote.cs
JakenVeina: What classes implement `IEmote`, but not `ISnowflakeEntity` then? Alternatively, what methods can operate only on `IGuildEmote`?

src/Discord.Net/Entities/Guilds/GuildProperties.cs
```diff
@@ -0,0 +1,48 @@
...
+        public Optional<int> AfkTimeout { get; set; }
```
JakenVeina: Another candidate for a `TimeSpan`.

src/Discord.Net/Entities/Guilds/IGuild.cs
```diff
@@ -0,0 +1,90 @@
...
+        MfaLevel MfaLevel { get; }
+        VerificationLevel VerificationLevel { get; }
+        ExplicitContentFilter ExplicitContentFilter { get; }
+
+        string? IconId { get; }
+        string? IconUrl { get; }
+        string? SplashId { get; }
+        string? SplashUrl { get; }
+
+        bool Attached { get; } // review: how to make this more consumable
```
JakenVeina: How do you mean?

src/Discord.Net/Entities/ICacheable.cs
```diff
@@ -0,0 +1,12 @@
+using System.Threading.Tasks;
+
+namespace Discord
+{
+    public interface ICacheable<TEntity> : ISnowflakeEntity
```
JakenVeina: Would be cool with keeping this as a `struct` so long as `public` constructors are added, so mocks/fakes can be injected in.

src/Discord.Net/Entities/IEntity.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+
+namespace Discord
+{
+    public interface IEntity<TId>
```
JakenVeina: This always kinda bugged me: are there any actual cases where this is used with `TId` as something other than `ulong`?

src/Discord.Net/Entities/IEntity.cs
```diff
@@ -0,0 +1,18 @@
...
+        where TId : IEquatable<TId>
+    {
+        /// <summary>
+        ///     Gets the <see cref="IDiscordClient"/> that created this object.
+        /// </summary>
+        IDiscordClient Discord { get; }
```
JakenVeina: Would `Owner` maybe be a better name?

src/Discord.Net/Entities/Invites/IInvite.cs
JakenVeina: Is the name `Code` used in the context of an invite anywhere in the official docs?

src/Discord.Net/Entities/Messages/IReactionInfo.cs
```diff
@@ -0,0 +1,9 @@
+// todo: docs
+namespace Discord
+{
+    public interface IReactionInfo
+    {
+        int Count { get; }
+        bool Me { get; }
```
JakenVeina: Personal preference, but `IsMe` or in this case maybe `IncludesMe` strikes me as more readable in context.

src/Discord.Net/Entities/Messages/IUserMessage.cs
JakenVeina: `System.Uri` enforces a good deal of validation rules. Are we sure all of those are applicable here?

src/Discord.Net/Entities/Messages/MessageProperties.cs
JakenVeina: Tough call. The vast majority of consumers won't care, but it could be really handy for anyone wanting to extend/enhance the library. Or even for temporary workarounds for bugs or new features.

src/Discord.Net/Entities/Roles/IRole.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        string Name { get; }
+        Color Color { get; }
+        int Position { get; }
+
+        bool IsHoisted { get; }
+        bool IsManaged { get; }
+        bool IsMentionable { get; }
+
+        GuildPermissions Permissions { get; } // review: use wump permissions directly or wrap?
```
JakenVeina: The prior permissions wrapper didn't bug me all that much, but it didn't strike me as that great either. Using the raw `enum` seems much simpler and much more approachable for consumers (as it's a core language feature). Discord.NET can always add some extension methods for reading/manipulating the values if desired.

src/Discord.Net/Entities/Users/IMember.cs
JakenVeina: That thought should apply to almost all API functionality. Except, perhaps, when there should be a "does this ID exist" check, in which case performing the `Get()` call serves that purpose.

src/Discord.Net/DiscordClientBuilder.cs
```diff
@@ -0,0 +1,11 @@
+namespace Discord
+{
+    public sealed class DiscordClientBuilder
```
Chris-Johnston: What will this do different from setting properties in DiscordClientConfig?

src/Discord.Net/Entities/Activities/IActivity.cs
```diff
@@ -0,0 +1,10 @@
+// todo: meth
```
Chris-Johnston: Don't think that's good for your health.

src/Discord.Net/DiscordClientConfig.cs
```diff
@@ -0,0 +1,13 @@
+namespace Discord
+{
+    public class DiscordClientConfig
+    {
+        public string? Token { get; set; }
```
Chris-Johnston: Does this mean that TokenType is dead, and will be assumed based on the type of client that is used?

src/Discord.Net/Entities/Users/IUser.cs
JakenVeina: I can find no usage of it that I wouldn't be happy to replace.

src/Discord.Net/DiscordClientConfig.cs
```diff
@@ -0,0 +1,13 @@
...
+
+        public int Shard { get; set; } = 0;
```
Chris-Johnston: Are the sharded client and websocket client going to be merged?

src/Discord.Net/Entities/Channels/INestedChannel.cs
Chris-Johnston: So, should we do the same for stuff like ChannelPermissions, lest we forget the best url: https://docs.stillu.cc/api/Discord.ChannelPermissions.html#Discord_ChannelPermissions__ctor_System_Boolean_System_Boolean_System_Boolean_System_Boolean_System_Boolean_System_Boolean_System_Boolean_System_Boolean_System_Boolean_System_Boolean_System_Boolean_System_Boolean_System_Boolean_System_Boolean_System_Boolean_System_Boolean_System_Boolean_System_Boolean_System_Boolean_System_Boolean_System_Boolean_

src/Discord.Net/Entities/Channels/INewsChannel.cs
Chris-Johnston: So what happens when discord decides to add (more) special functionality to news channels? (they can switch between news and text types) Or, are we still sticking to treating them just like text channels and not supporting them (like store channels)?

src/Discord.Net/Entities/Channels/ITextChannel.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Discord
+{
+    public interface ITextChannel : IMessageChannel, INestedChannel, ITaggable
+    {
+        string Topic { get; }
+        int SlowModeInterval { get; } // review: int vs int? vs TimeSpan
+        // present behavior: 0 = no slow mode, > 1 = yes slow mode
```
Chris-Johnston: IsSlowMode prop/extension method?

src/Discord.Net/Entities/Webhooks/IWebhook.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Discord
+{
+    public interface IWebhook
+    {
+        ValueTask<IUser> GetCreatorAsync(StateBehavior stateBehavior = default, RequestOptions? options = default);
+        ValueTask<ITextChannel> GetChannelAsync(StateBehavior stateBehavior = default, RequestOptions? options = default);
+        ValueTask<IGuild> GetGuildAsync(StateBehavior stateBehavior = default, RequestOptions? options = default);
+        // review: does this need to be an async retrieval? the only path to get an IWebhook should be via an ITextChannel
+
```
JakenVeina: Doesn't hurt to be consistent. The performance impact should be negligible, particularly when using `ValueTask`.

src/Discord.Net/Entities/Webhooks/IWebhook.cs
```diff
@@ -0,0 +1,25 @@
...
+        string Token { get; }
+
+        // note: yeah, the full creator is included in the payload, but we don't want entities to hold direct references to other entities
```
JakenVeina: I can get behind that. So long as there's a hot path optimization within `GetCreatorAsync()` if there's potential for an `async` call when fresh data has just come in on the same message.

src/Discord.Net/Entities/Channels/ITextChannel.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        bool IsNsfw { get; }
+
+        Task ModifyAsync(TextChannelProperties properties, RequestOptions? options = default);
+
+        Task DeleteMessagesAsync(IEnumerable<SnowflakeOrEntity<IMessage>> messages, RequestOptions? options = default);
+
+        Task<IWebhook> CreateWebhookAsync(string name, Stream avatar = default, RequestOptions? options = default);
```
Chris-Johnston: Similar to the overload for SendMessageAsync, should this have an overload for the avatar file path?

src/Discord.Net/Entities/Guilds/GuildProperties.cs
Chris-Johnston: timespan?

src/Discord.Net/Entities/Guilds/IGuild.cs
```diff
@@ -0,0 +1,90 @@
...
+{
+    public interface IGuild : ISnowflakeEntity, IDeletable
+    {
+        // todo: IAudioClient
+        string Name { get; }
+        ulong? ApplicationId { get; }
+        ulong EveryoneRoleId { get; }
+        ulong OwnerId { get; }
+
+        int AFKTimeout { get; }
```
Chris-Johnston: timespan?

src/Discord.Net/Entities/Guilds/IGuild.cs
```diff
@@ -0,0 +1,90 @@
...
+        ValueTask<IMember?> GetMemberAsync(ulong id, StateBehavior stateBehavior = default, RequestOptions? options = default);
+        Task<IMember> AddMemberAsync(ulong id, string accessToken, MemberProperties? properties = default, RequestOptions? options = default);
+        Task<int> PruneMembersAsync(int days = 30, bool simulate = false, RequestOptions? options = default);
+        Task DownloadMembersAsync(); // review: do we want to keep this
+        // extensions -- review: should move to an extensions class? see above
+        ValueTask<IMember> GetOwnerAsync(StateBehavior stateBehavior = default, RequestOptions? options = default);
+
+        ValueTask<IReadOnlyCollection<IAttachedEmote>> GetEmotesAsync(StateBehavior stateBehavior = default, RequestOptions? options = default);
+        ValueTask<IAttachedEmote> GetEmoteAsync(StateBehavior stateBehavior = default, RequestOptions? options = default);
+        Task CreateEmoteAsync(string name, Image image, RequestOptions? options = default);
```
Chris-Johnston: Should this consume an EmoteBuilder instead?

src/Discord.Net/Entities/Webhooks/IWebhook.cs
```diff
@@ -0,0 +1,25 @@
...
+        ulong CreatorId { get; } 
+        ulong ChannelId { get; }
+        ulong GuildId { get; } // review: this is impl'd as a ulong? in d.net legacy, todo-ensure there is no case we wouldn't have the guild ID
```
JakenVeina: This sounds eerily familiar.

After looking back through our issue tracker...

We had a specific issue with an NRE coming out of webhook users, when trying to access `.GuildId`. Tracked it down to the fact that both `PRESENCE_UPDATE` and `GUILD_MEMBER_UPDATE` were being presented through `DiscordSocketClient` as `GuildMemberUpdated`. The problematic part was that `PRESENCE_UPDATE` by definition, does not include the full details of the user, just its ID, so the old client attempts to grab the full user from cache to present through `GuildMemberUpdated`. For webhook users, however, there would never be any cache entry, and the client would just build up an `IWebhookUser` with the `Guild` left as `null`.

I would definitely support making this not-nullable, if possible.

src/Discord.Net/Entities/Invites/IInviteMetadata.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Discord
+{
+    public interface IInviteMetadata
+    {
+        ValueTask<IUser> GetInviterAsync(StateBehavior stateBehavior = default, RequestOptions? options = default);
+        ulong InviterId { get; }
+
+        bool IsRevoked { get; }
+        bool IsTemporary { get; }
+        int? MaxAge { get; }
```
Chris-Johnston: Timespan?

src/Discord.Net/Entities/Roles/IRole.cs
Chris-Johnston: I don't see any point in wrapping imo

src/Discord.Net/IDiscordClient.cs
JakenVeina: `EventArgs` or similar please. For one, we do basically that anyway, in a wrapper layer around the client. For two, it's WAY better from a documentation standpoint. For example, `GuildMemberUpdated` today has a signature of 

```cs
public event Func<SocketGuildUser, SocketGuildUser, Task> GuildMemberUpdated;
```

If I don't already know how this event works, and want to know, which user is the "old" version and which is the "new" one, the only way to really know for sure is to go dig through the source on github and see that when it's actually invoked, it's...

```cs
await TimedInvokeAsync(_guildMemberUpdatedEvent, nameof(GuildMemberUpdated), before, user).ConfigureAwait(false);
```

src/Discord.Net/Entities/Users/IMember.cs
```diff
@@ -0,0 +1,38 @@
...
+        ValueTask<IUser> GetUserAsync(StateBehavior stateBehavior = default, RequestOptions? options = default);
+        ValueTask<IGuild> GetGuildAsync(StateBehavior stateBehavior = default, RequestOptions? options = default);
+        ValueTask<IReadOnlyCollection<IRole>> GetRoles(StateBehavior stateBehavior = default, RequestOptions? options = default);
+
+        ulong GuildId { get; }
+        IReadOnlyCollection<ulong> RoleIds { get; }
+        IMemberPresence Presence { get; }
+
+        DateTimeOffset? JoinedAt { get; }
+        string Nickname { get; }
```
Chris-Johnston: nit: thoughts on a DisplayName extension that does `Nickname ?? Username`, since it's asked for so often

src/Discord.Net/IDiscordClient.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Discord
+{
+    public interface IDiscordClient
+    {
+        event Action Connected;
+        event Action Disconnected;
+        event Action Ready;
+
+        event Action<IChannel> ChannelCreated;
```
JakenVeina: Are events no longer asynchronous? If not, presumably we become responsible for dispatching the event to a new `async` context, if desired?

I'm totally cool with that, since that's what we do currently anyway, but it might be a good idea to have a built-in mechanism to do said dispatching, for beginner consumers. Something like an injectable `IDiscordEventDispatcher`, with a couple built-in implementations?

src/Discord.Net/State/ConcurrentMemoryStateProvider.cs
```diff
@@ -0,0 +1,104 @@
+// todo: impl, docs
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Discord
+{
+    internal class ConcurrentMemoryStateProvider : IStateProvider
```
JakenVeina: This makes me smile.

src/Discord.Net/State/IStateProvider.cs
JakenVeina: Not sure I agree with that assessment of immutables, in general, but I definitely agree that `IReadOnlyCollection` is the best bet here. Immutables would be appropriate if you KNOW your consumer is going to be doing a lot of collection manipulation, and also wants immutability.

The other option would be `IReadOnlyList` but do we really expect consumers to want indexing by position?

src/Discord.Net/State/StateBehavior.cs
```diff
@@ -0,0 +1,47 @@
...
+        ///     This flag is used to indicate that the retrieval of this entity should not leave the
+        ///     synchronous path of the <see cref="System.Threading.Tasks.ValueTask"/>. When true,
+        ///     the calling method *should* not ever leave the calling task, and never generate an async
+        ///     state machine.
+        ///
+        ///     Bear in mind that the true behavior of this flag depends entirely on the <see cref="IStateProvider"/> to
+        ///     abide by design implications of this flag. Once Discord.Net has called out to the state provider with this
+        ///     flag, it is out of our control whether or not an async method is evaluated.
+        /// </remarks>
+        SyncOnly = 1,
```
JakenVeina: Possible need for clarification: What's the difference between `SyncOnly` and `CacheOnly`?

JakenVeina (commented):
Thanks a ton for the opportunity to comment.

src/Discord.Net/Entities/Activities/IActivity.cs
JakenVeina: Unless you have lung cancer.

src/Discord.Net/DiscordClientBuilder.cs
foxbot: Nothing, it was just created because (prior to using c#8) there was no other way to create an IDiscordClient. I just tested, and we can now just have a static `IDiscordClient.FromConfig(conf)` method, or whatever naming seems more elegant.

src/Discord.Net/DiscordClientConfig.cs
foxbot: Not specifically, I just hadn't implemented it yet. Though, this client will only support bot tokens, as it makes more sense for OAuth users to just use Wumpus directly.

src/Discord.Net/Entities/Users/IMember.cs
JakenVeina: Seconded.

src/Discord.Net/DiscordClientConfig.cs
foxbot: Since we're just wrapping over a WumpusGatewayClient and not actually implementing any client logic ourselves here, I think it seems reasonable to allow an IDiscordClient to support multiple shards.

src/Discord.Net/Entities/Guilds/IGuild.cs
foxbot: EmoteBuilder is a static class for constructing IEmote/IGuildEmote from strings; see above comment regarding DiscordClientBuilder. I'll probably just replace this with IEmote.Parse/IGuildEmote.Parse

src/Discord.Net/Entities/Users/IMember.cs
foxbot: sounds good to me, I guess we can just make this a default interface implementation, since it doesn't look like extension everything is going to make the cut for c#8

src/Discord.Net/Entities/Channels/ICategoryChannel.cs
foxbot: `ICategoryChannel` is used as a `SnowflakeOrEntity` when setting modifying another channel's property, so there is a weak compile-time check against the channel type (since we accept a plain ID in these properties, a user could always pass the ID of a non-category channel). 

I'm hesitant to add booleans for non-standard channel types, which are presently limited to only News, Store, and LFG. I would like to avoid property bloat where possible, however using an `enum` on `IGuildChannel` also feels limiting - as we would either have an enum that is only used for three obscure channel types and no others, or we would be duplicating information - e.g. a text channel could be named either via `ChannelType.Text` or `channel is ITextChannel`.

That being said, if the general consensus is that a property set or enum would be better design, that change can be easily made.

src/Discord.Net/Entities/Emotes/IAttachedEmote.cs
foxbot: `IAttachedEmote` is used to distinguish a user-created `IGuildEmote` from the API created variant. In 1.x, we supported creating guild emotes from their mention string ("<:foxbot_dog:560278330969030656>"), which does not carry the additional metadata (such as RequireColons or RoleIds) that Discord sends when retrieving the emote from the API.

Maybe it would be better design in 3.0 to require using `IGuild.GetEmoteAsync(emoteId)` to always fetch the emote from cache, rather than allowing the user to create an emote from a mention string, and just assuming that they gave us valid data. I would lean against this, though, because when writing bots, emotes are generally hard-coded (i.e., "I know that I have an emote with ID xxx", rather than "I want to find the emote with ID xxx or name yyy from guild zzz").

src/Discord.Net/Entities/Emotes/IGuildEmote.cs
foxbot: `IEmote` is a wrapper around a standard emoji for use in reactions. It's more clear when invoking `AddReactionAsync(IEmote)` to use an emoji than if just presented with `AddReaction(string)` - where the value of the `string` would have to be inferred from documentation.

Ideally we will validate the string passed to `IEmote.Create/Parse` as being valid unicode, though I don't know if I want the library to take on the burden of being the emoji 👮‍♂️.

src/Discord.Net/Entities/Invites/IInvite.cs
foxbot: Yes, invites don't actually have an `Id` property, as their `Code` property serves as the ID.

The `Id` property comes from `IEntity<string>`

examples/0_Using_the_Client/Program.cs
foxbot: I recall having problems with how exceptions were handled when using async main directly when the feature came out with c#7. That issue might be fixed by now, I'm just used to doing it the old way

src/Discord.Net/Entities/Guilds/IGuild.cs
foxbot: I don't know that `Attached` makes much sense to end-users, nor is really relevant to them. This property is `true` when the guild is active (i.e., the client is receiving events for) in the WebSocket connection, `false` when the guild hasn't been streamed yet or is unavailable because Discord is down.

src/Discord.Net/Entities/IEntity.cs
foxbot: `IInvite` is the only case that comes to mind

src/Discord.Net/Entities/IEntity.cs
foxbot: `Owner` is ambiguous with the owner of an actual entity (guild owner, message author/owner), though `Client` might also work

src/Discord.Net/IDiscordClient.cs
foxbot: Yes, users are now responsible for dispatching to an async context. (see the included example)

I'm not familiar with what you mean by providing something like an `IDiscordEventDispatcher`, if you could link me to an example of how that might look I would be happy to include it.

src/Discord.Net/State/StateBehavior.cs
foxbot: `SyncOnly` requests (can't enforce) that a state provider never generates an async-state machine, since all state lookups are ValueTasks.

`CacheOnly` requests that a state provider doesn't contact Discord when retrieving an entity.

The motivation for implementing this might be if you're using an additional caching layer for long-term storage, such as redis, where a state provider might need to make an async call to retrieve information. If you're in a context where the ValueTask must remain synchronous (maybe in a TypeReader, or if you're `.Result`'ing), you could use SyncOnly.

src/Discord.Net/Entities/Guilds/IGuild.cs
JakenVeina: Seems like it COULD be useful, but probably only if it's named something a little more obvious like `IsAttached` and comes along with `Attached` and `Detached` events. Or maybe `IsConnected`?

src/Discord.Net/Entities/IEntity.cs
JakenVeina: Fair enough, then. If not for that, I'd say remove it as a pointless abstraction.

src/Discord.Net/Entities/Messages/IUserMessage.cs
AntiTcb: We already went down that rabbit hole of swapping some of the appropriate things to `System.Uri` and general consensus was not a fan of it.

src/Discord.Net/IDiscordClient.cs
JakenVeina: In my mind, the ideal way to handle discord events is the same way ASP.NET handles HTTP requests: each incoming message spawns a new "logical" thread. I.E. a new `IServiceScope` in the DI system, and a spawned `Task` that is simply not `await`ed, so that it runs in parallel to other "logical" threads.

For the current Discord.NET version, we do this with several pieces fitting together:
https://github.com/discord-csharp/MODiX/blob/master/Modix.Common/Messaging/MessageDispatcher.cs
https://github.com/discord-csharp/MODiX/blob/master/Modix.Common/Messaging/INotificationHandler.cs
https://github.com/discord-csharp/MODiX/blob/master/Modix.Services/Core/DiscordSocketListeningBehavior.cs

You can see how subscribing to each event we need, wrapping up the args into an object, and then handing off the work to the dispatcher is a tad tedious.

An `IDiscordEventDispatcher` could look like the following...

```cs
public interface IDiscordEventDispatcher
{
    Task DispatchEvent(IEnumerable<Func<Task>> eventHandlers);
}
```

Where you could have the following built-in implementations that provide behavior similar to `Discord.Commands.RunMode`.

```cs
public class SyncDiscordEventDispatcher
    : IDiscordEventDispatcher
{
    public Task DispatchEvent(IEnumerable<Func<Action>> eventHandlers)
    {
        foreach(var eventHandler in eventHandlers)
            await eventHandler.Invoke();
    }
}
```

```cs
public class ParallelDiscordEventDispatcher
    : IDiscordEventDispatcher
{
    public Task DispatchEvent(IEnumerable<Func<Task>> eventHandlers)
    {
        foreach(var eventHandler in eventHandlers)
            _ = eventAction.Invoke();
        return Task.CompletedTask;
    }
}
```

You can throw in some basic error handling around the `Task`s that aren't `await`ed, or leave that to the consumer.

In our case, we'd implement a custom dispatcher that also handles DI scoping, since Discord.NET doesn't have any built-in DI. Which, actually, now that I think about it, wouldn't fit with the above interface definition.

So, if you're interested at all, it definitely needs some brainstorming. But, I think the concept of handling messages the way ASP.NET handles HTTP requests would be appealing to a lot of consumers.

src/Discord.Net/State/StateBehavior.cs
JakenVeina: That's an excellent thought. You should probably just take that last paragraph and inject it into the XML.

src/Discord.Net/Entities/Messages/ITag.cs
Joe4evr: What purpose can a Span even serve to replace an Index?

src/Discord.Net/Entities/Messages/IUserMessage.cs
Joe4evr: Of note: Even the CoreFX API team doesn't like `System.Uri`.

src/Discord.Net/Entities/Guilds/IGuild.cs
Joe4evr: Agree that at least the name needs a change: Design guidelines say that boolean property/method names should begin with a verb.

src/Discord.Net/Entities/Users/IUser.cs
NovusTheory: Would it be better to make this a ushort and people can just ToString() it if needed? I feel like it could stay consistent since discriminators are numbers.

src/Discord.Net/Entities/Guilds/IGuild.cs
Chris-Johnston:
```suggestion
        bool IsAttached { get; } // review: how to make this more consumable
```
I think attached (with proper docs explaining what it is later on) is less ambiguous than `IsConnected`

src/Discord.Net/Entities/Users/IUser.cs
Chris-Johnston: I think that it would be most appropriate to keep Discriminators strings to match the API: https://discordapp.com/developers/docs/resources/user#user-object
There's less risk of things breaking should they change what discriminators can be.

src/Discord.Net/Entities/Channels/ICategoryChannel.cs
JakenVeina: Can't say either way in particular has sold me. The arguments on both side are valid, I think. The amount of type forking seems to be one of the less-popular design choices of previous versions, but as others have mentioned, it offers more flexibility for future functionality. I.E. not a breaking change to add a method to an empty interface, or to IGuild.

src/Discord.Net/Entities/Emotes/IAttachedEmote.cs
JakenVeina: So, it's kinda the same concept as "attached" on other interfaces, in that an "attached" emote has all the latest metadata?

If so, maybe "Synced" makes more sense, across the board?

src/Discord.Net/Entities/Emotes/IGuildEmote.cs
JakenVeina: So, an `IEmote` could just be an emote ID that's been parsed, without any validation of whether it's a real entity. It's really more of a model than an entity. Maybe there's a naming or structural change that can make that clearer? `IEmoteIdentifier`? or `IEmoteId`? Or maybe just make it a concrete class/struct and `IGuildEmote` has it as a property.

src/Discord.Net/Entities/Invites/IInvite.cs
JakenVeina: That suggests to me that `IInvite` should not extend `IEntity<T>`.

src/Discord.Net/Entities/Emotes/EmoteProperties.cs
PassiveModding: bots are able to assign them

src/Discord.Net/Entities/IEntity.cs
k-boyle: I second renaming it to `Client`

src/Discord.Net/Entities/Channels/IMessageChannel.cs
xSke: Can we have a way to upload multiple files per message, too? I know this is supported (mobile clients do it!). Perhaps an overload taking an `IEnumerable<SendFileOptions>`.

src/Discord.Net/IDiscordClient.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using System.Threading.Tasks;
+using Wumpus;
+
+namespace Discord
+{
+    public interface IDiscordClient
+    {
+        event Action Connected;
```
xSke: What's the logic behind having these defined as `event`s? Personally, I'd much prefer something like exposing the `AsyncEvent<T>` fields from v2 (which are private, and used internally anyway). This way, you can work with events as first-class objects, and e.g. do something like `public async Task<T> WaitFor(this AsyncEvent<T>)` like [in discord.py](https://discordpy.readthedocs.io/en/latest/api.html#discord.Client.wait_for) - super useful for making interactive stuff.

csmir:
@quinchs are there any plans to act on this PR in the future?

PR #1348, opened 2019-07-25
Add Quote Formatting

Chris-Johnston:
Adds support for block quote text formatting (with unit tests). This feature is currently only implemented in the Canary client. This formatting adds a `> ` to each new line from the input text.

I don't recall if Discord cares about CRLF vs LF line endings, so I just split on LF endings.

This does not check to see that the input text has been already quoted.

### DIFF
--- a/src/Discord.Net.Core/Format.cs
+++ b/src/Discord.Net.Core/Format.cs
@@ -47,8 +47,10 @@ namespace Discord
         /// <returns>Gets the formatted quote text.</returns> // TODO: better xmldoc
         public static string Quote(string text)
         {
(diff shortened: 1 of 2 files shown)


Joe4evr:
> This feature is currently only implemented in the Canary client.

If this was true then why does it work on non-Canary?

Chris-Johnston:
Oh, I guess it's on the normal client too.

foxbot (approved):
Looks good. This format has changed at least once already, so I'm going to wait a few more days before merging in.

foxbot (approved):

PR #1354, opened 2019-08-14
Add support for setting X-RateLimit-Precision

Chris-Johnston:
Implements feature described in https://github.com/discordapp/discord-api-docs/pull/1064

Adds an option to specify `RateLimitPrecision` in `DiscordConfig` so that the `X-RateLimit-Reset` response header can be more precise. This feature is not yet deployed (from what I have found by hitting ratelimits).

https://github.com/discordapp/discord-api-docs/pull/1069 Adds X-RateLimit-Reset-After, which instead of specifying the timestamp when ratelimits are reset, specifies the amount of time to delay. This comes with a caveat:

> network latency will cause you to be less efficient than using X-RateLimit-Reset with a synchronized clock source.

Should we even bother with `X-RateLimit-Reset-After`?

Also, when converting from the `RateLimitPrecision` to a string for the header:

https://github.com/discord-net/Discord.Net/blob/0c7d59040803ff12319579e19512c0c6cf76f882/src/Discord.Net.Rest/DiscordRestApiClient.cs#L76

Do we think that `.ToString().ToLower()` is fine, or should we add an extension or util method to convert `RateLimitPrecision` into strings?

### DIFF
--- a/src/Discord.Net.Rest/DiscordRestApiClient.cs
+++ b/src/Discord.Net.Rest/DiscordRestApiClient.cs
@@ -45,17 +45,19 @@ namespace Discord.API
         internal string AuthToken { get; private set; }
         internal IRestClient RestClient { get; private set; }
         internal ulong? CurrentUserId { get; set; }
-
+        public RateLimitPrecision RateLimitPrecision { get; private set; }
+        
         internal JsonSerializer Serializer => _serializer;
 
         /// <exception cref="ArgumentException">Unknown OAuth token type.</exception>
         public DiscordRestApiClient(RestClientProvider restClientProvider, string userAgent, RetryMode defaultRetryMode = RetryMode.AlwaysRetry,
-            JsonSerializer serializer = null)
+            JsonSerializer serializer = null, RateLimitPrecision rateLimitPrecision = RateLimitPrecision.Second)
         {
             _restClientProvider = restClientProvider;
             UserAgent = userAgent;
             DefaultRetryMode = defaultRetryMode;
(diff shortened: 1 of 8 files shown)


Joe4evr:
> Do we think that `.ToString().ToLower()` is fine, or should we add an extension or util method to convert `RateLimitPrecision` into strings?

If this is used *a lot*, I'd say to write a util method so the resulting strings are cached.

src/Discord.Net.Rest/DiscordRestApiClient.cs
```diff
@@ -71,6 +73,7 @@ internal void SetBaseUrl(string baseUrl)
             RestClient.SetHeader("accept", "*/*");
             RestClient.SetHeader("user-agent", UserAgent);
             RestClient.SetHeader("authorization", GetPrefixedToken(AuthTokenType, AuthToken));
+            RestClient.SetHeader("X-RateLimit-Precision", RateLimitPrecision.ToString().ToLower());
```
FiniteReality, line 76: As this conversion only occurs here at the moment, I am alright with it remaining as `ToString().ToLower()` - however, if in the future we decide (for some reason) to allow this on a per-request basis, we should definitely move to an extension method, a la `RateLimitPrecision.AsHeaderValue()`

FiniteReality (approved):

AntiTcb:
Is there any value in not having milliseconds be the default?

Chris-Johnston:
> Is there any value in not having milliseconds be the default?

I figured that since the API uses Seconds as the default, I'll stick to that. I don't have any strong opinion on this, though, so either works.

foxbot (approved):

PR #1355, opened 2019-08-14
Provide ParameterInfo with error ParseResult

adamgauthier:
### Description

Currently, when handling parsing errors, there is no way to know what
parameter caused the error. This change makes the `CommandParser` create
the parsing error with the current parameter info when `ParseAsync()`
fails. It is then available through the `ErrorParamInfo` of the
`ParseResult`.

### Example usage

This is just an example of what sort of thing this enables. 
```csharp
public Task OnCommandExecuted(Discord.Optional<CommandInfo> optCommandInfo, ICommandContext context, IResult result)
{
    if (!result.IsSuccess && optCommandInfo.IsSpecified)
    {
        switch (result)
        {
            case ParseResult parseResult when parseResult.ErrorParamInfo != null:
                System.Console.WriteLine($"Parsing error for command {optCommandInfo.Value.Name} and param {parseResult.ErrorParamInfo.Name}: {parseResult.ErrorReason}.");
                break;
        }
    }
    return Task.CompletedTask;
}
```

### Compatibility

I didn't touch the signature of any `.FromError()` and added new overrides for specifying this new property, otherwise, it is null by default. This makes it so it doesn't break any current code but new code can use the new property where appropriate.

This is my first contribution to this repository, if I did anything wrong, please tell me. 😃

### DIFF
--- a/src/Discord.Net.Commands/Results/ParseResult.cs
+++ b/src/Discord.Net.Commands/Results/ParseResult.cs
@@ -18,30 +18,40 @@ namespace Discord.Commands
         /// <inheritdoc/>
         public string ErrorReason { get; }
 
+        /// <summary>
+        ///     Provides information about the parameter that caused the parsing error.
+        /// </summary>
+        /// <returns>
+        ///     A <see cref="ParameterInfo" /> indicating the parameter info of the error that may have occurred during parsing; 
+        ///     <c>null</c> if the parsing was successful or the parsing error is not specific to a single parameter.
+        /// </returns>
+        public ParameterInfo ErrorParameter { get; }
+
         /// <inheritdoc/>
         public bool IsSuccess => !Error.HasValue;
 
-        private ParseResult(IReadOnlyList<TypeReaderResult> argValues, IReadOnlyList<TypeReaderResult> paramValues, CommandError? error, string errorReason)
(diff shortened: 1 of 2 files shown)


src/Discord.Net.Commands/Results/ParseResult.cs
```diff
@@ -18,30 +18,33 @@ public struct ParseResult : IResult
         /// <inheritdoc/>
         public string ErrorReason { get; }
 
+        public ParameterInfo ErrorParamInfo { get; }
```
foxbot: Try to avoid curt abbreviations in property names. I think this could be acceptable as just "ErrorParameter" or "Parameter"

Also, add a documentation comment specifying what this property does, and that it may not always be specified.

foxbot (changes requested):
Looks good, thanks for your contribution! Just one naming concern I'd like addressed before this is merged.

src/Discord.Net.Commands/Results/ParseResult.cs
adamgauthier: Hey you're right, `ErrorParameter` is a better name. I have made the change and added a documentation comment, tell me if everything looks good.

foxbot (approved):

PR #1357, opened 2019-08-16
Added the "Stream" permission.

RheaAyase:
I hope that I didn't miss anything.

### DIFF
--- a/src/Discord.Net.Core/Entities/Permissions/ChannelPermissions.cs
+++ b/src/Discord.Net.Core/Entities/Permissions/ChannelPermissions.cs
@@ -13,7 +13,7 @@ namespace Discord
         /// <summary> Gets a <see cref="ChannelPermissions"/> that grants all permissions for text channels. </summary>
         public static readonly ChannelPermissions Text = new ChannelPermissions(0b01100_0000000_1111111110001_010001);
         /// <summary> Gets a <see cref="ChannelPermissions"/> that grants all permissions for voice channels. </summary>
-        public static readonly ChannelPermissions Voice = new ChannelPermissions(0b00100_1111110_0000000010100_010001);
+        public static readonly ChannelPermissions Voice = new ChannelPermissions(0b00100_1111110_0000000011100_010001);
         /// <summary> Gets a <see cref="ChannelPermissions"/> that grants all permissions for category channels. </summary>
         public static readonly ChannelPermissions Category = new ChannelPermissions(0b01100_1111110_1111111110001_010001);
         /// <summary> Gets a <see cref="ChannelPermissions"/> that grants all permissions for direct message channels. </summary>
@@ -82,6 +82,8 @@ namespace Discord
         public bool UseVAD => Permissions.GetValue(RawValue, ChannelPermission.UseVAD);
         /// <summary> If <c>true</c>, a user may use priority speaker in a voice channel. </summary>
         public bool PrioritySpeaker => Permissions.GetValue(RawValue, ChannelPermission.PrioritySpeaker);
+        /// <summary> If <c>true</c>, a user may stream video in a voice channel. </summary>
+        public bool Stream => Permissions.GetValue(RawValue, ChannelPermission.Stream);
 
         /// <summary> If <c>true</c>, a user may adjust role permissions. This also implictly grants all other permissions. </summary>
         public bool ManageRoles => Permissions.GetValue(RawValue, ChannelPermission.ManageRoles);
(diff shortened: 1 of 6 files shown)


src/Discord.Net.Core/Entities/Permissions/GuildPermissions.cs
```diff
@@ -70,7 +70,9 @@ public struct GuildPermissions
         /// <summary> If <c>true</c>, a user may use voice-activity-detection rather than push-to-talk. </summary>
         public bool UseVAD => Permissions.GetValue(RawValue, GuildPermission.UseVAD);
         /// <summary> If True, a user may use priority speaker in a voice channel. </summary>
-        public bool PrioritySpeaker => Permissions.GetValue(RawValue, ChannelPermission.PrioritySpeaker);
+        public bool PrioritySpeaker => Permissions.GetValue(RawValue, GuildPermission.PrioritySpeaker);
```
RheaAyase, line 73: Oh this is probably worth highlighting. Thoughts?

Chris-Johnston: Values are the same, but yeah that's a good catch.

Chris-Johnston (approved):

src/Discord.Net.Core/Entities/Permissions/ChannelPermissions.cs
```diff
@@ -82,6 +82,8 @@ public static ChannelPermissions All(IChannel channel)
         public bool UseVAD => Permissions.GetValue(RawValue, ChannelPermission.UseVAD);
         /// <summary> If <c>true</c>, a user may use priority speaker in a voice channel. </summary>
         public bool PrioritySpeaker => Permissions.GetValue(RawValue, ChannelPermission.PrioritySpeaker);
+        /// <summary> If <c>true</c>, a user may stream video in a voice channel. </summary>
+        public bool Stream => Permissions.GetValue(RawValue, ChannelPermission.Stream);
```
Chris-Johnston, line 86: Could you add `Stream` to the relevant static read-only fields at the top of this file and GuildPermissions?

Chris-Johnston (changes requested):
Oops, I missed a spot

src/Discord.Net.Core/Entities/Permissions/ChannelPermissions.cs
RheaAyase: Hmm. Could that not be constructed from the enums instead of hardcoded?

FiniteReality (commented):
Even though the permission on Discord's side is called `Stream`, I'm half considering suggesting we call these permissions `GoLive` instead...

RheaAyase:
Yeah it's inconsistent... ~~I was going off of the actual server->roles->permission name, which is also `Stream` same as in the api docs.~~
__I was wrong heh, it's Go Live there.__
/shrug

RheaAyase:
Want me to change it?

src/Discord.Net.Core/Entities/Permissions/ChannelPermissions.cs
Chris-Johnston: It could be. I believe this was done just for conciseness.

RheaAyase: Well, unless I'm told otherwise, I'll keep it the way it is for consistency with other places... :)

Chris-Johnston: Could you please also update this line in GuildPermissions?

https://github.com/discord-net/Discord.Net/blob/ff0fea98a65d907fbce07856f1a9ef4aebb9108b/src/Discord.Net.Core/Entities/Permissions/GuildPermissions.cs#L15

Chris-Johnston (changes requested):
One more thing to update. Also, no need to keep force-pushing since it'll all get squash merged in the end.

https://github.com/discord-net/Discord.Net/blob/ff0fea98a65d907fbce07856f1a9ef4aebb9108b/src/Discord.Net.Core/Entities/Permissions/GuildPermissions.cs#L15

RheaAyase:
I'm a git ninja, I amend commits and rebase left and right! :D 

It's common workflow to keep commits clean and not everyone likes squashing, many repositories like rebase, etc...

Anyway I pushed the change for now hardcoded, I'll probably get back to it next week to change it to use the enums if the consensus is that it should be changed. But let's get this out of the way now to move onto the announcement channels :D

Chris-Johnston (approved):

foxbot (approved):

MazeXP:
OverwritePermissions is missing the "Stream" permission

PR #1359, opened 2019-08-22
[ci skip] Fixing GatewayEncoding comment

moiph:
Const, not a property

### DIFF
--- a/src/Discord.Net.WebSocket/DiscordSocketConfig.cs
+++ b/src/Discord.Net.WebSocket/DiscordSocketConfig.cs
@@ -26,7 +26,7 @@ namespace Discord.WebSocket
     public class DiscordSocketConfig : DiscordRestConfig
     {
         /// <summary>
-        ///     Gets or sets the encoding gateway should use.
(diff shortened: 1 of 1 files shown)


Niewiarowski:
Thanks for your contribution. Can't wait for this to be merged. I can't believe no one noticed this before. I'm fairly certain that this will provide a huge performance boost. This is just what this library needed. I cannot express how thankful I am to you. I hope to see more from you! Keep it up.

jacksonrakena:
Documentation changes should have a `[ci skip]` label (See [this docs PR](https://github.com/discord-net/Discord.Net/commit/15b2a36790ceaab453d001d939d039860592dd0d))

GreemDev:
I'd recommend changing that from `returns` to `represents` or something similar; as a const is not a method.

moiph:
> I'd recommend changing that from `returns` to `represents` or something similar; as a const is not a method.

I'd agree but I'm just following language already set by other `const` related comments

GreemDev:
> I'd agree but I'm just following language already set by other `const` related comments

change all the other ones then lol

PR #1360, opened 2019-08-22
added ModuleInfo.ModuleType

Foxite:
I wasn't able to find a way to get the Type of a module based on its ModuleInfo, so I made a way.


FiniteReality:
This is intentional. Modules do not always have types, and can be built from a custom ModuleBuilder.

k-boyle:
In Qmmands we just use the dark magic that is null

PR #1362, opened 2019-08-24
Made properties that should have been Cacheable, Cacheable

k-boyle:
No reason for these to be optional. I didn't find any other instances where optionals were used instead of cacheable, if there are any let me know and I'll add them to the PR.

I may or may not be smart enough to figure out how to resolve the conflicts, but everything should be fine.


FiniteReality:
While I'm not ruling it out, this is unlikely to be merged as it is a breaking change, and AFAIK fox still wants to push for Wumpus.Net for 3.0.

However, I do think this is a good "middle ground" for exposing properties which may not be cached.

PR #1368, opened 2019-09-02
Fix #1365, Update ISystemMessage interface to allow reactions

### LINKED ISSUE #1365: Reactions can now be added to System messages, Entities need to be updated
Not sure when this behavior changed, but reactions can now be added to system messages. (May also come with other changes?)

image

This will change the interface of RestSystemMessage and SocketSystemMessage. Maybe all that's required is to make them inherit from IUserMessage instead of IMessage?

Chris-Johnston:
Fix #1365 . Updates the interface of `ISystemMessage` `RestSystemMessage` and `SocketSystemMessage` so that they now support the various reaction methods that are already part of `IUserMessage` (and implementations of it)

The reaction implementation from `[Rest|Socket]UserMessage` has been copied into `[Rest|Socket]Message`.

`IUserMessage` is still distinct from `IMessage`, since system messages cannot be pinned or modified.


src/Discord.Net.Core/Entities/Messages/IReactionMessage.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Discord
+{
+    /// <summary>
+    ///     Represents a message where reactions can be added or removed.
+    /// </summary>
+    public interface IReactionMessage : IMessage
```
foxbot: Is there a reason these methods couldn't have just been added to IMessage?

Chris-Johnston: Nope, must have thought there was another type of message I was forgetting. Resolved in 118217387e20d1f4e9704db47c91eeafdc2ed39a

foxbot (approved):

PR #1373, opened 2019-09-10
Audit log improvements

NeKzor:
**New**
- ChannelCreateAuditLogData
  - SlowModeInterval¹
  - IsNsfw
  - Bitrate
- ChannelDeleteAuditLogData
  - SlowModeInterval
  - IsNsfw
  - Bitrate
- ChannelInfo
  - SlowModeInterval
  - IsNsfw
- GuildInfo
  - SystemChannelId²
  - EmbedChannelId
  - IsEmbeddable
- OverwriteCreateAuditLogData
  - ChannelId
- OverwriteDeleteAuditLogData
  - ChannelId
- OverwriteUpdateAuditLogData
  - ChannelId

¹ using "rate_limit_per_user" key which is not in the [official docs](https://discordapp.com/developers/docs/resources/audit-log#audit-log-change-object-audit-log-change-key) somehow
² using "system_channel_id" key which is also missing in the docs

**Bug Fixes**
- OverwriteUpdateAuditLogData
- RoleUpdateAuditLogData

**Breaking Changes**
- Removed UserLimit in ChannelInfo (there is no "user_limit" key)
- Removed AvatarHash in MemberInfo ("avatar_hash" only exists for webhooks)
- Changed name ContentFilterLevel of type int? to name ExplicitContentFilter of type ExplicitContentFilterLevel? (name change for consistency)

### DIFF
--- a/src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ChannelInfo.cs
+++ b/src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ChannelInfo.cs
@@ -34,6 +34,7 @@ namespace Discord.Rest
         /// <returns>
         ///     An <see cref="Int32"/> representing the time in seconds required before the user can send another
         ///     message; <c>0</c> if disabled.
+        ///     <c>null</c> if this is not mentioned in this entry.
         /// </returns>
         public int? SlowModeInterval { get; }
         /// <summary>
@@ -41,14 +42,15 @@ namespace Discord.Rest
         /// </summary>
         /// <returns>
         ///     <c>true</c> if this channel has the NSFW flag enabled; otherwise <c>false</c>.
+        ///     <c>null</c> if this is not mentioned in this entry.
         /// </returns>
         public bool? IsNsfw { get; }
         /// <summary>
         ///     Gets the bit-rate of this channel if applicable.
         /// </summary>
         /// <returns>
-        ///     An <see cref="System.Int32"/> representing the bit-rate set for the voice channel; <c>null</c> if not
-        ///     applicable.
+        ///     An <see cref="Int32"/> representing the bit-rate set for the voice channel;
+        ///     <c>null</c> if this is not mentioned in this entry.
         /// </returns>
         public int? Bitrate { get; }
     }
--- a/src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ChannelInfo.cs
+++ b/src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ChannelInfo.cs
@@ -5,12 +5,13 @@ namespace Discord.Rest
     /// </summary>
     public struct ChannelInfo
     {
-        internal ChannelInfo(string name, string topic, int? bitrate, int? limit)
+        internal ChannelInfo(string name, string topic, int? rateLimit, bool? nsfw, int? bitrate)
         {
             Name = name;
             Topic = topic;
+            SlowModeInterval = rateLimit;
+            IsNsfw = nsfw;
             Bitrate = bitrate;
-            UserLimit = limit;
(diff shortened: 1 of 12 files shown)


src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ChannelInfo.cs
```diff
@@ -28,20 +29,27 @@ internal ChannelInfo(string name, string topic, int? bitrate, int? limit)
...
+        ///     An <see cref="Int32"/> representing the time in seconds required before the user can send another
+        ///     message; <c>0</c> if disabled.
         /// </returns>
-        public int? Bitrate { get; }
+        public int? SlowModeInterval { get; }
+        /// <summary>
+        ///     Gets the value that indicates whether this channel is NSFW.
+        /// </summary>
+        /// <returns>
+        ///     <c>true</c> if this channel has the NSFW flag enabled; otherwise <c>false</c>.
```
Chris-Johnston, line 44: I think it may be helpful to include what a null value indicates, [similar to how it's done here](https://github.com/discord-net/Discord.Net/blob/cff00608e1e53b9168a6e55855128e22067055dd/src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/RoleEditInfo.cs#L31) and for Bitrate. I think the "`<c>null</c> if this is not mentioned in this entry.`" wording works better.

src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/GuildInfo.cs
```diff
@@ -82,6 +86,35 @@ public struct GuildInfo
...
+        /// <returns>
+        ///     A <see cref="ulong"/> representing the snowflake identifier of the embedded channel found within the
+        ///     widget settings of this guild; <c>null</c> if none is set.
+        /// </returns>
+        public ulong? EmbedChannelId { get; }
+        /// <summary>
+        ///     Gets a value that indicates whether this guild is embeddable (i.e. can use widget).
+        /// </summary>
+        /// <returns>
+        ///     <c>true</c> if this guild can be embedded via widgets; otherwise <c>false</c>.
```
Chris-Johnston, line 123: Similar to the other properties I think we should mention `null`.

src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/GuildInfo.cs
```diff
@@ -82,6 +86,35 @@ public struct GuildInfo
...
-        public int? ContentFilterLevel { get; }
+        /// <summary>
+        ///     Gets the level of content filtering applied to user's content in a Guild.
+        /// </summary>
+        /// <returns>
+        ///     The level of explicit content filtering.
+        /// </returns>
+        public ExplicitContentFilterLevel? ExplicitContentFilter { get; }
+        /// <summary>
+        ///     Gets the ID of the channel where randomized welcome messages are sent.
```
Chris-Johnston: System channels also do boost messages, I'd reword to just say "the channel where system messages are sent".

Chris-Johnston (changes requested):
Looks good. I have some minor suggestions for some of the xmldoc comments, didn't see any issue with the implementation.

> using "rate_limit_per_user" key which is not in the official docs somehow
> using "system_channel_id" key which is also missing in the docs

was able to confirm these by checking with the client debug tools

> Removed UserLimit in ChannelInfo (there is no "user_limit" key)

Can confirm, tried modifying this setting with a non-owner user and didn't see entries produced with this key in it. (Which is why I assigned "priority: bug")

Chris-Johnston (approved):

foxbot (approved):

PR #1375, opened 2019-09-13
Bug: Use double precision for X-Reset-After, set CultureInfo when parsing numeric types

### LINKED ISSUE #1374: Exception when sending embeds on multiple servers
As mentioned in the title, this happens when sending embeds on multiple servers at once.
The installed version is the latest nightly build (2.2.0-dev-20190908.10)

Exception:
```
System.ArgumentOutOfRangeException: Valid values are between -62135596800000 and 253402300799999, inclusive.
Parameter name: milliseconds
   at System.DateTimeOffset.FromUnixTimeMilliseconds(Int64 milliseconds)
   at Discord.Net.RateLimitInfo..ctor(Dictionary`2 headers)
   at Discord.API.DiscordRestApiClient.SendJsonAsync[TResponse](String method, String endpoint, Object payload, String bucketId, ClientBucketType clientBucket, RequestOptions options)
   at Discord.API.DiscordRestApiClient.CreateMessageAsync(UInt64 channelId, CreateMessageParams args, RequestOptions options)
   at Discord.Rest.ChannelHelper.SendMessageAsync(IMessageChannel channel, BaseDiscordClient client, String text, Boolean isTTS, Embed embed, RequestOptions options)```

### PR DESCRIPTION
**Pull request by Chris-Johnston**
Fixes #1374 . This PR addresses two issues that were found with parsing the X-Reset-After header.

Float did not contain enough precision to store the millisecond unix time value, which resulted in the second and millisecond values being slightly off.

```cs
> DateTimeOffset.FromUnixTimeMilliseconds((long)(1470173022.123f * 1000)).Millisecond
160 // wrong

> DateTimeOffset.FromUnixTimeMilliseconds((long)(1470173022.123 * 1000)).Millisecond
123 // correct
```

As a continuation of #1376 , occurrences of Parse or TryParse of numeric types were updated to explicitly state the NumberStyle and IFormatProvider, not just in RateLimitInfo. This is to prevent against this bug re-appearing in the future.

### DIFF
--- a/src/Discord.Net.Commands/Readers/UserTypeReader.cs
+++ b/src/Discord.Net.Commands/Readers/UserTypeReader.cs
@@ -47,7 +47,7 @@ namespace Discord.Commands
             if (index >= 0)
             {
                 string username = input.Substring(0, index);
-                if (ushort.TryParse(input.Substring(index + 1), NumberStyles.None, CultureInfo.InvariantCulture, out ushort discriminator))
+                if (ushort.TryParse(input.Substring(index + 1), out ushort discriminator))
                 {
                     var channelUser = await channelUsers.FirstOrDefault(x => x.DiscriminatorValue == discriminator &&
                         string.Equals(username, x.Username, StringComparison.OrdinalIgnoreCase)).ConfigureAwait(false);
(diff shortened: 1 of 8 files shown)


FiniteReality:
I'm thinking the command parser and type reader stuff should still use the default culture, as they are parsing information from the user. This allows setting things like `Thread.CurrentCulture` for localization. The internal API code should definitely use the invariant culture though.

Chris-Johnston:
Oops, yeah I agree. Was intending to only cover the internal stuff but modified a type reader by mistake. Resolved in 80cfa925b5d02a6f7cf768dca1a9b98588da4f93.

foxbot (approved):

PR #1382, opened 2019-09-30
Fix .NET Core 3.0 compatibility + Drop NS1.3

Hawxy:
Due to IAsyncEnumerable being a part of the standard library with .NET Core 3.0, some users have been reporting problems with System.Interactive.Async. This PR updates the dependency to v4 with support for NC3, but doing has a number of flow-on effects:

- Support for NS1.3 (.NET Core 1.x) has to be removed as the lx library no longer supports it.
- NETFX support also needs to be bumped to 4.6.1 (NS2.0 is now the API baseline)
- A NS2.1 target has to be added so the correct BCL is consumed when referenced by a NC3 application.


src/Discord.Net.Core/Discord.Net.Core.csproj
```diff
@@ -4,13 +4,13 @@
     <AssemblyName>Discord.Net.Core</AssemblyName>
     <RootNamespace>Discord</RootNamespace>
     <Description>The core components for the Discord.Net library.</Description>
-    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">net46;netstandard1.3;netstandard2.0</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netstandard1.3;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">net461;netstandard2.0;netstandard2.1</TargetFrameworks>
```
AntiTcb, line 7: Would it be possible to instead add a specific target to netcoreapp3.0 to handle fixing this issue?
Something along the lines of `<TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">net461;netstandard2.0;netstandard2.1;netcoreapp3.0</TargetFrameworks>`

src/Discord.Net.Core/Discord.Net.Core.csproj
```diff
@@ -4,13 +4,13 @@
...
+    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netstandard2.0;netstandard2.1</TargetFrameworks>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.2" />
     <PackageReference Include="System.Collections.Immutable" Version="1.3.1" />
-    <PackageReference Include="System.Interactive.Async" Version="3.2.0" />
+    <PackageReference Include="System.Interactive.Async" Version="4.0.0" />
```
AntiTcb, line 13: carrying on from the above comment
```
<ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp3.0' ">
    <PackageReference Include="System.Interactive.Async" Version="4.0.0" />
  </ItemGroup>
  <ItemGroup Condition=" '$(TargetFramework)' != 'netcoreapp3.0' ">
    <PackageReference Include="System.Interactive.Async" Version="3.2.0" />
  </ItemGroup>
```

samples/01_basic_ping_bot/01_basic_ping_bot.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
AntiTcb: I'd like if all the samples were updated to target netcoreapp3.0.

src/Discord.Net.Core/Discord.Net.Core.csproj
Hawxy: If I went down that path, it couldn't be a NC3 switch as lx 4 supports NS2 & NETFX461 (a newer version of lx might be resolved through other nuget packages the user installs and screw over everything). The old version would exist purely to support NS1.3. There'd also need to be a few ifdefs added to support the old PagedEnumerable implementation. I can do a setup like that if required but I'm not sure it's worth the effort just to support .NET Core 1.x which has been EoL for quite a long time now.

src/Discord.Net.Examples/Discord.Net.Examples.csproj
```diff
@@ -15,7 +15,7 @@
   <ItemGroup>
     <ProjectReference Include="..\Discord.Net.Core\Discord.Net.Core.csproj" />
     <ProjectReference Include="..\Discord.Net.WebSocket\Discord.Net.WebSocket.csproj" />
-    <PackageReference Include="JetBrains.Annotations" Version="2018.3.0" />
+    <PackageReference Include="JetBrains.Annotations" Version="2019.1.3" />
```
Chris-Johnston, line 18: Why do we have this dependency, anyways?

test/Discord.Net.Tests.Unit/EmbedBuilderTests.cs
```diff
@@ -250,7 +250,7 @@ public void Length()
                 .WithFooter("This is the footer", url)
                 .WithImageUrl(url)
                 .WithThumbnailUrl(url)
-                .WithTimestamp(DateTime.MinValue)
+                .WithTimestamp(DateTimeOffset.MinValue)
```
Chris-Johnston: Might want to cherry-pick this into a different PR?

src/Discord.Net.Examples/Discord.Net.Examples.csproj
Hawxy: The PublicAPI attribute is being used for a number of the examples, so I'm guessing its for the docs.

foxbot (approved):

PR #1383, opened 2019-10-01
Allow the ability to set presence on start

k-boyle:
All the other libs have it


AraHaan:
So @k-boyle mind checking if the code can be applied to the latest changes on the dev branch?

I would like to know if this is still wanted as well? This would simplify my ``ReadAsync`` event to just an ``Console.WriteLine`` saying the bot is ready then.

PR #1385, opened 2019-10-01
Added a Ready event to DiscordShardedClient

k-boyle:
Currently users are having to manually hack in their own Ready event, which shouldn't be necessary


FiniteReality:
I believe we intentionally did not include a Ready event when this was made. Closing for now, but if @foxbot decides it's useful, it can easily be reopened.

PR #1387, opened 2019-10-01
Fix #1381 Guild PreferredLocale support

### LINKED ISSUE #1381: support preferred locale
`guild.preferred_locale`

```
"Value must be one of (u'el', u'vi', u'it', u'bg', u'cs', u'sv-SE', u'es-ES', u'en-GB', u'ru', u'nl', u'no', u'tr', u'lt', u'th', u'ro', u'pl', u'fr', 'en-US', u'hr', u'de', u'da', u'pt-BR', u'fi', u'hu', u'ja', u'zh-TW', u'ko', u'zh-CN', u'uk')."
```

this is not yet in discord's UI and will probably change

Chris-Johnston:
Fix #1381 
Adds support for getting and modifying a guild's `preferred_locale`. This is a language tag in IETF BCP 47 format, which works with the built-in `CultureInfo` class.

While Discord only supports a number of cultures (as listed in the linked issue), I think that this restriction should be handled by the API and not by the wrapper.

Happy Hacktoberfest 💀 🎺

### DIFF
--- a/src/Discord.Net.WebSocket/Entities/Guilds/SocketGuild.cs
+++ b/src/Discord.Net.WebSocket/Entities/Guilds/SocketGuild.cs
@@ -5,6 +5,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
+using System.Globalization;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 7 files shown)


src/Discord.Net.WebSocket/Entities/Guilds/SocketGuild.cs
```diff
@@ -105,6 +106,16 @@ public class SocketGuild : SocketEntity<ulong>, IGuild, IDisposable
...
+        /// <inheritdoc />
+        public string PreferredLocale { get; private set; }
+
+        /// <summary>
+        ///     Gets the preferred culture of this guild.
+        /// </summary>
+        /// <returns>
+        ///     The preferred culture information of this guild.
+        /// </returns>
+        public CultureInfo PreferredCulture { get; private set; }
```
foxbot, line 113: Is there a reason this was left off IGuild?

Chris-Johnston: Didn't think to add it to IGuild at the time, but it makes sense now that I think about it again. Will update later today.

PR #1388, opened 2019-10-01
Fix test that fails on half of the planet 

Hawxy:
If you happen to live somewhere with a positive offset from UTC, the implicit conversion from `DateTime.MinValue` to `DateTimeOffset` tends to results in a `System.ArgumentOutOfRangeException` as it takes the local timezone into account.


Chris-Johnston (approved):
LGTM. Just did search for `DateTime`, couldn't find any other usage of it that should break things.

foxbot (approved):

PR #1391, opened 2019-10-03
Use '??='

Joe4evr:
Very minor, but it does make the code nicer to read.

~~This change will require the .NET Core 3.0 SDK to compile, but since #1382 already changes the CI script to use that, this would be predicated on that getting merged in first.~~


Joe4evr:
@foxbot PTAL

PR #1393, opened 2019-10-05
Utilize ValueTuples

Joe4evr:
Found this as a potential improvement.

Should #1382 get merged, the polyfill can be dropped.


Joe4evr:
Oops, well *there's* a problem. 😳

TehPers:
Rather than adding your own ValueTuple implementation, would it be better to use [the Microsoft NuGet package for it](https://www.nuget.org/packages/System.ValueTuple/)?

Joe4evr:
I guess that's another option, but it seems like we might drop NetStandard 1.3 support soon anyway.

Joe4evr:
@foxbot PTAL

foxbot (approved):

PR #1399, opened 2019-10-16
Outer groups

LtLi0n:
Adding an optional property **Parent** for **GroupAttribute**.

It allows group modules to be within separate files, outside of parent class, while still having a dependency on that parent module.

Tested on my end, works fine, but I may have missed something. 

Please take a closer look.


LtLi0n:
Outer subtypes still leave **Parent** module property uninitialized. 
Trying to solve how to bind them properly, so there is a connection.

I should be able to initialize them on autoLoad, but can't figure out binding through manual adding.

I suspect that currently manual adding doesn't work for nested group modules too?

FiniteReality:
What the hell is the use case for this?

Does this not work?

```cs
// File1
public partial class X
{
    [Module("derp")]
    public class Y
    {
    }
}

[Module("thing")]
public partial class X
{
    [Module("derpderp")]
    public class Z
    {
    }
}
```

LtLi0n:
Seriously?

What if you have a large group module, say for a game within the bot that will have inner 20 submodules or more, so each having their own distinct class and file would be a lot more maintainable. Not to mention little advantages, such as `ctrl` + `left-click` to open parent class when it's not partial.

You tell me, which one looks clearer

this:

```csharp
[Group("game1")]
public partial class Game1Module
{
	[Group("owner")]
	public class Game1OwnerModule
	{
	}
}

public partial class Game1Module
{
	[Group("info")]
	public class Game1InfoModule
	{
	}
}

public partial class Game1Module
{
	[Group("user")]
	public class Game1UserModule
	{
	}
}
```

or this:
```csharp
[Group("game1")]
public class Game1Module
{

}

[Group("owner", Parent = typeof(Game1Module))]
public class Game1OwnerModule
{

}

[Group("info", Parent = typeof(Game1Module))]
public class Game1InfoModule
{

}

[Group("user", Parent = typeof(Game1Module))]
public class Game1UserModule
{

}
```

If submodules are expected to contain a lot of code, all I can see are a lot of useless tab indentations judging from the first example.

k-boyle:
What happens if you do
```cs
[GroupAttribute("", Parent = typeof(MyOtherModule))]
public class MyModule : ModuleBase 
{

}

[GroupAttribute("", Parent = typeof(MyModule))]
public class MyOtherModule : ModuleBase
{

}
```

Or set the parent to a nested module?

src/Discord.Net.Commands/Attributes/GroupAttribute.cs
```diff
@@ -13,6 +13,11 @@ public class GroupAttribute : Attribute
         /// </summary>
         public string Prefix { get; }
 
+        /// <summary>
+        ///     Uses this module as parent instead. Allows creation of group modules outside of parent class.
```
k-boyle: as the parent

src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
```diff
@@ -10,45 +10,59 @@ namespace Discord.Commands
...
+                bool typeInfoIsValid = await typeInfo.IsValidModuleType(autoLoadable: true, logger: service._cmdLogger);
+                if(typeInfoIsValid)
                 {
-                    await service._cmdLogger.WarningAsync($"Class {typeInfo.FullName} is not public and cannot be loaded. To suppress this message, mark the class with {nameof(DontAutoLoadAttribute)}.").ConfigureAwait(false);
+                    standardModules.Add(typeInfo);
                 }
             }
 
-            return result;
+            return Tuple.Create<IReadOnlyList<TypeInfo>, IReadOnlyList<TypeInfo>>(standardModules, queuedModules);
```
k-boyle: value tuple who?

src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
```diff
@@ -118,6 +134,20 @@ private static void BuildModule(ModuleBuilder builder, TypeInfo typeInfo, Comman
                     case GroupAttribute group:
                         builder.Name = builder.Name ?? group.Prefix;
                         builder.Group = group.Prefix;
+                        if(group.ParentModule != null)
+                        {
+                            if(moduleDefs.ContainsKey(group.ParentModule))
+                            {
+                                foreach (string parentAlias in moduleDefs[group.ParentModule].Aliases)
```
k-boyle: use `TryGetValue` instead of a double lookup

src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
```diff
@@ -10,45 +10,59 @@ namespace Discord.Commands
 {
     internal static class ModuleClassBuilder
     {
-        private static readonly TypeInfo ModuleTypeInfo = typeof(IModuleBase).GetTypeInfo();
+        public static readonly TypeInfo ModuleTypeInfo = typeof(IModuleBase).GetTypeInfo();
 
-        public static async Task<IReadOnlyList<TypeInfo>> SearchAsync(Assembly assembly, CommandService service)
+        ///<summary>Value1 - ready types, Value2 - types dependant on Value1.</summary>
+        public static async Task<Tuple<IReadOnlyList<TypeInfo>, IReadOnlyList<TypeInfo>>> SearchAsync(Assembly assembly, CommandService service)
```
k-boyle: value tuple

k-boyle:
> Seriously?
> 
> What if you have a large group module, say for a game within the bot that will have inner 20 submodules or more, so each having their own distinct class and file would be a lot more maintainable. Not to mention little advantages, such as `ctrl` + `left-click` to open parent class when it's not partial.
> 
> You tell me, which one looks clearer
> 
> this:
> 
> ...
> 
> If submodules are expected to contain a lot of code, all I can see are a lot of useless tab indentations judging from the first example.

Adding an overload to `ModuleBuilder#AddModule` to take a type and then override `OnModuleBuilding` and adding the modules that way would prob be cleaner ngl

LtLi0n:
Thank you for your review. 
I appreciate your effort and you stated many valid arguments with some test cases to keep in mind.

> What happens if you do
> 
> ```cs
> [GroupAttribute("", Parent = typeof(MyOtherModule))]
> public class MyModule : ModuleBase 
> {
> 
> }
> 
> [GroupAttribute("", Parent = typeof(MyModule))]
> public class MyOtherModule : ModuleBase
> {
> 
> }
> ```
> 
> Or set the parent to a nested module?

The above example throws an exception, because parent was not found within a dictionary of regular modules. 
Nested submodules don't work for that exact same reason, good observation.
I really need to join modules before they get built, then cycle through them until it encounters a circular dependency or it's finished. Would solve parent binding too, so **Module.Parent** and **Module.SubModules** properties get initialized.

Will try to resolve:
- [ ] circular dependencies
- [x] nested outer subtype parent support 
- [x] proper parent links

> > Seriously?
> > What if you have a large group module, say for a game within the bot that will have inner 20 submodules or more, so each having their own distinct class and file would be a lot more maintainable. Not to mention little advantages, such as `ctrl` + `left-click` to open parent class when it's not partial.
> > You tell me, which one looks clearer
> > this:
> > ```cs
> > [Group("game1")]
> > public partial class Game1Module
> > {
> > 	[Group("owner")]
> > 	public class Game1OwnerModule
> > 	{
> > 	}
> > }
> > 
> > public partial class Game1Module
> > {
> > 	[Group("info")]
> > 	public class Game1InfoModule
> > 	{
> > 	}
> > }
> > 
> > public partial class Game1Module
> > {
> > 	[Group("user")]
> > 	public class Game1UserModule
> > 	{
> > 	}
> > }
> > ```
> > 
> > 
> > or this:
> > ```cs
> > [Group("game1")]
> > public class Game1Module
> > {
> > 
> > }
> > 
> > [Group("owner", Parent = typeof(Game1Module))]
> > public class Game1OwnerModule
> > {
> > 
> > }
> > 
> > [Group("info", Parent = typeof(Game1Module))]
> > public class Game1InfoModule
> > {
> > 
> > }
> > 
> > [Group("user", Parent = typeof(Game1Module))]
> > public class Game1UserModule
> > {
> > 
> > }
> > ```
> > 
> > 
> > If submodules are expected to contain a lot of code, all I can see are a lot of useless tab indentations judging from the first example.
> 
> Adding an overload to `ModuleBuilder#AddModule` to take a type and then override `OnModuleBuilding` and adding the modules that way would prob be cleaner ngl

How come? The user would have to manually add each outer subtype then.
That would create extra work for a user when they decide to add, rename or remove outer subtype classes.

k-boyle:
Huh? It's as much work as specify a group on each module and its parent.
Plus afaik the implementation within the library should already but there,
just need to expose a method

src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
LtLi0n: Unable to use value tuples, for now. More in-depth: #1393

src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
LtLi0n: Unable to use value tuples, for now. More in-depth: #1393

src/Discord.Net.Commands/Attributes/GroupAttribute.cs
LtLi0n: Derp, fixed with https://github.com/discord-net/Discord.Net/pull/1399/commits/b693af5a0dd5ef73834b71e3743f462dfee4453b

src/Discord.Net.Commands/Builders/ModuleClassBuilder.cs
LtLi0n: fixed with https://github.com/discord-net/Discord.Net/pull/1399/commits/eae4839e1ec6d8d82fc61a64696402a0a7f7b1e5

LtLi0n:
```csharp
[Group("panel")]
public class PanelModule : ModuleBase<SocketCommandContext>
{
    [Command("test")]
    public async Task Test() => await ReplyAsync("panel here");
}

[Group("settings", ParentModule = typeof(PanelModule))]
public class PanelSettingsModule : ModuleBase<SocketCommandContext>
{
    [Command("test")]
    public async Task Test() => await ReplyAsync("settings here");
}

[Group("guild", ParentModule = typeof(PanelSettingsModule))]
public class PanelSettingsGuildModule : ModuleBase<SocketCommandContext>
{
    [Command("test")]
    public async Task Test() => await ReplyAsync("settings:guild here");
}

[Group("info", ParentModule = typeof(PanelModule))]
public class PanelInfoModule : ModuleBase<SocketCommandContext>
{
    [Command("test")]
    public async Task Test() => await ReplyAsync("info here");

    [Group("nested")]
    public class PanelInfoNestedModule : ModuleBase<SocketCommandContext>
    {
        [Command("test")]
        public async Task Test() => await ReplyAsync("nested info here");
    }
}
```

### output:

image

I don't fully know what to do with circular dependencies yet.
Currently, it doesn't crash, it just skips the modules that are conflicting, so the user would see that the module isn't even loaded.

LtLi0n (commented):
Fixes a priority bug for outer subtypes, where inner subtypes would be ignored if outer subtypes were also present.

```csharp
public class TestModule : ModuleBase<SocketCommandContext>
{
    [Group("test")]
    public class TestWithin : ModuleBase<SocketCommandContext>
    {
        [Group("inner")]
        public class InnerModule : ModuleBase<SocketCommandContext>
        {
            [Command]
            public async Task Test() => await ReplyAsync("inner hello");
        }
    }
}

[Group("outer", ParentModule = typeof(TestModule.TestWithin))]
public class OuterModule : ModuleBase<SocketCommandContext>
{
    [Command]
    public async Task Test() => await ReplyAsync("outer hello");
}
```

### output:
image

quinchs:
@Cenngo Could you take a look at this?

Cenngo:
> @Cenngo Could you take a look at this?

Honestly i dont see why this is should preferred over partial classes. with partial classes you can have shared properties, methods etc. If the main objective is to be able to seperate modules into multiple files, partial classes should cover that.

PR #1405, opened 2019-10-25
Add .NET Standard 2.1 support Color

### LINKED ISSUE #1404: Cannot cast System.Drawing.Color to Discord.Color as part of .NET Core 3.0/.NET Standard 2.1
Hard coded conditional flags make this impossible to use in a .NET Core 3.0/.NET Standard 2.1 environment, it seems.

https://github.com/discord-net/Discord.Net/blob/dev/src/Discord.Net.Core/Entities/Roles/Color.cs#L193-L198

patrickklaeren:
Adds NETSTANDARD2_1 as a flag to the Discord Color operator. Fixes #1404

### DIFF
--- a/src/Discord.Net.Core/Entities/Roles/Color.cs
+++ b/src/Discord.Net.Core/Entities/Roles/Color.cs
@@ -1,8 +1,6 @@
 using System;
 using System.Diagnostics;
-#if NETSTANDARD2_1 || NETSTANDARD2_0 || NET45
 using StandardColor = System.Drawing.Color;
-#endif
 
 namespace Discord
 {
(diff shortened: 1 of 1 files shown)


src/Discord.Net.Core/Entities/Roles/Color.cs
```diff
@@ -1,6 +1,6 @@
 using System;
 using System.Diagnostics;
-#if NETSTANDARD2_0 || NET45
+#if NETSTANDARD2_1 || NETSTANDARD2_0 || NET45
```
AntiTcb: Given we just merged #1382 to drop NS1.3, this conditional can be removed entirely since all supported runtimes have System.Drawing.Color now.

src/Discord.Net.Core/Entities/Roles/Color.cs
```diff
@@ -190,7 +190,7 @@ public override bool Equals(object obj)
 
         public override int GetHashCode() => RawValue.GetHashCode();
 
-#if NETSTANDARD2_0 || NET45
+#if NETSTANDARD2_1 || NETSTANDARD2_0 || NET45
```
AntiTcb: See above.

AntiTcb (changes requested):

src/Discord.Net.Core/Entities/Roles/Color.cs
patrickklaeren: All removed @AntiTcb

FiniteReality (approved):

PR #1412, opened 2019-11-05
RestGuild.GetUsersAsync limited to 100/request instead of supported 1000.

LtLi0n:
Fixed `GetUsersAsync` method to use `MaxUsersPerBatch` const as a limit instead of `MaxMessagesPerBatch`.

Requests are now returning up to 1000 guild user entities instead of the previous 100.


AntiTcb:
Rebase your fork; the only commits should be those relevant to your PR.

LtLi0n:
Sorry about that 😅

foxbot (approved):

PR #1416, opened 2019-11-13
Fix: False-positive detection of CustomStatusGame based on Id property

Chris-Johnston:
This change fixes a bug that was introduced in PR #1406. (regular) Games were falsely detected to be CustomStatusGames, based on the Id property being included in the activity model payload. [Original conversation here.](https://discordapp.com/channels/81384788765712384/381889909113225237/643873873951195166)

This fixes the false detection of Games as CustomStatusGame. An activity will only be considered a CustomStatusGame if the Id has a value of "custom".

This change has been manually tested by listening to the GuildMemberUpdated event, and opening/closing games with a custom status set.


foxbot (approved):
This might not be the final behavior here (I believe they plan on changing custom status in the future) but that should be addressable through a patch

PR #1438, opened 2020-01-07
Add guild model

trinitrotoluene:


SubZero0:
Not sure what this is about since these are already implemented? I'll be closing since there's no explanation (and it's almost a year old so I don't expect them to be around anymore).

PR #1458, opened 2020-03-05
Implement missing audit log types

NeKzor:
**New Types**
- BotAddAuditLogData
- MemberDisconnectAuditLogData
- MemberMoveAuditLogData
- MessageBulkDeleteAuditLogData
- MessagePinAuditLogData
- MessageUnpinAuditLogData

**Internal Changes**
- Renamed MessageDeleteCount to Count
- Renamed MessageDeleteChannelId to ChannelId

**Breaking Change**
- MessageDeleteAuditLogData property `AuthorId` of type `ulong` is now `Target` of type `IUser`

### DIFF
--- a/src/Discord.Net.Rest/API/Common/AuditLogOptions.cs
+++ b/src/Discord.Net.Rest/API/Common/AuditLogOptions.cs
@@ -4,11 +4,12 @@ namespace Discord.API
 {
     internal class AuditLogOptions
     {
-        //Message delete
         [JsonProperty("count")]
-        public int? MessageDeleteCount { get; set; }
+        public int? Count { get; set; }
         [JsonProperty("channel_id")]
-        public ulong? MessageDeleteChannelId { get; set; }
+        public ulong? ChannelId { get; set; }
+        [JsonProperty("message_id")]
+        public ulong? MessageId { get; set; }
 
         //Prune
         [JsonProperty("delete_member_days")]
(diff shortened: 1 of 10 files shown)


src/Discord.Net.Rest/API/Common/AuditLogOptions.cs
```diff
@@ -4,11 +4,12 @@ namespace Discord.API
 {
     internal class AuditLogOptions
     {
-        //Message delete
         [JsonProperty("count")]
-        public int? MessageDeleteCount { get; set; }
+        public int? Count { get; set; }
         [JsonProperty("channel_id")]
-        public ulong? MessageDeleteChannelId { get; set; }
```
Joe4evr, line 11: Why are these properties renamed and that comment removed?

FiniteReality: Looking at the code, probably because they're re-used for multiple audit log events. This isn't actually a breaking change as it's an internal type.

FiniteReality:
FWIW, your "breaking" changes aren't breaking since they're internal types. They'd only be breaking to people with InternalsVisibleTo access, which should in theory be very little.

FiniteReality (commented):
Looking at our other audit log data types, we provide IUser APIs instead of exposing the snowflake for properties such as the target. Maybe we should do that here?

src/Discord.Net.Rest/API/Common/AuditLogOptions.cs
FiniteReality: For context, when I initially made this type, only message deletes had `channel_id` and `count` in the `options` object in each audit log entry.

Issue #1471, opened 2020-03-31
GuildAvailable event not fired & IsConnected false; after joining a new guild.


TheKingEagle:
DiscordShardedClient is not firing the GuildAvailable event after a user adds the bot to their guild, as a result, IsConnected property is not properly updated. The client has to be restarted in order for the new guild to be considered connected and available.

renanrcp:
I think you wanna the JoinedGuild event

TheKingEagle:
Right, I am using that event but I also need the client to check SocketGuild.IsConnected for a management screen. That variable is false until I fully restart the client.

renanrcp:
I'm taking a look in the lib code, https://github.com/discord-net/Discord.Net/blob/adf823ca9a4fe8d6d812d952630023c9e7bfdcb9/src/Discord.Net.WebSocket/DiscordSocketClient.cs#L625 the lib really doesn't setting this property true when the CurrentUser join.

TheKingEagle:
Would this be considered an API, rather than library issue? `if (data.Unavailable == false)` is evaluating `false` in this case, which would tell me that the data is being sent as unavailable. I'm not entirely sure if that would mean anything though... [sorry]

rmsoftware-development:
Fixed with #1473

PR #1473, opened 2020-04-05
Call GuildAvailableAsync for dispatch(GUILD_CREATE) case

TheKingEagle:
This addition will raise the `GuildAvailable` event when the client joins a new guild. Solving issue #1471

### DIFF
--- a/src/Discord.Net.WebSocket/DiscordSocketClient.cs
+++ b/src/Discord.Net.WebSocket/DiscordSocketClient.cs
@@ -627,11 +627,8 @@ namespace Discord.WebSocket
                                         var guild = AddGuild(data, State);
                                         if (guild != null)
                                         {
-                                            if (_unavailableGuildCount != 0)
-                                                _unavailableGuildCount--;
-                                                
-                                            await GuildAvailableAsync(guild).ConfigureAwait(false);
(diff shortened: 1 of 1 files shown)


src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -627,6 +627,10 @@ private async Task ProcessMessageAsync(GatewayOpCode opCode, int? seq, string ty
                                         var guild = AddGuild(data, State);
                                         if (guild != null)
                                         {
+                                            if (_unavailableGuildCount != 0)
```
foxbot: unavailableGuildCount doesn't need to be affected here, since newly joined guilds aren't considered unavailable.

src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -627,6 +627,10 @@ private async Task ProcessMessageAsync(GatewayOpCode opCode, int? seq, string ty
...
+                                                _unavailableGuildCount--;
+                                                
+                                            await GuildAvailableAsync(guild).ConfigureAwait(false);
```
foxbot: nit: wait until after JoinedGuild to raise GuildAvailable; event subscribers should first be informed that we've joined a guild before informing them that the guild is now available.

foxbot (changes requested):

foxbot (approved):

PR #1478, opened 2020-04-11
Add cache purging methods

### LINKED ISSUE #1474: Clearing Guild and User cache
I have recently added a command that allows people to see what users from their guild I store in my database. For that command I use the method [`GetUsersAsync()`](https://discord.foxbot.me/docs/api/Discord.IGuild.html#Discord_IGuild_GetUsersAsync_Discord_CacheMode_Discord_RequestOptions_) from `IGuild`.

This gets all users and the guild info and stores it forever in the cache which (in my case) results in storing 2 massive lists in memory. This is visible in DotMemory:
image

image

Some background info; the bot is on ~5k guild and some of the largest servers have around 120k members. I only use this request to get the IDs from the users in the server; I don't care about any other data.

As far as I can see (and asked on Discord) there is currently no easy way to clear the cache or set a lifetime on this cache. I do see a `CacheMode`, but that is only used to set if the requests requests new objects or only uses the cache.
The only way to clear the cache would be to restart the application, or else it will just keep storing more servers even if they only use the command once.

My question is: Is it possible to add a method to clear the user/guild cache? Or maybe set a lifetime on the cache?

Joe4evr:
Fixes #1474.

### DIFF
--- a/src/Discord.Net.Core/Extensions/CollectionExtensions.cs
+++ b/src/Discord.Net.Core/Extensions/CollectionExtensions.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -15,7 +15,7 @@ namespace Discord
         //public static IReadOnlyCollection<TValue> ToReadOnlyCollection<TKey, TValue>(this IReadOnlyDictionary<TKey, TValue> source)
         //    => new CollectionWrapper<TValue>(source.Select(x => x.Value), () => source.Count);
         public static IReadOnlyCollection<TValue> ToReadOnlyCollection<TKey, TValue>(this IDictionary<TKey, TValue> source)
-            => new CollectionWrapper<TValue>(source.Select(x => x.Value), () => source.Count);
+            => new CollectionWrapper<TValue>(source.Values, () => source.Count);
         public static IReadOnlyCollection<TValue> ToReadOnlyCollection<TValue, TSource>(this IEnumerable<TValue> query, IReadOnlyCollection<TSource> source)
             => new CollectionWrapper<TValue>(query, () => source.Count);
         public static IReadOnlyCollection<TValue> ToReadOnlyCollection<TValue>(this IEnumerable<TValue> query, Func<int> countFunc)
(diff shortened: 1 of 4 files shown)


th0mk:
Would anyone have time to review this? I currently need to restart my bot every 3/4 days or the memory usage becomes too high because it caches users indefinitely

SubZero0:
I believe the lib isn't fitted to be used without a channel cache, a lot of gateway events depend on it and would make some of them unusable. This change would require reworking this behavior.

EDIT:
For the users, I think it's fine and our aim should be making the lib work without relying on a user cache, since it seems Discord plans on not sending those anymore with the recent changes.

k-boyle:
This is gonna break more shit than it fixes (which is nothing) and is just gonna have users yeet their cache and run into issues. Why not just actually clear up cache through channel destroyed, user left etc

SubZero0:
> This is gonna break more shit than it fixes (which is nothing) and is just gonna have users yeet their cache and run into issues. Why not just actually clear up cache through channel destroyed, user left etc

I'm more of the kind that the lib should let you decide if you want to maintain a cache or not (users, channels, guilds), but it would require reworking a lot of places to make it usable without it [the cache].

I belive 3.0 was supposed to have the barebones and a way to cache them, separated, but I have no idea on what foot it currently is.

Joe4evr:
> Why not just actually clear up cache through channel destroyed, user left etc

Because that ain't gonna do shit when you're talking about multiple guilds with 120k+ members.

th0mk:
> I'm more of the kind that the lib should let you decide if you want to maintain a cache or not (users, channels, guilds), but it would require reworking a lot of places to make it usable without it [the cache].

That would indeed be nice, maybe a `CacheMode` that would allow you to not store users in the cache, or some kind of way to set a certain lifetime. I think setting expiration time on caches (or just always expiring after a little bit) would always be a good thing to do.

That being said having a method to just clear the entire cache regardless of how old it is is maybe a less-elegant solution, but I would still prefer it over restarting my bot.

Still34 (commented):
Docs strings LGTM - need additional reviews from other contributors for the concept/implementation though.

foxbot (approved):

PR #1483, opened 2020-04-18
Revise minor grammatical issues in 3.0 docs outline

Still34:
# Summary

This PR fixes minor grammatical issues from the 3.0 docs outline. 

## Additional Comments

Since the project is still in its early infancy, I don't feel the need to revise the xml docs strings just yet. The initial overview of the structure seems fine to me otherwise.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,7 +1,7 @@
 # Discord.Net #
 [![Discord](https://discordapp.com/api/guilds/81384788765712384/widget.png)](https://discord.gg/discord-api)
 
-An unofficial .NET library for the Discord API. (https://discordapp.com/)
+An unofficial .NET library for the [Discord API](https://discordapp.com/).
 
 ## Installation ##
 
@@ -14,9 +14,10 @@ Unstable "nightly" builds are available through our MyGet feed:
 
 ## Building ##
 
-At minimum, .NET Core 2.1 is required to successfully build Discord.Net, but
-.NET core 3.1 is recommended. You should install the appropriate tools and
+At the minimum, .NET Core 2.1 is required to successfully build Discord.Net,
+but .NET Core 3.1 is recommended. You should install the appropriate tools and
 components for your editor to open the Discord.Net solution file:
+
 - The .NET Core workload for Visual Studio
 - The C# build tools for Visual Studio Code
 - The .NET Core SDK for command line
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -28,15 +29,15 @@ As a rule, Discord.Net attempts to follow
...
-to add code to support these. To do this, we will increment the minor version
-when adding new API surface to the library. To counter the inevitable breaking
+to add code to support these changes. To do this, we increment the minor version
+when adding new API surfaces to the library. To counter any inevitable breaking
 changes, we have decided that our public interfaces should be for
 *consumption only* - bot developers should not need to implement these
 interfaces for normal code.
 
-Additionally, as the Discord API changes, we may need to add, remove or change
+Additionally, as the Discord API changes, we may need to add, remove, or change
```
FiniteReality, line 38: IIRC the comma here is unnecessary since `or` is a conjunction - I've always been taught to either use a comma or a conjunction, never both.

docs/README.md
```diff
@@ -14,9 +14,10 @@ Unstable "nightly" builds are available through our MyGet feed:
 
 ## Building ##
 
-At minimum, .NET Core 2.1 is required to successfully build Discord.Net, but
-.NET core 3.1 is recommended. You should install the appropriate tools and
+At the minimum, .NET Core 2.1 is required to successfully build Discord.Net,
```
FiniteReality: I believe "at minimum" sounds better than "at the minimum" here. I'd usually associate "at the minimum" with the end of a sentence or clause, putting "at minimum" at the beginning.

FiniteReality (commented):
Thanks for cleaning up some of my errors! :sweat_smile:

Though I feel some of the changes may be unnecessary, which I have pointed out.

docs/README.md
Still34: Formally and grammatically speaking, they're both technically correct, though commas for a long list before the `or` is more common in American English. 

It's not an essential change, but figured I'd plop that in there.

docs/README.md
Still34: I agree with `at minimum` sounding more natural than `at a/the minimum`, so that's fair.

docs/README.md
Joe4evr: I'mma just leave this here for reference: https://en.wikipedia.org/wiki/Serial_comma

PR #1486, opened 2020-04-19
Fix add type reader for entity implementations

### LINKED ISSUE #1485: Impossible to overwrite default type reader for type that implements IMessage, IChannel, IRole or IUser
### Steps to reproduce
1. Create and register a type reader for `IMessageChannel` that simply returns the current channel regardless of input.
2. Add a command that uses a parameter of type `IMessageChannel`.
3. Start the client and attempt to use the command with gibberish input.
4. Notice the default type reader for `IChannel` is used and the command does not have the current channel as the parameter.

### Expected Result
It is possible to overwrite the default type reader for types that implement the entity types `IMessage`, `IChannel`, `IRole` and `IUser`.

### Code Snippet
```csharp
// Custom Type Reader
public class MessageChannel : TypeReader
{
    public override Task<TypeReaderResult> ReadAsync(ICommandContext context, string input, IServiceProvider services)
    {
        return Task.FromResult(TypeReaderResult.FromSuccess(context.Channel));
    }
}

// CommandService configuration
commandService.AddTypeReader<IMessageChannel>(new MessageChannel(), replaceDefault: true);

// Command
[Command("channelid")]
public Task ChannelIdAsync(IMessageChannel channel)
{
    return ReplyAsync(channel.Id.ToString());
}
```

### Additional Information
- Tested with version `2.3.0-dev-20200417.1`.
- It is possible to overwrite default type readers for entity types (`IMessage`, `IChannel`, `IRole` or `IUser`) directly, but not types that implement them (`IMessageChannel`, `IGuildUser`, ...). 
- Issue was introduced with #1083.

adamgauthier:
When adding a type reader that replaces a default type reader,
CommandService checks that it replaces a type in its default type
readers or entity type readers (IMessage, IChannel, IRole, IUser). For
entity types, it checks that the target type is the same as the entity
type or it implements its type reader. Adding a type reader for a
default type reader doesn't make much sense and the original intent was
likely to check that the target type is or implements the entity type.

Fixes #1485.

### DIFF
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -408,7 +408,7 @@ namespace Discord.Commands
             var typeInfo = type.GetTypeInfo();
             if (typeInfo.IsEnum)
                 return true;
-            return _entityTypeReaders.Any(x => type == x.EntityType || typeInfo.ImplementedInterfaces.Contains(x.TypeReaderType));
+            return _entityTypeReaders.Any(x => type == x.EntityType || typeInfo.ImplementedInterfaces.Contains(x.EntityType));
         }
         internal void AddNullableTypeReader(Type valueType, TypeReader valueTypeReader)
         {
@@ -511,7 +511,7 @@ namespace Discord.Commands
                 await _commandExecutedEvent.InvokeAsync(Optional.Create<CommandInfo>(), context, searchResult).ConfigureAwait(false);
                 return searchResult;
             }
-                
+
 
             var commands = searchResult.Commands;
             var preconditionResults = new Dictionary<CommandMatch, PreconditionResult>();


SubZero0:
While this indeed fixes replacing the default type reader for the exact type you want it to (unlike how it currently is that doesn't change anything), there's two things to notice.

1) This will add it in `_defaultTypeReaders`, not `_typeReaders`, so `CommandService.TypeReaders` won't show that custom type reader (this isn't the case with the current implementation, even if it doesn't actually replace it because of how they are prioritized, requiring to manually override with OverrideTypeReaderAttribute). A "band aid fix" would be adding it to the other list, even if it will never be accessed from there internally.

2) It won't work as a "default type reader", so adding it as ` _commands.AddTypeReader<IChannel>(new CustomChannelTypeReader(), replaceDefault: true);`, will only replace when the command is `IChannel`, not `IMessageChannel` or any other. This issue is for both how it currently is and after this change.

Tl;dr: it does replace the default type reader for the specific type, but doesn't list the custom type reader nor work as an entity type reader (this being another issue currently too).

EDIT: Seeing this again, it seems indeed a needed fix, but not to "Fix add type reader for entity implementations", since entity type readers added by the user continue not working (understand entity as a reader that works for themselves and classes that implement them).

adamgauthier:
@SubZero0 Hey, thank you for the quick response. I agree with your points and there are definitely changes that are needed to improve the type reader system, some of which might require breaking changes.

This PR doesn't address those and simply aims to bring the addition of type readers for types that implement IMessage, IChannel, IRole, or IUser *in line with* the addition of type readers for these types directly, which was the intended behavior with the current implementation.

As for the commit subject, sometimes it's hard to summarize the changes concisely, I now see what "entity" means more clearly, I was simply referring to types that implement types that are added as entity type readers by default.

PR #1487, opened 2020-04-20
Change priority to user type readers and implement user entity readers

SubZero0:
## Summary

This change will rollback the change from #941 that gave priority to default type readers instead of the ones added by the user while also fixing the issue that PR was merged for by creating a specific list for override type readers (so it isn't added like a global one).

It will also give the user the ability to add entity type readers of their own (example: a global IChannel type reader that applies to IMessageChannel etc, like the default one).

## Changes

- Change priority of type readers (user > default).
- Add internal property to type readers to signal a override one and prevent #936 .
- Add `AddEntityTypeReader<T>(Type typeReaderGenericType)` and `AddEntityTypeReader(Type type, Type typeReaderGenericType)` to `CommandService`
- Add logic for entity type readers added by the user (adding a IChannel type reader will apply it to classes/interfaces that implement it, example: IMessageChannel).
- Deprecate `AddTypeReader<T>(TypeReader reader, bool replaceDefault)` and  `AddTypeReader(Type type, TypeReader reader, bool replaceDefault)` (not needed anymore since default type readers don't have priority over the ones added by the user.

## How to use

1. Create a class that inherits from `TypeReader` with a single generic argument that has at least a reference type constraint, e.g. `class CustomUserTypeReader<T> : TypeReader where T : class, IUser`
2. Add it to the command service, e.g. `AddEntityTypeReader<IUser>(typeof(CustomUserTypeReader<>))`
3. Done!

## Examples (with test cases done)

The examples will say what type reader were added (following the order that they were added), the argument used, and the type reader assigned to them. Following this format: `Argument type: Type Reader used`.
Note: default = default Discord.Net type reader, otherwise it's the one added by the user.

### Adding IChannel type reader and IMessageChannel type reader
- IChannel: IChannel
- IMessageChannel: IMessageChannel
- ITextChannel: IMessageChannel
- SocketTextChannel: IMessageChannel
- IVoiceChannel: IChannel
- SocketChannel: IChannel

### Adding only IMessageChannel type reader
- IChannel: default
- IMessageChannel: IMessageChannel
- ITextChannel: IMessageChannel
- SocketTextChannel: IMessageChannel
- IVoiceChannel: default
- SocketChannel: default

### Adding IVoiceChannel type reader and IMessageChannel type reader
- IChannel: default
- IMessageChannel: IMessageChannel
- ITextChannel: IMessageChannel
- SocketTextChannel: IMessageChannel
- IVoiceChannel: IVoiceChannel
- SocketChannel: default

## Notes

This PR includes the change in #1486, so if accepted, please merge that one first since they are the one that fixed and deserve the credit for finding it.

If you have any test case that you want to be done, please comment it here and I can do it since this change might have something I didn't expect and didn't test.

### DIFF
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -49,6 +49,7 @@ namespace Discord.Commands
         private readonly ConcurrentDictionary<Type, ModuleInfo> _typedModuleDefs;
         private readonly ConcurrentDictionary<Type, ConcurrentDictionary<Type, TypeReader>> _typeReaders;
         private readonly ConcurrentDictionary<Type, TypeReader> _defaultTypeReaders;
+        private readonly ConcurrentDictionary<Type, TypeReader> _overrideTypeReaders;
         private readonly ImmutableList<(Type EntityType, Type TypeReaderType)> _entityTypeReaders;
         private readonly HashSet<ModuleInfo> _moduleDefs;
         private readonly CommandMap _map;
@@ -109,6 +110,7 @@ namespace Discord.Commands
             _moduleDefs = new HashSet<ModuleInfo>();
             _map = new CommandMap(this);
             _typeReaders = new ConcurrentDictionary<Type, ConcurrentDictionary<Type, TypeReader>>();
+            _overrideTypeReaders = new ConcurrentDictionary<Type, TypeReader>();
 
             _defaultTypeReaders = new ConcurrentDictionary<Type, TypeReader>();
             foreach (var type in PrimitiveParsers.SupportedTypes)
@@ -348,10 +350,11 @@ namespace Discord.Commands
         /// <param name="reader">An instance of the <see cref="TypeReader" /> to be added.</param>
         public void AddTypeReader(Type type, TypeReader reader)
         {
-            if (_defaultTypeReaders.ContainsKey(type))
-                _ = _cmdLogger.WarningAsync($"The default TypeReader for {type.FullName} was replaced by {reader.GetType().FullName}." +
-                    "To suppress this message, use AddTypeReader<T>(reader, true).");
-            AddTypeReader(type, reader, true);
+            var readers = _typeReaders.GetOrAdd(type, x => new ConcurrentDictionary<Type, TypeReader>());
+            readers[reader.GetType()] = reader;
+
+            if (type.GetTypeInfo().IsValueType)
+                AddNullableTypeReader(type, reader);
         }
         /// <summary>
         ///     Adds a custom <see cref="TypeReader" /> to this <see cref="CommandService" /> for the supplied object
@@ -365,8 +368,9 @@ namespace Discord.Commands
         ///     Defines whether the <see cref="TypeReader"/> should replace the default one for
         ///     <see cref="Type" /> if it exists.
         /// </param>
+        [Obsolete("This method is deprecated. Use the method without the replaceDefault argument.")]
         public void AddTypeReader<T>(TypeReader reader, bool replaceDefault)
-            => AddTypeReader(typeof(T), reader, replaceDefault);
+            => AddTypeReader(typeof(T), reader);
         /// <summary>
         ///     Adds a custom <see cref="TypeReader" /> to this <see cref="CommandService" /> for the supplied object
         ///     type.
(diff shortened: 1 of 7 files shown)


monoclex:
Bump - any status updates on this?

SubZero0:
Had to change the way to add entity readers since **there's no way for type readers to know what kind of object is being expected to be returned**. This could cause an unexpected error when converting types (like a SocketGuildUser to a RestGuildUser).

Now they will need to be added with `AddEntityTypeReader` that has 2 overloads:
- `AddEntityTypeReader<T>(Type typeReaderGenericType)`
- `AddEntityTypeReader(Type type, Type typeReaderGenericType)`

This will make the creation of entity type readers a bit more complex so they can know what kind of Type is being expected.

Entity type readers will need to inherit from TypeReader (like usual) but will required to have a single generic argument that has at least a reference type constraint, e.g. `class CustomUserTypeReader<T> : TypeReader where T : class, IUser`.

To add it to the command service, it'll be needed to pass the generic type definition of the TypeReader, e.g. `AddEntityTypeReader<IUser>(typeof(CustomUserTypeReader<>))`.

Another way would be adding a `Type outputType` to `TypeReader.ReadAsync`, but this would require breaking all current implementations of all type readers when updating, so I chose to just add new methods instead.

SubZero0:
Tagging @Still34 for review 🙃

FiniteReality:
Fix the build errors first.

SubZero0:
Done, missed a bracket in the example.

src/Discord.Net.Commands/CommandService.cs
```diff
@@ -341,17 +350,55 @@ public void AddTypeReader<T>(TypeReader reader)
...
+                AddNullableTypeReader(type, reader);
+        }
+        /// <summary>
+        ///     Adds a custom entity <see cref="TypeReader" /> to this <see cref="CommandService" /> for the supplied
+        ///     object type.
+        /// </summary>
+        /// <example>
+        ///     <para>The following example adds a custom entity reader to this <see cref="CommandService"/>.</para>
+        ///     <code language="cs" region="AddEntityTypeReader"
+        ///           source="..\..\..\Discord.Net.Examples\Commands\CommandService.Examples.cs" />
```
Still34:
```cs

        ///     <code language="cs" region="AddEntityTypeReader"
        ///           source="..\Discord.Net.Examples\Commands\CommandService.Examples.cs" />
```

src/Discord.Net.Commands/CommandService.cs
```diff
@@ -77,6 +78,15 @@ public class CommandService : IDisposable
         /// </summary>
         public ILookup<Type, TypeReader> TypeReaders => _typeReaders.SelectMany(x => x.Value.Select(y => new { y.Key, y.Value })).ToLookup(x => x.Key, x => x.Value);
 
+        /// <summary>
+        ///     Represents all entity type reader <see cref="Type" />s loaded within <see cref="CommandService"/>.
+        /// </summary>
+        /// <returns>
+        ///     A <see cref="ILookup{TKey, TElement}"/> that the key is the object type to be read by the <see cref="TypeReader"/>
```
Still34: Sounds kinda wordy and awkward. `A ILookup<TKey,TElement>; the key is the object type to be read by the TypeReader, while the element is the type of the TypeReader generic definition.` seems slightly better.

src/Discord.Net.Commands/CommandService.cs
```diff
@@ -341,17 +350,55 @@ public void AddTypeReader<T>(TypeReader reader)
...
+        ///     A <see cref="Type" /> that is a generic type definition with a single open argument
+        ///     of the <see cref="TypeReader" /> to be added.
+        /// </param>
+        public void AddEntityTypeReader<T>(Type typeReaderGenericType)
+            => AddEntityTypeReader(typeof(T), typeReaderGenericType);
+        /// <summary>
+        ///     Adds a custom entity <see cref="TypeReader" /> to this <see cref="CommandService" /> for the supplied
+        ///     object type.
+        /// </summary>
+        /// <param name="type">A <see cref="Type" /> instance for the type to be read.</param>
```
Still34, line 381: Add a similar `<example>` for this overload as well?

src/Discord.Net.Commands/CommandService.cs
```diff
@@ -341,17 +350,55 @@ public void AddTypeReader<T>(TypeReader reader)
...
+        ///     object type.
+        /// </summary>
+        /// <example>
+        ///     <para>The following example adds a custom entity reader to this <see cref="CommandService"/>.</para>
+        ///     <code language="cs" region="AddEntityTypeReader"
+        ///           source="..\..\..\Discord.Net.Examples\Commands\CommandService.Examples.cs" />
+        /// </example>
+        /// <typeparam name="T">The object type to be read by the <see cref="TypeReader"/>.</typeparam>
+        /// <param name="typeReaderGenericType">
+        ///     A <see cref="Type" /> that is a generic type definition with a single open argument
```
Still34: Sounds a bit weird 🤔

Still34 (changes requested):

src/Discord.Net.Commands/CommandService.cs
SubZero0: Changed it to `A generic type definition (with one open argument) of the <see cref="TypeReader" />.`. Sounds better?

quinchs:
@Cenngo Can you look at this?

Cenngo:
> @Cenngo Can you look at this?

I need to go over this to see whats the state of this pr but ill try to wrap this up.

PR #1488, opened 2020-04-20
Fallback on the default value on a failed read

Joe4evr:
But only if the command parameter is optional.

Could also be put behind an additional attribute flag so the behavior becomes more explicit.


SubZero0:
I don't like this change because it treats a failure as a success and, imo, an optional is something possible but not necessary but when available should be correct.

Could also be easily solved using a custom type reader that would fit the user needs (another error message or a different approach than the default ones) and wouldn't require to copy-paste supposed error handling code inside the method (since you won't know if the user didn't provide it or it failed parsing).

Joe4evr:
Alright fine, jeez.

PR #1441, opened 2020-01-20
Feature: Add newly supported events and endpoint

sh30801:
# Summary
Implements new event and endpoint support as documented in a discord docs PR [here](https://github.com/discordapp/discord-api-docs/pull/1309).

# Details
This PR introduces a few new events that bots can now interact with:

`InviteCreated` - Fired when an invite is created, emits a `SocketInvite` argument.

`InviteDeleted` - Fired when an invite is deleted, emits a partial `SocketInvite` argument.

`RemovedEmojiReactions` - Fired when all of a specific reaction are deleted, emits a `SocketReaction` argument.

SocketInvite is a new class to handle these invites over the Websocket, mainly due to the properties sent being unique to other invite models.

Additionally, there is an endpoint that can be used to directly cause the `RemovedEmojiReactions` event:

`IMessage.RemoveEmojiReactionsAsync(IEmote)` - Clears the message of all reactions of a specific emote, specified by IEmote.

# Notes
I chose to include the endpoint in the same PR as the events, as the endpoint directly causes the `RemovedEmojiReactions` event, and it is the only way to test it.

PR #1491, opened 2020-04-21
Feature: Add INVITE_CREATE and INVITE_DELETE events

SubZero0:
## Summary

This change will add the new gateway events related to invites, [INVITE_CREATE](https://discordapp.com/developers/docs/topics/gateway#invite-create) and [INVITE_DELETE](https://discordapp.com/developers/docs/topics/gateway#invite-delete), introducing a new class, `SocketInvite`.

## Notes

I'm not sure about three points in this PR that might require more opinions.

- Should I try to update the properties Channel, Guild, and Invite in the Update method?
- I'm unsure about the delete event that just returns the code, so it's not possible to instantiate a complete `SocketInvite` and we don't keep invites in the cache (not sure if we should, since there's no event to update this cache so it'll always be outdated or require to update them manually via REST). Should we cache the invites? Should we change the return of INVITE_DELETE to `SocketInvite` even if it only has the code, channel, and guild?
- Should `SocketInvite` implement `IUpdateable` and make it updateable via REST?

### DIFF
--- a/src/Discord.Net.Core/Entities/Invites/TargetUserType.cs
+++ b/src/Discord.Net.Core/Entities/Invites/TargetUserType.cs
@@ -2,7 +2,7 @@ namespace Discord
 {
     public enum TargetUserType
     {
-        /// <summary> The invite target user type is not defined. </summary>
+        /// <summary> The invite whose target user type is not defined. </summary>
         Undefined = 0,
         /// <summary> The invite is for a Go Live stream. </summary>
         Stream = 1
--- a/src/Discord.Net.Core/Entities/Invites/TargetUserType.cs
+++ b/src/Discord.Net.Core/Entities/Invites/TargetUserType.cs
@@ -3,7 +3,7 @@ namespace Discord
     public enum TargetUserType
     {
         /// <summary> The invite target user type is not defined. </summary>
-        NotDefined = 0,
+        Undefined = 0,
         /// <summary> The invite is for a Go Live stream. </summary>
         Stream = 1
     }
--- a/src/Discord.Net.Core/Entities/Invites/TargetUserType.cs
+++ b/src/Discord.Net.Core/Entities/Invites/TargetUserType.cs
@@ -2,9 +2,13 @@ namespace Discord
 {
     public enum TargetUserType
     {
-        /// <summary> The invite whose target user type is not defined. </summary>
+        /// <summary>
+        ///     The invite whose target user type is not defined.
+        /// </summary>
         Undefined = 0,
-        /// <summary> The invite is for a Go Live stream. </summary>
+        /// <summary>
+        ///     The invite is for a Go Live stream.
+        /// </summary>
         Stream = 1
     }
 }
--- a/src/Discord.Net.Core/Entities/Invites/TargetUserType.cs
+++ b/src/Discord.Net.Core/Entities/Invites/TargetUserType.cs
@@ -1,0 +1,8 @@
+namespace Discord
+{
+    public enum TargetUserType
+    {
+        /// <summary> The invite is for a Go Live stream. </summary>
+        Stream = 1
+    }
+}
--- a/src/Discord.Net.Core/Entities/Invites/TargetUserType.cs
+++ b/src/Discord.Net.Core/Entities/Invites/TargetUserType.cs
@@ -2,6 +2,8 @@ namespace Discord
 {
     public enum TargetUserType
     {
+        /// <summary> The invite target user type is not defined. </summary>
+        NotDefined = 0,
         /// <summary> The invite is for a Go Live stream. </summary>
         Stream = 1
     }
--- a/src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
+++ b/src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
@@ -369,7 +369,9 @@ namespace Discord.WebSocket
         internal readonly AsyncEvent<Func<SocketGroupUser, Task>> _recipientRemovedEvent = new AsyncEvent<Func<SocketGroupUser, Task>>();
 
         //Invites
-        /// <summary> Fired when an invite is created. </summary>
+        /// <summary>
+        ///     Fired when an invite is created.
+        /// </summary>
         /// <remarks>
         ///     <para>
         ///         This event is fired when an invite is created. The event handler must return a
@@ -385,7 +387,9 @@ namespace Discord.WebSocket
             remove { _inviteCreatedEvent.Remove(value); }
         }
         internal readonly AsyncEvent<Func<SocketInvite, Task>> _inviteCreatedEvent = new AsyncEvent<Func<SocketInvite, Task>>();
-        /// <summary> Fired when an invite is deleted. </summary>
+        /// <summary>
+        ///     Fired when an invite is deleted.
+        /// </summary>
         /// <remarks>
         ///     <para>
         ///         This event is fired when an invite is deleted. The event handler must return
--- a/src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
+++ b/src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
@@ -1,3 +1,4 @@
+using Discord.API;
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
@@ -367,5 +368,22 @@ namespace Discord.WebSocket
             remove { _recipientRemovedEvent.Remove(value); }
         }
         internal readonly AsyncEvent<Func<SocketGroupUser, Task>> _recipientRemovedEvent = new AsyncEvent<Func<SocketGroupUser, Task>>();
+
+        //Invites
+        /// <summary> Fired when an invite is created. </summary>
+        public event Func<SocketInvite, Task> InviteCreated
+        {
+            add { _inviteCreatedEvent.Add(value); }
+            remove { _inviteCreatedEvent.Remove(value); }
+        }
+        internal readonly AsyncEvent<Func<SocketInvite, Task>> _inviteCreatedEvent = new AsyncEvent<Func<SocketInvite, Task>>();
+        /// <summary> Fired when an invite is deleted. </summary>
+        /// <remarks> The string is the invite code. </remarks>
+        public event Func<SocketGuildChannel, string, Task> InviteDeleted
+        {
+            add { _inviteDeletedEvent.Add(value); }
+            remove { _inviteDeletedEvent.Remove(value); }
+        }
+        internal readonly AsyncEvent<Func<SocketGuildChannel, string, Task>> _inviteDeletedEvent = new AsyncEvent<Func<SocketGuildChannel, string, Task>>();
     }
 }
--- a/src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
+++ b/src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
@@ -1,4 +1,3 @@
-using Discord.API;
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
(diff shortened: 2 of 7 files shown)


Neuheit:
This is a duplicate of #1441.

src/Discord.Net.WebSocket/Entities/Invites/SocketInvite.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// <inheritdoc />
+        IChannel IInvite.Channel
+        {
+            get
+            {
+                if (Channel != null)
+                    return Channel;
+                throw new InvalidOperationException("Unable to return this entity's parent unless it was fetched through that object.");
+            }
+        }
```
AntiTcb: Should also be obtainable from the cache.

src/Discord.Net.WebSocket/Entities/Invites/SocketInvite.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// <summary>
+        ///     Gets the guild where this invite was created.
+        /// </summary>
+        public SocketGuild Guild { get; private set; }
+        /// <inheritdoc />
+        ChannelType IInvite.ChannelType => throw new NotImplementedException();
+        /// <inheritdoc />
+        string IInvite.ChannelName => throw new NotImplementedException();
+        /// <inheritdoc />
+        string IInvite.GuildName => throw new NotImplementedException();
```
AntiTcb: These props *should* be able to be fulfilled using the cache, and aren't potentially blocked behind something like privileges or needing something from REST, so, should always be available to grab from client state. Can easily grab these from the discord client in the ctor.

AntiTcb (changes requested):

SubZero0:
> This is a duplicate of #1441.

This is another version that the only similarity is the `SocketInvite` name since you yourself said you wouldn't fix your own PR, that has more mistakes than I originally reviewed, and told me to open my own.

Neuheit:
> This is another version that the only similarity is the SocketInvite name since you yourself said you wouldn't fix your own PR, that has more mistakes than I originally reviewed, and told me to open my own.

I said I wouldn't fix it without an official review, as some of your suggested changes are questionable. I'm simply listing my PR as the original.

SubZero0:
> I said I wouldn't fix it without an official review, as some of your suggested changes are questionable. I'm simply listing my PR as the original.

This isn't the first case of multiple implementations of the same feature or change. There's no such thing as "original PR". You could have used the time you used to reply to actually read the changes here and fix your own PR (and test it).

src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
```diff
@@ -367,5 +367,22 @@ public partial class BaseSocketClient
             remove { _recipientRemovedEvent.Remove(value); }
         }
         internal readonly AsyncEvent<Func<SocketGroupUser, Task>> _recipientRemovedEvent = new AsyncEvent<Func<SocketGroupUser, Task>>();
+
+        //Invites
+        /// <summary> Fired when an invite is created. </summary>
```
Joe4evr: @Still34 We've become inconsistent with how the doc comments are written. Should we start asking for one specific formatting, and if so, which one?

src/Discord.Net.WebSocket/Entities/Invites/SocketInvite.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// <inheritdoc />
+        bool IInviteMetadata.IsRevoked => throw new NotImplementedException();
+        /// <inheritdoc />
+        public bool IsTemporary { get; private set; }
+        /// <inheritdoc />
+        int? IInviteMetadata.MaxAge { get => MaxAge; }
+        /// <inheritdoc />
+        int? IInviteMetadata.MaxUses { get => MaxUses; }
+        /// <inheritdoc />
+        int? IInviteMetadata.Uses { get => Uses; }
```
Joe4evr, line 60: 1. Why are *these* properties explicitly implemented?
2. If they really should be, why the extra braces?

Joe4evr (commented):
Looks good, just a few minor questions.

src/Discord.Net.WebSocket/Entities/Invites/SocketInvite.cs
SubZero0: They are required by IInviteMetadata, but shouldn't be optional for the socket entity according to the [api docs](https://discordapp.com/developers/docs/topics/gateway#invite-create) so I added non-nullable ones and redirected these to them.

SubZero0:
I'll be adding the two event fields that are now listed in the official docs api (it was probably updated today), `target_user?` and `target_user_type?`.

src/Discord.Net.WebSocket/Entities/Invites/SocketInvite.cs
Joe4evr: But [the Invite Metadata object](https://discordapp.com/developers/docs/resources/invite#invite-metadata-object) also has them listed as not optional.

SubZero0: For some reason they were [added as nullables](https://github.com/discord-net/Discord.Net/blob/dev/src/Discord.Net.Core/Entities/Invites/IInviteMetadata.cs#L39-L61).
Not sure if it was changed later by discord or it's some workaround for the lib.

Also noticed our IInvite is missing two fields that were added to INVITE_CREATE and that it doesn't match discord, ~~there's no `channel_name` or `channel_type` and~~ `inviter` should be part of `Invite`, not `InviteMetadata`, to follow the [discord object](https://discordapp.com/developers/docs/resources/invite#invite-object).

Might be remaining code trying to stay backward compatible, not sure.

SubZero0: Seeing again, the `channel_name` and `channel_type` are just what discord sends and we use the cache to get the channel, since it gets a partial channel unlike the gateway event.

src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
SubZero0: Changed the summary and remarks so they are more descriptive and easier to understand what's being passed like the Channel and Message events.

src/Discord.Net.Core/Entities/Invites/TargetUserType.cs
```diff
@@ -0,0 +1,8 @@
+namespace Discord
+{
+    public enum TargetUserType
+    {
```
Joe4evr, line 4: Need a 0 value, even if it's something like `NotApplicable`.

src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -1658,7 +1658,15 @@ private async Task ProcessMessageAsync(GatewayOpCode opCode, int? seq, string ty
...
                                         }
 
-                                        var invite = SocketInvite.Create(this, guild, channel, inviter, data);
+                                        SocketUser target = null;
+                                        if (data.TargetUser.IsSpecified)
+                                        {
+                                            target = guild.GetUser(data.TargetUser.Value.Id);
+                                            if (target == null)
+                                                target = SocketUnknownUser.Create(this, State, data.TargetUser.Value);
+                                        }
```
Joe4evr:
```suggestion
                                        SocketUser target =  data.TargetUser.IsSpecified
                                            ? (guild.GetUser(data.TargetUser.Value.Id) ?? SocketUnknownUser.Create(this, State, data.TargetUser.Value))
                                            : null;
```

src/Discord.Net.WebSocket/Entities/Invites/SocketInvite.cs
```diff
@@ -80,22 +80,31 @@ ChannelType IInvite.ChannelType
...
         /// </summary>
         public DateTimeOffset CreatedAt => DateTimeUtils.FromTicks(_createdAtTicks);
+        /// <summary>
+        ///     Gets the user targeted by this invite if available.
+        /// </summary>
+        public SocketUser TargetUser { get; private set; }
+        /// <summary>
+        ///     Gets the type of the user targeted by this invite if available.
+        /// </summary>
+        public TargetUserType? TargetUserType { get; private set; }
```
Joe4evr: Is this still required to be nullable if the enum gets a 0 value?

Joe4evr (changes requested):
I'm a bit more critical about these.

Joe4evr:
Looks good so far. Just wanna hear back from Still about the doc comments, and (presumably fox) about what we're gonna do with those Invite Metadata fields.

src/Discord.Net.Core/Entities/Invites/TargetUserType.cs
```diff
@@ -0,0 +1,10 @@
...
+        /// <summary> The invite target user type is not defined. </summary>
```
Still34: `The invite whose target user type is not defined.`

src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
```diff
@@ -367,5 +367,43 @@ public partial class BaseSocketClient
...
+        /// <remarks>
+        ///     <para>
+        ///         This event is fired when an invite is created. The event handler must return a
+        ///         <see cref="Task"/> and accept a <see cref="SocketInvite"/> as its parameters.
```
Still34: `its parameter` singular

src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
Still34: > @Still34 We've become inconsistent with how the doc comments are written. Should we start asking for one specific formatting, and if so, which one?

I might be late to the party but I'm not spotting the inconsistency?

src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
```diff
@@ -367,5 +367,43 @@ public partial class BaseSocketClient
...
+            add { _inviteCreatedEvent.Add(value); }
+            remove { _inviteCreatedEvent.Remove(value); }
+        }
+        internal readonly AsyncEvent<Func<SocketInvite, Task>> _inviteCreatedEvent = new AsyncEvent<Func<SocketInvite, Task>>();
+        /// <summary> Fired when an invite is deleted. </summary>
+        /// <remarks>
+        ///     <para>
+        ///         This event is fired when an invite is deleted. The event handler must return
+        ///         a <see cref="Task"/> and accept a <see cref="SocketGuildChannel"/> and
+        ///         <see cref="string"> as its parameters.
```
Still34: singular `parameter`

Still34: `<see>` tag is not closed; this will cause docfx to discard the entire XML doc string.

src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
```diff
@@ -367,5 +367,43 @@ public partial class BaseSocketClient
...
+        ///     </para>
+        ///     <para>
+        ///         The channel where this invite was will be passed into the <see cref="SocketGuildChannel"/> parameter.
```
Still34: `was created`?

Still34 (changes requested):
Overall looks fine to me, but the small inconsistencies and the unclosed `<see>` tag must be fixed before merge.

src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
Joe4evr: We have `<summary> Text </summary>` (one line) and
```
<summary>
    Text
</summary>
```
(multiple lines, indented) both used in this PR alone.

Still34: Personally, I'm leaning towards the latter. Although it takes up more space, it is easier to read when writing documentation. Perhaps we could open a new issue and or PR that addresses this in our contribution guide.

SubZero0: Changed all summaries to use the latter.

Still34 (approved):

Joe4evr (approved):

PR #1503, opened 2020-04-25
Move content check for ModifyMessage

### LINKED ISSUE #1501: Can't set a message's content to null/empty string with ModifyAsync even if that message has an embed
Discord.Net 2.2.0

Steps:

1. Create a message with an embed and content.
2. Call ModifyAsync on that message, and set the content to string.Empty.

SubZero0:
## Summary

It's not possible to know if the message has an embed without the message cache, so we shouldn't assume or block the user from setting the content to null or empty, it should be done only when we have it.

Fix #1501 

## Changes

- Moved content check from `DiscordRestApiClient` to `MessageHelper`.
- Added a check for content and embed, not having both will preemptively throw an exception.


Joe4evr (approved):
Looks simple enough.

Issue #1439, opened 2020-01-17
Thread pool gets blocked inside rate limit code


xSke:
A while back I spent several days debugging a strange issue in my bot (~15k guilds, relies on reactions quite a lot), where all the worker threads would one by one get blocked until the entire bot would crash once they're all blocked. After... too much digging around, I figured out all the threads were getting blocked inside the `EnterAsync` function in `Discord.Net.Rest/Net/Queue/RequestQueueBucket.cs`.

I'm not sure exactly *what* was going wrong, but the while loop never exited even as the rate limit bucket hit its reset time, and it just sat and spun forever. I "solved" the issue by adding a return from the method if the current time is past the reset time, but I'm not sure if this is the *correct* solution, hence the issue rather than a PR. This gets easier to reproduce locally by setting [the min/max thread count to 1](https://docs.microsoft.com/en-us/dotnet/api/system.threading.threadpool.setminthreads?view=netframework-4.8#System_Threading_ThreadPool_SetMinThreads_System_Int32_System_Int32_), and firing off some reaction-based commands (they seem to make it occur faster, unsure if specifically those, though). At default ThreadPool settings, it only seems to happen at a relatively large scale (around the ~15k guilds previously mentioned).

This is the single-line patch I made on my personal fork: https://github.com/xSke/Discord.Net/commit/2aaa5e7ce50f056969e4b000386fc9bc336880a5

I hope someone more well-versed with D.NET internals can help figure out what's actually going on, and solve the issue in the main repository. If it helps, the server was running .NET Core 3.1 on Ubuntu 18.04 inside a Docker container.

BrammyS:
I'm running into the same problem.

goaaats:
Also running into this! Triggered by queuing a lot of messages into the same channel asynchronously.

Lymdun:
Same here.

Kein:
Huh, I suppose this what happened to me as well (judging from my limited knowledge of debugging async core apps)

FiniteReality:
Thanks for the detailed investigation. As you have already mentioned, the Discord.Net internals are a pain in the arse, in particular our ratelimiting code. It will likely take some time for us to fix this properly, as debugging threadpool exhaustion like this is not easy.

kirsan31:
Any news?

turulix:
As far as my research goes it is possible to "DDOS" a bot using this. The Bot will not respond to any commands on other guilds or take very long to respond.

[...]

PR #1511, opened 2020-05-04
fix: avoid infinite loop on rate limit

Lymdun:
Source: #1439 
Due to the current situation, even though this fix may not be the perfect one, it is still better than nothing.


FiniteReality:
I disagree with the sentiment that a patch we can't verify is better than no patch at all. As part of the work I started yesterday to investigate removing the ConnectionManager type, I'll probably look into this too.

FiniteReality:
I forgot to close this, but as I've already pushed some fixes for #1439, this should in theory not be needed.

PR #1525, opened 2020-05-12
fix: AllowedMentions and AllowedMentionTypes

SubZero0:
## Summary

Currently the values for the flag enum `AllowedMentionTypes` aren't there, so they won't work as expected.

## Changes

- Changed member values
- Added None as 0
- Initialized `UserIds` and `RoleIds` lists

### DIFF
--- a/src/Discord.Net.Core/Entities/Messages/AllowedMentionTypes.cs
+++ b/src/Discord.Net.Core/Entities/Messages/AllowedMentionTypes.cs
@@ -9,10 +9,13 @@ namespace Discord
     public enum AllowedMentionTypes
     {
         /// <summary>
-        ///     No flag was set.
+        ///     No flag is set.
         /// </summary>
         /// <remarks>
-        ///     It is not used to control mentions and does not mean no mentions will be allowed.
+        ///     This flag is not used to control mentions.
+        ///     <note type="warning">
+        ///         It will always be present and does not mean mentions will not be allowed.
+        ///     </note>
         /// </remarks>
         None        = 0,
         /// <summary>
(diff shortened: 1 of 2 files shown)


Chris-Johnston (approved):

src/Discord.Net.Core/Entities/Messages/AllowedMentionTypes.cs
```diff
@@ -8,17 +8,24 @@ namespace Discord
     [Flags]
     public enum AllowedMentionTypes
     {
+        /// <summary>
+        ///     No flag was set.
+        /// </summary>
+        /// <remarks>
+        ///     It is not used to control mentions and does not mean no mentions will be allowed.
+        /// </remarks>
+        None        = 0,
```
Joe4evr, line 20: I believe `Default` would be a better name in this case. And the summary could be something along the lines of `The default set of Allowed Mentions` (which just happens to be all, iirc).

SubZero0: The default behavior (`new AllowedMentions()` or `new AllowedMentions(AllowedMentionTypes.None)`) is none (because an empty `parse` array means no mentions allowed), so, while not intentionally trying to, `AllowedMentionTypes.None` would be none if there's no OR.

Joe4evr: Oh yeah, my b. In that case `None` is fine.

Joe4evr:
https://github.com/discord-net/Discord.Net/blob/0c16d2f538a7cb9a683800014881dc60ec34dbb3/src/Discord.Net.Core/Entities/Messages/AllowedMentions.cs#L42-L50
These lists should be given an initializer, that way the user could write, for example:
```cs
new AllowedMentions
{
    UserIds = {
        123456789123456789ul
    }
};
```

Joe4evr (approved):

foxbot (approved):

PR #1526, opened 2020-05-13
feature: Add Direction.Around to GetMessagesAsync

SubZero0:
## Summary

This PR will add the missing implementation for Direction.Around.
If the limit is inside the max for discord, it will use "around", otherwise it'll query before and after since discord doesn't give any other ways to do it.

## Changes

- Add Direction.Around logic for cache and rest
- Do not return empty pages if cache doesn't have any
- Use cache for Direction.After when possible


SubZero0:
Did a few more tests using all directions with and without cache, seems to be working as expected now.

foxbot (approved):
This feels like a hacky way to support this, but looks good otherwise. Thanks!

PR #1537, opened 2020-05-18
feature: Implement gateway ratelimit

### LINKED ISSUE #1442: Clients don't respect the gateway identify rate-limit.
This causes bots with more than 1 gateway connection to get instantly disconnected upon attempting to identify. The more shards, the more disconnections this causes.

To fix this, DiscordShardedClient and individual clients/shards should respect the 1/5s identify limit with a shared connection lock that would properly delay subsequent identify attempts.
#519 is somewhat related.

SubZero0:
## Summary

This PR aims to add a ratelimiter to the gateway that currently has none except a `Task.Delay(5000)`.
It follows what's described in the [discord api docs](https://discord.com/developers/docs/topics/gateway#rate-limiting).

Fixes #1442 

## Changes

- Add a new type of bucket, `GatewayBucket`
- Add `IdentifyMaxConcurrency` property to `DiscordSocketConfig`
- Add `GetBotGatewayAsync` to `IDiscordClient`
- Add `session_start_limit` to `GetBotGatewayResponse`
- Sharded client will automatically deal with concurrent identify requests when possible


src/Discord.Net.Rest/Entities/Gateway/GatewayLimit.cs
```diff
@@ -0,0 +1,23 @@
+namespace Discord.Rest
+{
+    /// <summary>
+    ///     Represents the limits for a gateway request.
+    /// </summary>
+    public struct GatewayLimit
+    {
+        /// <summary>
+        ///     The maximum amount of this type of request in a time window, that is set by <see cref="Seconds"/>.
```
Still34: Start with `Gets or sets` as the verb.

src/Discord.Net.Rest/Entities/Gateway/GatewayLimit.cs
```diff
@@ -0,0 +1,23 @@
...
+        /// </summary>
+        public int Count { get; set; }
+        /// <summary>
+        ///     The amount of seconds until the rate limiter resets the remaining requests <see cref="Count"/>.
```
Still34: Ditto.

src/Discord.Net.Rest/Entities/Gateway/GatewayLimits.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// <summary>
+    ///     Contains the rate limits for the gateway.
+    /// </summary>
+    public class GatewayLimits
+    {
+        /// <summary>
+        ///     Gets or sets the global limits for the gateway rate limiter.
+        /// </summary>
+        /// <remarks>
+        ///     It includes all the other limits, like Identify.
```
Still34: Use `This property` instead of `it`.

src/Discord.Net.Rest/Net/Queue/GatewayBucket.cs
```diff
@@ -0,0 +1,62 @@
+using Discord.Rest;
+using System.Collections.Immutable;
+
+namespace Discord.Net.Queue
+{
+    public enum GatewayBucketType
```
Still34, line 5: Should this be in its own class?

src/Discord.Net.WebSocket/DiscordSocketConfig.cs
```diff
@@ -124,6 +125,16 @@ public class DiscordSocketConfig : DiscordRestConfig
         /// </summary>
         public bool GuildSubscriptions { get; set; } = true;
 
+        /// <summary>
+        ///     Gets or sets the gateway limits.
+        ///     <note type="warning">
```
Still34: Move this to `<remarks>`. Replace `It` with `This property`.

src/Discord.Net.Rest/Entities/Gateway/GatewayLimits.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// </remarks>
+        public GatewayLimit Global { get; set; }
+        /// <summary>
+        ///     Gets or sets the limits of Identify requests.
+        /// </summary>
+        public GatewayLimit Identify { get; set; }
+
+        public GatewayLimits()
```
Still34: Ctor should include a summary of `Initializes a new...`

Still34 (changes requested):

src/Discord.Net.Rest/Net/Queue/GatewayBucket.cs
SubZero0: I followed the same design as `ClientBucket` that has `ClientBucketType` in the same file, should that be changed?

src/Discord.Net.WebSocket/DiscordSocketConfig.cs
SubZero0: I believe this warning should stay in the summary because VS doesn't show remarks (or it's disabled by default) and wanted to make who is using the lib aware that they shouldn't change the limits unless they know what they are doing, it might have really bad consequences.

src/Discord.Net.Rest/Net/Queue/GatewayBucket.cs
Still34: Requesting input from other contributors.
cc @foxbot @Joe4evr

src/Discord.Net.Rest/Entities/Gateway/GatewayLimit.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    /// <summary>
+    ///     Represents the limits for a gateway request.
+    /// </summary>
+    public struct GatewayLimit
+    {
+        /// <summary>
+        ///     Gets or sets the maximum amount of this type of request in a time window, that is set by <see cref="Seconds"/>.
+        /// </summary>
+        public int Count { get; set; }
```
Still34: Might want to be clear with what the `int` here is with `<returns>`.

Still34 (changes requested):

src/Discord.Net.Rest/Net/Queue/GatewayBucket.cs
foxbot: the ClientBucket file is pretty weird as it is, I would just leave the same structure for consistency

src/Discord.Net.WebSocket/DiscordSocketConfig.cs
Joe4evr: VS 16.6 released just the other day which shows the Remarks section in Quick Info's. I didn't need to change any settings for that to come up.

src/Discord.Net.Rest/Net/Queue/GatewayBucket.cs
Joe4evr: Idk about this, so I defer to fox's judgement.

src/Discord.Net.WebSocket/DiscordSocketConfig.cs
SubZero0: Updated from 16.5 and it seems they added remarks, returns, parameters, etc to 16.6. Going to move it to remarks.

SubZero0:
I'll need to do a few more changes to this PR

SubZero0:
After a bit more digging, I learned that the global queue is per-session, only Identify is per-account, and there's a "soft" presence update limit, so I changed my approach taking this into consideration.

- Added `IdentifySemaphoreName` and `PresenceUpdate` to `GatewayLimits` 
- Gateway will use the same `RequestQueue` as REST per client, won't be sharing one anymore
- Identify will use a named semaphore (the name is changeable at `GatewayLimits`), this will make possible to share the same semaphore between clients with or without a sharded client

I also noticed (if I understood it correctly) that if the user hits the gateway ratelimit, it won't be able to send heartbeats and might end losing the gateway connection either way. It would be possible to create a reserved limit for it but do we want to interfere this much? @foxbot

SubZero0:
While browsing the discord api docs github, I noticed this pull request: [https://github.com/discord/discord-api-docs/pull/1621](https://github.com/discord/discord-api-docs/pull/1621)

It seems discord provides a key that allows multiple shards to concurrently identify in [Get Gateway Bot](https://discord.com/developers/docs/topics/gateway#get-gateway-bot).

I'll see if there's a better way to implement it.

SubZero0:
After some talk and more consideration, I reached the conclusion that Discord wouldn't change specific limits for the gateway since what needs to be changed is sent by REST (max concurrency) or can be solved by adding more shards (gateway global limits).

Taking this into consideration, I removed `GatewayLimits`.
I left the semaphore-related properties in `DiscordSocketConfig`, I'm also considering removing these since the limited use-case would be automatic identify control by independent `DiscordSocketClient`s (without a `DiscordShardedClient` as mediator).

SubZero0:
I decided to remove the named semaphore since it would add a lot more complexity. If someone is sharding for themselves, they have entire control when a client is going to identify.

Now only `IdentifyMaxConcurrency` is present at the config, since it's needed to create the "slave semaphores" for concurrent identify when automatic sharding is disabled.

PR #1548, opened 2020-05-24
fix: Incomplete Ready, DownloadUsersAsync, and optimize AlwaysDownloadUsers

SubZero0:
## Summary

There were three issues related to how these were being done:
- It was possible to Ready to fire before downloading all guild data because some large guilds can take more than two seconds to be received so it will assume it's over. I was able to always reproduce this locally with Discord API (the guild) with a small bot, would take a lot more time.
- AlwaysDownloadUsers would fire one gateway request to download members per guild, making it easily reach the gateway limit of 120 requests per minute, when this request, without gateway intents, can receive an array of `guild_id`s (tested up to 200, but limited to 100 since it isn't documented).
- The wrong variable was being used to calculate `count` in `DownloadUsersAsync`, returning a negative number.

## Changes

- Add property `MaxWaitBetweenGuildAvailablesBeforeReady` to `DiscordSocketConfig` that will let the user decide how long they want to wait, zero being an option (firing READY without waiting GUILD_AVAILABLEs).
- Change default wait time to 10,000ms instead of 2,000ms (large guilds can take more time than 2 seconds) to avoid firing READY before all guild data is downloaded.
- Adjust the maximum batch size for a `Request Guild Members` gateway request (tested up to 200 and it still worked fine).
- Fixed how `count` is calculated in `DownloadUsersAsync`

### DIFF
--- a/src/Discord.Net.WebSocket/DiscordSocketClient.Events.cs
+++ b/src/Discord.Net.WebSocket/DiscordSocketClient.Events.cs
@@ -21,7 +21,12 @@ namespace Discord.WebSocket
             remove { _disconnectedEvent.Remove(value); }
         }
         private readonly AsyncEvent<Func<Exception, Task>> _disconnectedEvent = new AsyncEvent<Func<Exception, Task>>();
-        /// <summary> Fired when guild data has finished downloading. </summary>
+        /// <summary>
+        ///     Fired when guild data has finished downloading.
+        /// </summary>
+        /// <remarks>
+        ///     It is possible that some guilds might be unsynced if <see cref="DiscordSocketConfig.MaxWaitBetweenGuildAvailablesBeforeReady" /> was not long enough to receive all GUILD_AVAILABLES before READY.
+        /// </remarks>
         public event Func<Task> Ready
         {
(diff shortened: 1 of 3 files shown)


src/Discord.Net.WebSocket/DiscordSocketClient.Events.cs
```diff
@@ -21,7 +21,12 @@ public partial class DiscordSocketClient
             remove { _disconnectedEvent.Remove(value); }
         }
         private readonly AsyncEvent<Func<Exception, Task>> _disconnectedEvent = new AsyncEvent<Func<Exception, Task>>();
-        /// <summary> Fired when guild data has finished downloading. </summary>
+        /// <summary>
+        ///     Fired when guild data has finished downloading.
+        /// </summary>
+        /// <remarks>
+        ///     It is possible that some guilds might be unsynced if <see cref="DiscordSocketConfig.MaxWaitBetweenGuildAvailablesBeforeReady" /> was not long enough to receive all GUILD_AVAILABLEs before READY.
```
Still34: Best to reflow/make new line per 70 chars

foxbot (approved):

PR #1562, opened 2020-06-09
Fixes OperationCancelledException and adds IAsyncEnumerable to wait without thread blocking

Jagreaper:
This pull requests addresses an OperationCancelledException exception that can occurred when exiting execution of a program after immediately disposing a DiscordSocketClient instance by waiting for the orphaned task to finish its execution in the dispose method.

Because the solution requires that the task be waited on I have also added a basic implementation of the IAsyncDisposable interface from the RequestQueue up to the DiscordSocketClient and other relevant classes to allow developers to reduce thread blocking


gragra33:
Any chance of getting this merged so we don't have to use a forked version?

AraHaan:
when can this be merged?

quinchs (approved):

PR #1569, opened 2020-06-17
Fix: Update nuget packages.

### LINKED ISSUE #1568: Update nuget packages.
Pretty straightforward, I tested with a local clone of Discord.Net and seems that my local changes which updates the nuget packages to the latest stable (and a try with prerelease) both changes work perfectly fine with no issues at all.

As such I think it is safe to update them.
(Will link pull request later to this issue).

AraHaan:
I decided to update the nuget packages here as they seem to work well at the moment locally as well anyway. Plus I see no harm in updating the analyzer as well so updated that as well.

Note: I locally use the pre-release packages from the dotnet5 feeds and stuff and as such I had to ensure that I use the latest stables instead on this pull request. Local copy works fine with the dotnet 5 preview packages though.
Fixes #1568
Changes:
- latest stable nuget packages from the nuget feeds.
- Replaced a few packages with other packages.
- Remove OS check on ``TargetFrameworks`` and instead only reference the nuget package for the .NET Framework reference assemblies with an ``TargetFramework`` that is .NET Framework is being compiled, it also makes them able to compile on non-windows os removing the need for the OS check on the project file.
- Added check on the CI to cleanly state if any packages are outdated in the solution file's projects.
- Added .NET 5.0 to the ``TargetFrameworks`` since some people would want to target Discord bots to that framework in the future possibly to use updated packages that resolves security issues that dropped support to anything older than .NET 5.0.

The reason I added .NET 5.0 support is because I do not know if the .NET 5.0 SDK would recognize Discord.NET as compatible with it at all.


src/Discord.Net.Core/Discord.Net.Core.csproj
```diff
@@ -8,10 +8,10 @@
     <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netstandard2.0;netstandard2.1</TargetFrameworks>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="12.0.2" />
-    <PackageReference Include="System.Collections.Immutable" Version="1.3.1" />
-    <PackageReference Include="System.Interactive.Async" Version="4.0.0" />
-    <PackageReference Include="IDisposableAnalyzers" Version="2.1.2">
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
+    <PackageReference Include="System.Collections.Immutable" Version="1.7.1" />
+    <PackageReference Include="System.Interactive.Async" Version="4.1.1" />
```
RheaAyase: `System.Interactive.Async` should not be referenced anymore, it overwrites `IAsyncEnumerable` references and causes problems. `IAsyncEnumerable` is loaded from `Microsoft.Bcl.AsyncInterfaces`.

AraHaan: So Discord.net should not reference it at all?

FiniteReality: This is false - System.Interactive.Async is fine to reference, since it depends on System.Linq.Async which depends on Microsoft.Bcl.AsyncInterfaces as necessary.

RheaAyase: Indeed D.NET should not reference it. It's causing easily googled wide spread issues, e.g. https://github.com/dotnet/efcore/issues/15047 (One would have to read into the comments and other "duplicate" issues to understand it.)

TLDR referencing this is obsolete, it should **no longer** be referenced for `IAsyncEnumerable`. I guess it becomes a question of what's the lowest supported version (and only reference it if really necessary)

Reference often results in `System.TypeLoadException: Could not load type 'System.Collections.Generic.IAsyncEnumerable'1' from assembly 'System.Interactive.Async`

liny-fox:
Is there any update on this? A release with #1571 merged in is getting closer to needing to be released before all stable releases become unusable. The solution of getting the latest nightly from myget is not a solution for those that want only release versions.

AraHaan:
I could update this again.

src/Discord.Net.Core/Discord.Net.Core.csproj
AraHaan: Odd I never had issues like that before on Windows. Maybe because you use a different platform?

For me it just uses the highest version of the assemblies even if their are other assemblies that calls for an older version of it.

AraHaan:
It seems the example, and the analyzer project as well as the test for that analyzer project in this repository needs an update to some packages which I can do as well.

AraHaan:
Welp the warning when ``CheckEolTargetFramework`` defaults to true is annoying me, now I know how to turn it off 👀.

src/Discord.Net.Examples/Discord.Net.Examples.csproj
```diff
@@ -15,7 +16,7 @@
   <ItemGroup>
     <ProjectReference Include="..\Discord.Net.Core\Discord.Net.Core.csproj" />
     <ProjectReference Include="..\Discord.Net.WebSocket\Discord.Net.WebSocket.csproj" />
-    <PackageReference Include="JetBrains.Annotations" Version="2019.1.3" />
+    <PackageReference Include="JetBrains.Annotations" Version="2020.1.0" />
```
AraHaan: Isn't this a good analyzer that could be enabled on all of the Discord.net projects instead of just this one?

Edit: discovered it's not an analyzer.

src/Discord.Net.Core/Discord.Net.Core.csproj
AraHaan: @RheaAyase ok, I checked further and it seems to work on that package since .NET Core 2.1 with the 5.0.0 version of them.

However oddly enough some 5.0.0 packages on my bot not the case, I also think I should add ``net5.0`` target frameworks to Discord.NET since I would need that target framework on my bot and for those users who wants to use net5.0 as well for updated things like nullable support.

AraHaan:
Well shit looks like the other package breaks it.

AraHaan:
@RheaAyase, @FiniteReality should I open up an follow up pr for net5.0 support as well?

AraHaan:
There it all seems to work now.

quinchs:
This was added in 3.0 of dnet. ref #1923

PR #1570, opened 2020-06-19
feature: Support to multiple files for webhooks

### LINKED ISSUE #1538: Webhook add support for multiple file upload
Webhooks support uploading multiple files.
See https://birdie0.github.io/discord-webhooks-guide/structure/file.html for example.

Please add a method to the WebhookClient that allows multiple files.

SubZero0:
## Summary

Adds the possibility to upload multiple files with a single request.
I'm not sure if this should really be added since, while it works, it isn't something official (see [https://github.com/discord/discord-api-docs/issues/1552#issuecomment-620511268](https://github.com/discord/discord-api-docs/issues/1552#issuecomment-620511268)), so I'll leave the PR here and maybe get more input about it.

Fixes #1538 

## Changes
- Add `SendFileAsync(IEnumerable<string> filePaths, string text, bool isTTS = false, IEnumerable<Embed> embeds = null, string username = null, string avatarUrl = null, RequestOptions options = null, bool isSpoiler = false)` to `DiscordWebhookClient`
- Add `SendFileAsync(IEnumerable<KeyValuePair<string, Stream>> streams, string text, bool isTTS = false, IEnumerable<Embed> embeds = null, string username = null, string avatarUrl = null, RequestOptions options = null, bool isSpoiler = false)` to `DiscordWebhookClient`

### DIFF
--- a/src/Discord.Net.Rest/API/Rest/UploadWebhookFileParams.cs
+++ b/src/Discord.Net.Rest/API/Rest/UploadWebhookFileParams.cs
@@ -12,9 +12,8 @@ namespace Discord.API.Rest
     {
         private static JsonSerializer _serializer = new JsonSerializer { ContractResolver = new DiscordContractResolver() };
 
-        public Stream File { get; }
+        public IEnumerable<KeyValuePair<string, Stream>> Files { get; }
 
-        public Optional<string> Filename { get; set; }
         public Optional<string> Content { get; set; }
         public Optional<string> Nonce { get; set; }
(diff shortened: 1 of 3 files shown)


quinchs:
With the support of bots uploading multiple files and file descriptions I think webhooks should work off of the same principle with the [`FileAttachment`](https://github.com/discord-net/Discord.Net/blob/dev/src/Discord.Net.Core/Entities/Messages/FileAttachment.cs) struct instead of paths / streams as parameters. Since this PR is out of date I'll open a new one with the changes

PR #1578, opened 2020-06-21
fix: Trim token before passing it to the authorization header

### LINKED ISSUE #1577: Token validation and error 411: LengthRequired
For those that are lazy, there's a tl;dr.

## Issue

While reading the Discord.Net channel in Discord API, I saw someone asking why their `AddRoleAsync` was throwing a `The server responded with error 411: LengthRequired`, later they found out the reason: there's a new line in the file that they stored their token.

I thought it was odd they actually were able to log in on Rest and Gateway without issues and tried to test it myself and tested that Strings that end, start, or have `\r`, `\n`, or a space would throw an exception when validating but Discord.Net won't stop the program flow. That would be followed by Discord failing the Gateway authorization.

But there was a case where it wouldn't. If the token ended with `\r\n`, it would sucessfully connect to the Gateway as well, but when doing a POST request, it would throw that error (probably because the line break malformed the request).

So while they indeed missed the first warning from the token validation, I think it would be optimal to either fix this inside Discord.Net or add it somewhere in the documentation (example: common case for an error 411).

## Tl;dr
Tokens that end with `\r\n` will succesfully log in to the Gateway and do GET requests, but throw an error 411: LengthRequired when doing other http methods like POST. This isn't expected or intuitive since it would be expected a failed authorization for both.

## Fix

So I thought about a few ways to fix it:
1. Block the program flow when the token validation fails with an invalid character;
2. Trim the token (basically silently fix it);
3. Add it somewhere to the documentation since the error isn't really intuitive.

Any ideas or other suggestions?

SubZero0:
## Summary

Prevents the client from logging in successfully but throwing an exception when doing http methods that aren't GET with an `error 411: LengthRequired` that isn't intuitive to why it happened since it can occur long after the client started.

Fixes #1577

### DIFF
--- a/src/Discord.Net.Rest/DiscordRestApiClient.cs
+++ b/src/Discord.Net.Rest/DiscordRestApiClient.cs
@@ -82,9 +82,9 @@ namespace Discord.API
         {
             return tokenType switch
             {
-                default(TokenType) => token?.TrimEnd(),
-                TokenType.Bot => $"Bot {token?.TrimEnd()}",
-                TokenType.Bearer => $"Bearer {token?.TrimEnd()}",
+                default(TokenType) => token,
+                TokenType.Bot => $"Bot {token}",
+                TokenType.Bearer => $"Bearer {token}",
                 _ => throw new ArgumentException(message: "Unknown OAuth token type.", paramName: nameof(tokenType)),
             };
         }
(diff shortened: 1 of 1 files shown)


Emzi0767:
Instead of trimming it on every use, why not trim it in the config? This way the operation happens once.

SubZero0:
True, it could be done when assigning it to the `AuthToken` instead.

Issue #1572, opened 2020-06-19
Another disconnection issue


panthernet:
Recent connection issues on my server is killing my app connection to Discord to the point it is not reestablishing. I've already had to relaunch the app like 5 times or so.

I'm not sure if this is a bug or by design. Is there a `solid` way to ensure that lib will keep try to connect with Discord? Should I implement some kind of wrapper to reinitialize Discord connection?

```
06/19/2020 12:10:51 [    Info]: Connecting
06/19/2020 12:12:26 [ Warning]: 
06/19/2020 12:12:26 [    Info]: Disconnecting
06/19/2020 12:12:26 [Critical]: Disconnected! 
System.TimeoutException: The operation has timed out.
   at Discord.ConnectionManager.WaitAsync() in D:\Projects\EVE\ThunderED\Discord.Net-dev\src\Discord.Net.WebSocket\ConnectionManager.cs:line 176
   at Discord.WebSocket.DiscordSocketClient.OnConnectingAsync() in D:\Projects\EVE\ThunderED\Discord.Net-dev\src\Discord.Net.WebSocket\DiscordSocketClient.cs:line 249
   at Discord.WebSocket.DiscordSocketClient.OnConnectingAsync() in D:\Projects\EVE\ThunderED\Discord.Net-dev\src\Discord.Net.WebSocket\DiscordSocketClient.cs:line 261
   at Discord.ConnectionManager.ConnectAsync(CancellationTokenSource reconnectCancelToken) in D:\Projects\EVE\ThunderED\Discord.Net-dev\src\Discord.Net.WebSocket\ConnectionManager.cs:line 144
   at Discord.ConnectionManager.<>c__DisplayClass29_0.<<StartAsync>b__0>d.MoveNext() in D:\Projects\EVE\ThunderED\Discord.Net-dev\src\Discord.Net.WebSocket\ConnectionManager.cs:line 72
06/19/2020 12:12:26 [    Info]: Disconnected
06/19/2020 12:12:35 [    Info]: Connecting
06/19/2020 12:12:35 [ Warning]: Failed to resume previous session
06/19/2020 12:12:36 [    Info]: Connected
06/19/2020 12:12:36 [    Info]: Ready
06/19/2020 12:12:36 [ Warning]: Rate limit triggered: guilds/415097131184160779/members/@me/nick
06/19/2020 12:27:43 [ Warning]: 
06/19/2020 12:27:44 [    Info]: Disconnecting
[...]

PR #1580, opened 2020-06-24
fix: Possible fix to one cause of the "connection deadlock"

SubZero0:
Firsts things first: since this issue isn't easily reproducible, I based this fix on a stacktrace provided in one issue, in this case, #1572 . So there's a few things to consider:
1) It might not fix all possible causes to the connection deadlock;
2) It might have an unintended side effect that I didn't notice/predict (I'll explain at the end).

## The issue

I read the stacktrace from #1572 again and again and came to the conclusion that what happens is due to an exception thrown from a rest request being cancelled and reaching the ConnectionManager that will permanently disconnect due to it.

It goes as this:
1) DiscordSocketClient.StartAsync
2) ConnectionManager.StartAsync
3) ConnectionManager: `await ConnectAsync(reconnectCancelToken).ConfigureAwait(false);`
4) ConnectionManager: `await _onConnecting().ConfigureAwait(false);` (that is actually DiscordSocketClient.OnConnectingAsync)
5) DiscordSocketClient: `await ApiClient.ConnectAsync().ConfigureAwait(false);`
6) DiscordSocketApiClient: ConnectAsync
7) DiscordSocketApiClient: ConnectInternalAsync
8) ConnectInternalAsync: `var gatewayResponse = await GetGatewayAsync().ConfigureAwait(false);` fails and throws a `TaskCanceledException` (could be any other rest request in DiscordSocketClient.OnConnectingAsync)
9) TaskCanceledException bubbles up until it's back to the ConnectionManager.StartAsync
10) StartAsync will finally deal with the exception, since `TaskCanceledException` inherits from `OperationCanceledException`, it'll end in this catch:
```cs
catch (OperationCanceledException ex)
{
    Cancel(); //In case this exception didn't come from another Error call
    await DisconnectAsync(ex, !reconnectCancelToken.IsCancellationRequested).ConfigureAwait(false);
}
```
This catch will cancel all tokens in `Cancel()`, including the reconnect, so it'll just Disconnect and stop there.

## Solution

I decided to try-catch the `await _onConnecting().ConfigureAwait(false);` to prevent any `TaskCanceledException` from bubbling up (and disconnecting the client permanently) and sent the inner exception instead (added a null check to prevent throwing "null", but I didn't also see a reason why that would happen, so it's there just to be safe).

Now this could cause an unintended side effect. I wasn't able to identify a method would legitly throw that specific exception, so I don't believe it'll happen but it could if I missed it, and end not disconnecting permanently the client as expected.

If anyone that has this issue recurrently could test this PR, it would be highly appreaciated, same goes if you see something I missed that could potentially cause an unintended consequence.

### DIFF
--- a/src/Discord.Net.WebSocket/ConnectionManager.cs
+++ b/src/Discord.Net.WebSocket/ConnectionManager.cs
@@ -141,7 +141,16 @@ namespace Discord
                     catch (OperationCanceledException) { }
                 });
 
-                await _onConnecting().ConfigureAwait(false);
+                try
+                {
+                    await _onConnecting().ConfigureAwait(false);
+                }
+                catch (TaskCanceledException ex)
+                {
+                    Exception innerEx = ex.InnerException ?? new OperationCanceledException("Failed to connect.");
+                    Error(innerEx);
+                    throw innerEx;
+                }
 
                 await _logger.InfoAsync("Connected").ConfigureAwait(false);
                 State = ConnectionState.Connected;


xXBuilderBXx:
I have been getting some weird disconnection/timeout issues from the lib so i'm gonna test this on my big bot and see how well it handles will let you know if there is any issues.

AraHaan:
Nice, I have also identified another issue as well, but good work, this option is also reproduceable in the log event.

1. Like for example if you use the Log event to ``AppendAllTextAsync`` (from System.IO.File) and you do not handle the case where the file is in use (apperently from opening it too fast every line it tries to log making it throw an System.IO.IOException)
2. Calling an long blocking operation on 1 of the user side events.

I think letting bot devs know that this can happen on the ``Log``, ``Disconnected``, ``Connected``, and ``Ready`` events and for whatever means avoid any possible exceptions and long blocking operations it might eliminate it entirely as I recently did. However I am still using an ``DiscordSocketClientFactory`` that creates a new ``DiscordSocketClient`` every invoke of the ``Disconnected`` event and the code also disposes of the old one (and tries to connect with the new instance 5 seconds later that is multiplied by the number of reconnections and capped at 60 seconds) (similar to what the lib does somewhere) just to stay safe from ratelimits.

SubZero0:
@AraHaan This is unrelated to this PR so it shouldn't have been posted here but to answer you:

Throwing exceptions shouldn't cause an issue since it's try-catch'd (but imo you should deal with exception and not bubble them up in the event handler), you can see how it's done at [DiscordSocketClient.cs#L475](https://github.com/discord-net/Discord.Net/blob/dev/src/Discord.Net.WebSocket/DiscordSocketClient.cs#L475).

Blocking any event could kill your connection, this is already stated in our docs, see [https://docs.stillu.cc/guides/concepts/events.html#safety](https://docs.stillu.cc/guides/concepts/events.html#safety).

PR #1581, opened 2020-06-26
Implemented include_roles for guilds/id/prune get&post

RheaAyase:
[docs](https://discord.com/developers/docs/resources/guild#get-guild-prune-count)
* Get Guild Prune Count
* Begin Guild Prune

Added missing `include_roles`.

Feedback welcome! (For instance I wasn't sure what would be preferred way of adding a list to the GET method.)

### DIFF
--- a/src/Discord.Net.Core/Entities/Guilds/IGuild.cs
+++ b/src/Discord.Net.Core/Entities/Guilds/IGuild.cs
@@ -705,11 +705,12 @@ namespace Discord
         /// <param name="days">The number of days required for the users to be kicked.</param>
         /// <param name="simulate">Whether this prune action is a simulation.</param>
         /// <param name="options">The options to be used when sending the request.</param>
+        /// <param name="includeRoleIds">An array of role IDs to be included in the prune of users who do not have any additional roles.</param>
         /// <returns>
         ///     A task that represents the asynchronous prune operation. The task result contains the number of users to
         ///     be or has been removed from this guild.
         /// </returns>
-        Task<int> PruneUsersAsync(int days = 30, bool simulate = false, RequestOptions options = null);
+        Task<int> PruneUsersAsync(int days = 30, bool simulate = false, RequestOptions options = null, IEnumerable<ulong> includeRoleIds = null);
         /// <summary>
         ///     Gets a collection of users in this guild that the name or nickname starts with the
         ///     provided <see cref="string"/> at <paramref name="query"/>.
--- a/src/Discord.Net.Rest/API/Rest/GuildPruneParams.cs
+++ b/src/Discord.Net.Rest/API/Rest/GuildPruneParams.cs
@@ -9,9 +9,13 @@ namespace Discord.API.Rest
         [JsonProperty("days")]
         public int Days { get; }
 
-        public GuildPruneParams(int days)
+        [JsonProperty("include_roles")]
+        public ulong[] IncludeRoleIds { get; }
+
+        public GuildPruneParams(int days, ulong[] includeRoleIds)
         {
             Days = days;
+            IncludeRoleIds = includeRoleIds;
         }
     }
 }
--- a/src/Discord.Net.Rest/Entities/Guilds/RestGuild.cs
+++ b/src/Discord.Net.Rest/Entities/Guilds/RestGuild.cs
@@ -205,7 +205,7 @@ namespace Discord.Rest
                 role?.Update(model);
             }
         }
-        
+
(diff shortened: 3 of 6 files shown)


src/Discord.Net.Core/Entities/Guilds/IGuild.cs
```diff
@@ -704,12 +704,13 @@ public interface IGuild : IDeletable, ISnowflakeEntity
...
         /// <param name="days">The number of days required for the users to be kicked.</param>
         /// <param name="simulate">Whether this prune action is a simulation.</param>
+        /// <param name="includeRoleIds">An array of role IDs to include in the prune.</param>
         /// <param name="options">The options to be used when sending the request.</param>
         /// <returns>
         ///     A task that represents the asynchronous prune operation. The task result contains the number of users to
         ///     be or has been removed from this guild.
         /// </returns>
-        Task<int> PruneUsersAsync(int days = 30, bool simulate = false, RequestOptions options = null);
+        Task<int> PruneUsersAsync(int days = 30, bool simulate = false, ulong[] includeRoleIds = null, RequestOptions options = null);
```
SubZero0: Add it after `RequestOptions`, so it's accepted to be added without a major release and I would suggest changing it to `IEnumerable<ulong>`, so you could use Linq directly or arrays, would need to change some other places (and `.Any()` instead of Length).

src/Discord.Net.Rest/API/Rest/GuildPruneParams.cs
```diff
@@ -9,9 +9,13 @@ internal class GuildPruneParams
         [JsonProperty("days")]
         public int Days { get; }
 
-        public GuildPruneParams(int days)
+       [JsonProperty("include_roles")]
```
SubZero0: Missing one space.

src/Discord.Net.Rest/Entities/Guilds/RestGuild.cs
```diff
@@ -631,8 +631,8 @@ public Task<RestGuildUser> GetOwnerAsync(RequestOptions options = null)
         ///     A task that represents the asynchronous prune operation. The task result contains the number of users to
         ///     be or has been removed from this guild.
         /// </returns>
-        public Task<int> PruneUsersAsync(int days = 30, bool simulate = false, RequestOptions options = null)
-            => GuildHelper.PruneUsersAsync(this, Discord, days, simulate, options);
+        public Task<int> PruneUsersAsync(int days = 30, bool simulate = false, ulong[] includeRoleIds = null, RequestOptions options = null)
```
SubZero0: Same as `IGuild`.

src/Discord.Net.WebSocket/Entities/Guilds/SocketGuild.cs
```diff
@@ -746,8 +746,8 @@ public SocketGuildUser GetUser(ulong id)
             return null;
         }
         /// <inheritdoc />
-        public Task<int> PruneUsersAsync(int days = 30, bool simulate = false, RequestOptions options = null)
-            => GuildHelper.PruneUsersAsync(this, Discord, days, simulate, options);
+        public Task<int> PruneUsersAsync(int days = 30, bool simulate = false, ulong[] includeRoleIds = null, RequestOptions options = null)
```
SubZero0: Same as `IGuild`.

src/Discord.Net.Core/Entities/Guilds/IGuild.cs
```diff
@@ -704,12 +704,13 @@ public interface IGuild : IDeletable, ISnowflakeEntity
         /// </remarks>
         /// <param name="days">The number of days required for the users to be kicked.</param>
         /// <param name="simulate">Whether this prune action is a simulation.</param>
+        /// <param name="includeRoleIds">An array of role IDs to include in the prune.</param>
```
SubZero0: I believe this could get a better description to what it does (from discord docs: "Any inactive user that has a subset of the provided role(s) will be counted in the prune and users with additional roles will not.").

Maybe "An array of role ids to be disconsidered and prune the users if they don't have additional roles.", @Still34 might have a better suggestion.

SubZero0 (changes requested):

src/Discord.Net.Core/Entities/Guilds/IGuild.cs
RheaAyase: I kept it as an array internal with the public methods ienumerable...

PR #1604, opened 2020-08-07
feature: Add missing application properties (including Teams)

### LINKED ISSUE #1591: Add team info to the RestApplication entity
The Discord API returns team info in a request to `/oauth2/applications/@me`, yet because the entity in Discord.Net for `RestApplication` does not include team info, it never becomes available to the application using the library. Having this would be useful to implement features like multiple owners on a bot (ie. anyone on the team can be considered an owner).

SubZero0:
## Summary

This will add some missing properties related to bots to `IApplication`: `IsBotPublic`, `BotRequiresCodeGrant`, and `Team`.
Fixes #1591 

## Changes

- Add `CDN.GetTeamIconUrl`.
- Add `IApplication` (and `RestApplication`) properties:  `IsBotPublic`, `BotRequiresCodeGrant` and `Team`.
- Add team related api classes: `Team`, `TeamMember`, `API.MembershipState`
- Add team related interfaces: `ITeam`, `ITeamMember`
- Add team related classes: `RestTeam`, `RestTeamMember`, `Rest.MembershipState`

### DIFF
--- a/src/Discord.Net.Core/CDN.cs
+++ b/src/Discord.Net.Core/CDN.cs
@@ -8,6 +8,17 @@ namespace Discord
     public static class CDN
     {
         /// <summary>
+        ///     Returns a team icon URL.
+        /// </summary>
+        /// <param name="teamId">The team identifier.</param>
+        /// <param name="iconId">The icon identifier.</param>
+        /// <returns>
+        ///     A URL pointing to the team's icon.
+        /// </returns>
+        public static string GetTeamIconUrl(ulong teamId, string iconId)
+            => iconId != null ? $"{DiscordConfig.CDNUrl}team-icons/{teamId}/{iconId}.jpg" : null;
+
+        /// <summary>
         ///     Returns an application icon URL.
         /// </summary>
         /// <param name="appId">The application identifier.</param>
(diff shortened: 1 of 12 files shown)


k-boyle:
Should prob change the require owner precondition or add a new one that works with teams

SubZero0:
> Should prob change the require owner precondition or add a new one that works with teams

I thought about that but 1) changing the current one would change it's behaviour to a possible unwanted one without any way to people to know unless they read the changelogs carefully and 2) I believe Teams are still in the barebones considering there's "Permissions" and a list of roles that currently only has "Admin" so Discord will probably add more (atm might need a toggle if people consider all team members as owners or only the team owner), so it might add a lot more complexity.

Considering the above, I decided to refrain on changing or adding a new one, leaving to the consumer to create their own since it should be simple to fit it to their needs, so I'll leave it up to discussion here and see what's the consensus (if there's any).

Emzi0767:
Immutable list is very slow. In fact, up to 125x slower than plain list. Use an immutable array or a read only collection wrapper.

SubZero0:
@Emzi0767, thanks for the heads up. Read about it and changed it to an immutable array.

PR #1608, opened 2020-08-16
fix: Voice overwrites and CategoryId remarks

SubZero0:
## Summary

Missed the overwrites for voice channels when modifying them and changed the category id remark since zero will return a bad request.


hendrikdemir:
Ref https://dotnetfiddle.net/FJ1xzQ (0 vs zero)

NexganGH (approved):

PR #1624, opened 2020-09-06
Handle null PreferredLocale in rare cases where Discord messes up

hollowstrawberry:
Sometimes Discord messes up and leaves a guild with a null PreferredLocale, causing an error to be thrown. This fixes that from Discord.Net's end even though it's Discord's fault.

I made this this PR because I had been getting an error for a single guild every time my bot starts up.
```
Gateway6> Error handling Dispatch (GUILD_AVAILABLE): System.ArgumentNullException: Value cannot be null. (Parameter 'name')
   at System.Globalization.CultureInfo..ctor(String name, Boolean useUserOverride)
   at Discord.WebSocket.SocketGuild.Update(ClientState state, Guild model)
   at Discord.WebSocket.SocketGuild.Update(ClientState state, ExtendedGuild model)
   at Discord.WebSocket.DiscordSocketClient.ProcessMessageAsync(GatewayOpCode opCode, Nullable`1 seq, String type, Object payload)
```

### DIFF
--- a/src/Discord.Net.WebSocket/Entities/Guilds/SocketGuild.cs
+++ b/src/Discord.Net.WebSocket/Entities/Guilds/SocketGuild.cs
@@ -381,7 +381,7 @@ namespace Discord.WebSocket
             Description = model.Description;
             PremiumSubscriptionCount = model.PremiumSubscriptionCount.GetValueOrDefault();
             PreferredLocale = model.PreferredLocale;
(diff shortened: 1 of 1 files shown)


hollowstrawberry:
Actually I found the guild, I guess I'll make a report to the discord api

ID is 297408095137562625
image

foxbot (approved):

Issue #1636, opened 2020-09-26
Culture problem with webhook token regex validation


serkanyazicioglu:
Hello,

Webhook url validation regex fails when used with various different cultures. You can basically add 'RegexOptions.CultureInvariant' to WebhookUrlRegex object to mitigate the issue.

`private static Regex WebhookUrlRegex = new Regex(@"^.*(discord|discordapp)\.com\/api\/webhooks\/([\d]+)\/([a-z0-9_-]+)$", RegexOptions.CultureInvariant | RegexOptions.Compiled | RegexOptions.IgnoreCase);`

Thanks...

Starbors:
lol

SubZero0:
Closing since #1637 was merged.

PR #1637, opened 2020-09-27
issue1636: Added CultureInvariant RegexOption to WebhookUrlRegex

mathewsjwh:
As mentioned in issue #1636, webhook regex fails with other cultures; requesting confirmation testing on this branch from a third-party.


MarkusG:
Remember to leave a space before your # when referencing issue #1636 to make it link properly!

PR #1639, opened 2020-10-01
Fix: InviteMetadata and Invite definitions #1495

### LINKED ISSUE #1495: Invite and InviteMetadata missing fields and not corresponding to the discord object
There's a few missing fields and inconsistencies with how invites are currently implemented.

- Discord added two new fields to the [Invite object](https://discordapp.com/developers/docs/resources/invite#invite-object-invite-structure), `target_user?` and `target_user_type?` that are, at the moment, related only to Go Live invites. They are still missing in the library.
- The [`inviter?` field](https://discordapp.com/developers/docs/resources/invite#invite-object-invite-structure) should be part of [`Invite`](https://github.com/discord-net/Discord.Net/blob/dev/src/Discord.Net.Rest/API/Common/Invite.cs) and optional but it's defined in [`InviteMetadata`](https://github.com/discord-net/Discord.Net/blob/dev/src/Discord.Net.Rest/API/Common/InviteMetadata.cs).
- All the [fields in the invite metadata](https://discordapp.com/developers/docs/resources/invite#invite-metadata-object-invite-metadata-structure) aren't optional but [they are in the library](https://github.com/discord-net/Discord.Net/blob/dev/src/Discord.Net.Core/Entities/Invites/IInviteMetadata.cs).

As a note, a target user type enum was added in #1491, shared by REST and gateway, could be used to add the missing field (so it would need to be added to Core).

Fyers:
- added `target_user?`, `target_user_type?` to Invite
- removed `IsRevoked` / `revoked` Metadata-Field which does not exist - https://discord.com/developers/docs/resources/invite#invite-metadata-object-invite-metadata-structure
- removed nullable value type from InviteMetadata since its not nullable - https://discord.com/developers/docs/resources/invite#invite-metadata-object-invite-metadata-structure
- moved `inviter` from Metadata to the actual Invite object like in the docs.


kamranayub:
This is helpful, I'm generating invites for specific users and was looking for this 💯 Pulled it into my fork.

kamranayub:
@Fyers What do you think about including the needed params for `CreateInviteAsync`? Will need to add `target_user` and `target_user_type` there too. I can add it in my fork and test it out, I can reference the commit then for this PR.

See: https://discord.com/developers/docs/resources/channel#create-channel-invite

~edit: I _think_ `target_user` should be `snowflake` since docs say "user id" but the type it shows is `string` 😕 It would make sense to be an ID.~ Ah, user IDs are serialized as `string` in JSON.

edit 2: OK, nevermind, looks like this is for something else because it only works with `target_user_type` of 1 which is `STREAM` for streamers. It doesn't work if you try to pass a plain user ID without a type.

SubZero0:
If you want this merged in a minor version, you will need to not remove or change any public properties (and I don't see a major coming any time soon).

Fyers:
> If you want this merged in a minor version, you will need to not remove or change any public properties (and I don't see a major coming any time soon).

I could re-add the `IsRevoked` and `Revoked` property and mark it as obsolete.

SubZero0:
Obsolete would be fine plus reverting all return type changes for public properties.

PR #1666, opened 2020-11-03
Added negative TimeSpan handling

### LINKED ISSUE #1657: Current TimeSpanTypeReader doesn't parse negative TimeSpan
The current TimeSpanTypeReader doesn't parse negative TimeSpans like -1m30s, which is supported by players like lavaplayer/lavalink and Lavalink4NET.

kjhf:
* Added negative TimeSpan handling in the TimeSpanTypeReader.
* Added unit tests for the TimeSpanTypeReader in its own unit test file.
* Fixes https://github.com/discord-net/Discord.Net/issues/1657

### DIFF
--- a/src/Discord.Net.Commands/Readers/TimeSpanTypeReader.cs
+++ b/src/Discord.Net.Commands/Readers/TimeSpanTypeReader.cs
@@ -6,30 +6,50 @@ namespace Discord.Commands
 {
     internal class TimeSpanTypeReader : TypeReader
     {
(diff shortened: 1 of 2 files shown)


SubZero0:
While this change would be fine for me with a few tweaks, I would still consider it a breaking change since there might be people using it not expecting negative values being accepted so it would only be merged in a major version bump.

quinchs (approved):

PR #1667, opened 2020-11-04
fix: Fix audio dispose

### LINKED ISSUE #1274: Problem with sending voice through ffmpeg(spamming Sent Speaking)
Im sending audio using ffmpeg(`ffmpeg version N-93129-g9e1e521393`) and Discord.Net(`Discord.Net 2.0.2-dev-01059`).
I'm sending audio from local .mp3 using ffmpeg with methods like below.
When i'm sending first time from connection to voice channel - it is all ok.
For the next times bot spamming Sent Speaking and his "circle" in discord is blinking.
```cs
private Process CreateStream(string path)
{
	return Process.Start(new ProcessStartInfo
	{
		FileName = "ffmpeg",
		Arguments = $"-hide_banner -loglevel panic -i \"{path}\" -ac 2 -f s16le -ar 48000 pipe:1",
		UseShellExecute = false,
		RedirectStandardOutput = true,
	});
}
```
```cs
using (var ffmpeg = CreateStream(path))
using (var output = ffmpeg.StandardOutput.BaseStream)
using (var discord = AudioClient.CreatePCMStream(AudioApplication.Mixed))
{
	try
	{
		IsSending = true;
		await AudioClient.SetSpeakingAsync(true);//just try to fix
		await output.CopyToAsync(discord, _token);//_token to stop sending audio
		return true;
	}
	catch (OperationCanceledException)
	{
		return true;
	}
	finally
	{
		await discord.FlushAsync();
		await AudioClient.SetSpeakingAsync(false);//just try to fix
		IsSending = false;
	}
}
```

SubZero0:
## Summary

The audio streams weren't disposing the others so only one in the chain would be disposed, this would leave a stream open and trying "to shutdown" all audio.

If another stream was created, they would spam send speaking (one trying to enable and one trying to disable) causing a ratelimit and session invalidation.

It would be good to a few more people to test it to be sure I didn't miss anything.

Fixes #1274

## Changes

- Call Dispose on the `_next` stream

### DIFF
--- a/src/Discord.Net.WebSocket/Audio/Streams/BufferedWriteStream.cs
+++ b/src/Discord.Net.WebSocket/Audio/Streams/BufferedWriteStream.cs
@@ -61,14 +61,17 @@ namespace Discord.Audio.Streams
 
             _task = Run();
         }
+
         protected override void Dispose(bool disposing)
         {
             if (disposing)
             {
                 _disposeTokenSource?.Cancel();
                 _disposeTokenSource?.Dispose();
+                _cancelTokenSource?.Cancel();
                 _cancelTokenSource?.Dispose();
                 _queueLock?.Dispose();
+                _next.Dispose();
             }
             base.Dispose(disposing);
         }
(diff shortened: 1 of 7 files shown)


F0903:
Works absolutely phenomenal for me as of now. All of my previous problems seem to have gone now. 
Amazing work 😛

PR #1676, opened 2020-11-12
Added Method to delete Reaction without message object

Playwo:
- Added a method in the DiscordRestClient that allows you to delete reactions without needing the message object

- Useful for Pagination without needing the Paginator message in Cache

- This is probably not the best place to put it (If there even is a good one for it)


SubZero0:
Maybe add the others while at it? Plus adding the url encoding introduced by 04389a4f3e7120d70b6c6f7c21fcec3355b1b28e

SubZero0 (changes requested):
Just adding the requested changes so it isn't merged by accident. Same as said before: "Maybe add the others while at it? Plus adding the url encoding introduced by 04389a4f3e7120d70b6c6f7c21fcec3355b1b28e".

Playwo:
Well I had to pull the changes to have access to the UrlEncode method, only the last 2 commits matters.
Which other methods are meant?  I've added overloads for 
AddReaction, RemoveReaction, RemoveAllReactions, RemoveAllReactionsForEmote but theoretically there could be such overloads for every endpoint.

PR #1682, opened 2020-11-17
Fix: SocketGuild.HasAllMembers is false if a user left a guild

230Daniel:
SocketGuild.HasAllMembers is defined as `MemberCount == DownloadMemberCount`, but in some cases `DownloadMemberCount` is greater than `MemberCount`. For example, if a user left and the user wasn't removed from the cache.

If this occurs after this change, HasAllMembers will remain true because it is defined as `MemberCount >= DownloadMemberCount`.

This is useful for occasionally looping through guilds where HasAllMembers is false to download the users. We don't want to download users again if there are more users in the cache than in the server itself.


230Daniel:
Oops, I in my infinite wisdom got this the wrong way around. I'll open another with the correct boolean logic.

k-boyle:
> Oops, I in my infinite wisdom got this the wrong way around. I'll open another with the correct boolean logic.

Couldn't you have just pushed a new commit fixing it 🤔

PR #1690, opened 2020-11-21
Add max bitrate value to SocketGuild

vrachv:
Added a method in the SocketGuild which allows you to find out the maximum bitrate in the guild for voice channels.


SubZero0:
If we are adding this as a QoL (bitrate is a client-side change), it should be added to the interface and the rest implementation.
And afaik, the `<return>` should be more descriptive, like: `An <see langword="int"/> with the maximum bitrate for voice channels.`.

vrachv:
Why this merged? Still not done what @SubZero0 said

Still34:
Yeah I misread the comment - bit late now. Feel free to continue working on it on the same branch though! Can always merge again later.

PR #1700, opened 2020-11-27
Implementation of ValidateAndGetBestMatch

### LINKED ISSUE #1699: Rework `CommandService#ExecuteAsync` logic and allow easier extensibility.
In the current state, `ExecuteAsync` does two functions. Command Execution and `Command Validation (Valid command check/ predicate check/ parameter check/ override check, etc.)`.  
My idea is to split those into two separate functions, one responsible for validating the commands and another for execution.
For compatibility and no major breaks, i decided to keep the current public api unchanged with only the addition of a `ValidateAndGetBestMatch` function with the following signature.

```cs
public async Task<(IResult, Optional<CommandMatch>)> ValidateAndGetBestMatch(SearchResult matches,
ICommandContext context,
IServiceProvider provider,
MultiMatchHandling multiMatchHandling = MultiMatchHandling.Exception)
```

This function is to be called with `Search(String)`'s output making `ExecuteAsync`'s logic exclusive for execution,
`Search`'s exclusive for searching and `Validate...` exclusive for validation and command matching.

An *intended* side-effect is the extensibility this would allow if accepted. An external CommandService wrapper could leverage Discord.Net's `Validate...` in order to get the command that best fits **without executing** it, as the only current way of getting the command that **was** called is via an event handler, but there is no clean way for getting a command that *will* be called.

```cs
class Wrapper {
    //Task CheckForWhatever(string)
    var search = _commandService.Search(message.Content.Substring(argPos));
    var (status, command) = await ValidateAndGetBestMatch(search); //Other params ommited for sake of clarity
    if (!status.IsSucess) return;
    command.Value.ExecuteAsync(command.Command, context, status); 
}
```

Status
---
- [x] Implemented on [`siscodeorg/Discord.Net`](https://github.com/siscodeorg/Discord.Net/blob/commands/validate-get-best-match/src/Discord.Net.Commands/CommandService.cs)
- [x] Discussed by community on the discord channel. got mixed reactions.
- [x] PR opened - #1700

alikindsys:
As discussed on #1699 this pr comes with a rewrite of the logic on `CommandService#ExecuteAsync`, splitting it into two functions.    
  
`ExecuteAsync` - will continue to do the `Search-Verify-Match-Execute` loop as it did before.  
  
`ValidateAndGetBestMatch` - Verifies from a set of commands (`SearchResult`) with the specified `ICommandContext`, `IServiceProvider` and `MultiMatchHandling` which command is the optimal one, using the previously established classification and ranking systems. 

Breaking Changes
---
None.

Focus
---
Code cleanup and creation of an new public api.

### DIFF
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -511,7 +511,6 @@ namespace Discord.Commands
                 await _commandExecutedEvent.InvokeAsync(Optional.Create<CommandInfo>(), context, searchResult).ConfigureAwait(false);
                 return searchResult;
             }
-                
 
             var commands = searchResult.Commands;
             var preconditionResults = new Dictionary<CommandMatch, PreconditionResult>();
@@ -559,24 +558,6 @@ namespace Discord.Commands
                 parseResultsDict[pair.Key] = parseResult;
             }
 
-            // Calculates the 'score' of a command given a parse result
-            float CalculateScore(CommandMatch match, ParseResult parseResult)
-            {
-                float argValuesScore = 0, paramValuesScore = 0;
-
-                if (match.Command.Parameters.Count > 0)
-                {
-                    var argValuesSum = parseResult.ArgValues?.Sum(x => x.Values.OrderByDescending(y => y.Score).FirstOrDefault().Score) ?? 0;
-                    var paramValuesSum = parseResult.ParamValues?.Sum(x => x.Values.OrderByDescending(y => y.Score).FirstOrDefault().Score) ?? 0;
-
-                    argValuesScore = argValuesSum / match.Command.Parameters.Count;
-                    paramValuesScore = paramValuesSum / match.Command.Parameters.Count;
-                }
-
-                var totalArgsScore = (argValuesScore + paramValuesScore) / 2;
-                return match.Command.Priority + totalArgsScore * 0.99f;
-            }
-
             //Order the parse results by their score so that we choose the most likely result to execute
             var parseResults = parseResultsDict
                 .OrderByDescending(x => CalculateScore(x.Key, x.Value));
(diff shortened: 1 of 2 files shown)


quinchs (approved):

quinchs:
Going to cc @Cenngo on this just for sanity

src/Discord.Net.Commands/CommandService.cs
```diff
@@ -506,19 +506,66 @@ public async Task<IResult> ExecuteAsync(ICommandContext context, string input, I
...
+        /// <param name="matches">The SearchResult.</param>
+        /// <param name="context">The context of the command.</param>
+        /// <param name="provider">The service provider to be used on the command's dependency injection.</param>
+        /// <param name="multiMatchHandling">The handling mode when multiple command matches are found.</param>
+        /// <returns>A task that represents the asynchronous validation operation. The task result contains the result of the
+        ///     command validation and the command matched, if a match was found.</returns>
+        public async Task<(IResult, Optional<CommandMatch>)> ValidateAndGetBestMatch(SearchResult matches, ICommandContext context, IServiceProvider provider, MultiMatchHandling multiMatchHandling = MultiMatchHandling.Exception)
+        {
+            if (!matches.IsSuccess)
+                return (matches, Optional.Create<CommandMatch>());
```
Cenngo: I dont think returning an optional CommandMatch to indicate to the user that their SearchResult wasnt successful isnt the most user friendly way of validating a command search. Instead of returning a Tuple, i think you should encapsulate this in an `IResult` that way you can provide more information to the user and it would be easier to build a command execution pipeline around this, since your goal is to allow people to utilize this.

src/Discord.Net.Commands/CommandService.cs
```diff
@@ -506,19 +506,66 @@ public async Task<IResult> ExecuteAsync(ICommandContext context, string input, I
             services = services ?? EmptyServiceProvider.Instance;
 
             var searchResult = Search(input);
-            if (!searchResult.IsSuccess)
+
+            //Since ValidateAndGetBestMatch is deterministic on the return type, we can use pattern matching on the type for infering the code flow.
+            var (validationResult, commandMatch) = await ValidateAndGetBestMatch(searchResult, context, services, multiMatchHandling);
+
+            if(validationResult is SearchResult)
```
Cenngo: Tho i understand the way you are using this workflow to branch out the execution method, i dont think it is the most descript way of doing this. I think this should be more user readable to make it easier to maintain. Using an `IResult` to return the method result, with some additional info for branching out the pipeline, might be a better idea.

Cenngo:
Other than that, it all seems good.

src/Discord.Net.Commands/CommandService.cs
alikindsys: I see.

src/Discord.Net.Commands/CommandService.cs
alikindsys: Understood.

quinchs:
@roridev Are you able to make the requested changes?

alikindsys:
Im able now, just give me a moment

alikindsys:
Please check if the changes done align with what was requested.

Fun things that were discovered while doing this:

- A whole branch of execution was missing on my old code. The event wouldn't be fired if `ValidateAndGetBestMatch` returned `SearchResult`. This is now fixed.

quinchs (approved):

quinchs:
@Cenngo waiting on your final approval

src/Discord.Net.Commands/CommandService.cs
```diff
@@ -506,19 +506,83 @@ public async Task<IResult> ExecuteAsync(ICommandContext context, string input, I
...
+
+            if (validationResult is MatchResult matchResult)
+            {
+                return await handleCommandPipeline(matchResult, context, services);
+            }
+
+            return validationResult;
+        }
+
+        private async Task<IResult> handleCommandPipeline(MatchResult matchResult, ICommandContext context, IServiceProvider services)
```
Cenngo: Could you please make this pascal casing?

Cenngo (changes requested):
The changes you made address the issues i mentioned. Great to hear that this also led you to fixing a problem.

src/Discord.Net.Commands/CommandService.cs
alikindsys: oh sure rider autocorrected it to that.

Cenngo (approved):

MrCakeSlayer (approved):

PR #1713, opened 2020-12-13
Simplify code in Discord.DiscordComparers

Rekkonnect:
This change simplifies the code by:
- Removing the need for private fields to return the privately stored comparer instances
- Removing the variable declaration in the switch statement for already existing variables

### DIFF
--- a/src/Discord.Net.Core/Utils/Comparers.cs
+++ b/src/Discord.Net.Core/Utils/Comparers.cs
@@ -11,29 +11,23 @@ namespace Discord
         /// <summary>
         ///     Gets an <see cref="IEqualityComparer{T}"/> to be used to compare users.
         /// </summary>
-        public static IEqualityComparer<IUser> UserComparer => _userComparer ??= new EntityEqualityComparer<IUser, ulong>();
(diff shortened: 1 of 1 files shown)


src/Discord.Net.Core/Utils/Comparers.cs
```diff
@@ -46,7 +40,7 @@ public override bool Equals(TEntity x, TEntity y)
                     (null, null) => true,
                     (null, _)    => false,
                     (_, null)    => false,
```
Joe4evr, line 42: Have we updated to `<LangVersion>9.0</LangVersion>` yet? If yes, these two cases could be condensed with the `or` pattern operator. (Just thinking out loud.)

Joe4evr (approved):
Not a big fan of eagerly initializing all the comparers on first access, but I guess it's not **_terrible_**.

Rekkonnect:
> Not a big fan of eagerly initializing all the comparers on first access, but I guess it's not **_terrible_**.

Well, initializing them is not too expensive, it barely adds any overhead. That's why I thought it would be fine with these changes in this case, I'm glad you're good with it as well.

quinchs (approved):

PR #1717, opened 2020-12-16
Implementing new Discord features: Interaction & Slash commands

quinchs:
Hello, I've started writing support for the new discord interaction features.

I think we should talk about Slash commands here, There are some questions I think we all have:
- How will Slash commands be written?
- Will they take on the same style as writing normal commands?
- Will Discord.Net.Commands handle invoking and managing Slash Commands?

I have wrote in some methods for the RestClient to get, create, edit, and delete global and guild slash commands, I've also created some interfaces to use for Interaction with discord.

PR #1733, opened 2020-12-27
Slash command service

quinchs:
# Summary

This PR is for discussion and implementation of a command service for #1717. 

Discussion, questions, and feedback is welcome as this PR is based off of how *you* want slash commands to be written.

This PR also includes changes made in #1717 

# Issues to overcome

You should not need to rewrite all your current commands to implement slash commands. Basically, the slash command service should be able to use the attributes from the base command service, like `Command`, `Summary`, `Name`, etc.

A proposed idea from @Mijago is as follows:
>  Couldn't we parse the currently existing parameter attributes and use them for the SlashCommand?
> ```cs
> [Command("item")]
> [Summary("Search for an item in the manifest.")]
> public async Task ShowItems(
>     [Remainder]
>     [Name("Name")]
>     [Summary("Specify the name of the item here. You can use RegExp to give a more specific search term.")]
>     string name)
> {
>     await ShowItemsLang(name, "en");
> }
> ```
> This way, we would have all the information your example above would also have. The types are already defined, so I doubt there is a need to re-define them.
> - Name and description of the command
> - Name and description of the argument
> - Type of the argument
> - Whether the argument is required or not

Another form of attribute based commands was suggested by @Cenngo:

> I think creating choices can also be accomplished by attributes since there are only 2 fields that need to be filled. I don't think there will be a lot of people who will add more than a handful of choices per option so i think this should'nt be that messy, relatively speaking.
> ```cs
> [SlashCommand("rainbow6", "description")]
>     public class Rainbow6
>     {
>         [SubCommandGroup("player", "description")]
>         [SubCommand("casual", "description")]
>         public async Task GetCasual ( [Description("")] [Required]string username, 
>             [Description("")] [Choices("UPlay", "uplay")][Choices("XBox Live", "xbl")][Choices("PlayStation Network", "psn")]string platform)
>         {
> 
>         }
> 
>         [SubCommandGroup("player", "description")]
>         [SubCommand("ranked", "description")]
>         public void GetRanked ( [Description("")] [Required]string username, 
>             [Choices("UPlay", "uplay")][Choices("XBox Live", "xbl")][Choices("PlayStation Network", "psn")] string platform )
>         {
> 
>         }
> 
>         [SubCommandGroup("operator", "description")]
>         [SubCommand("stats", "description")]
>         public void GetStats ( [Description("")] [Required]string username,
>             [Choices("Seasonal", "seasonal")][Choices("Overall", "overall")]string scope, 
>             [Choices("Close", 5)][Choices("Long", 10)]int saesons)
>         {
> 
>         }
>     }
> ```

# Changes
The first comit of this PR adds the `SlashCommandBuilder` to the command service, as suggested by @ZargorNet.
The `SlashCommandBuilder` is used to build `SlashCommandCreationProperties` in a much nicer way.

# Questions
- Should a slash command service/handler be a new project? For example: `Discord.Net.SlashCommands`
- Should a slash command service implemented into `CommandService` or should it be its own service?


docs/guides/commands/application-commands.md
```diff
@@ -0,0 +1,216 @@
...
+      Console.WriteLine("Unsupported interaction type: " + arg.Type);
+      break;
+  }
+}
+
+private async Task MySlashCommandHandler(SocketInteraction arg)
+{
+  switch(arg.Name)
+  {
+    case "hug":
```
ZargorNET, line 203: I really don't like this idea. Why not build a command handler that takes the name and parses it correctly to the desired method like the `InstallCommands` method? And I would stick with the current command design pattern: One class for one command(group) so that we can make use of the IServiceProvider system

docs/guides/commands/application-commands.md
```diff
@@ -0,0 +1,216 @@
...
+      // Get the user argument
+      var option = arg.Data.Options.First(x => x.Name == "user");
```
ZargorNET, line 205: This is also a point where I disagree. Why don't parse it out from the method parameters?

docs/guides/commands/application-commands.md
```diff
@@ -0,0 +1,216 @@
...
+    case InteractionType.ApplicationCommand: // If it is a command
+      await MySlashCommandHandler(arg); // Handle the command somewhere
+      break;
+     default: // We dont support it
+      Console.WriteLine("Unsupported interaction type: " + arg.Type);
+      break;
+  }
+}
+
+private async Task MySlashCommandHandler(SocketInteraction arg)
```
ZargorNET, line 199: Remove this parameter and make it a variable in like `SlashModuleBase<BaseSlashCommandContext>` (don't omit the generic because it's a really nice addition to modify the context to fit your needs (as I do in my project))

ZargorNET (commented):
> Should a slash command service/handler be a new project? For example: Discord.Net.SlashCommands

I think yes. It'd go well with the current design pattern and with this way you could omit unused files to speed up compilation time and reduce binary file size. And on top of that, you can easily replace for example the current CommandService to your own (because to be honest, it's not that good in the current state)

docs/guides/commands/application-commands.md
quinchs: The md was for #1717, its only here because this PR is based off of #1717

TheHeartOfFire:
Really looking forward to this feature. How can i help?

quinchs:
If you want to help write the nitty gritty reflection with me, just dm me on discord: quin#3017

src/Discord.Net.Commands/Builders/SlashCommandBuilder.cs
```diff
@@ -0,0 +1,491 @@
...
+                throw new ArgumentException(nameof(Options), $"Cannot have options on {Type} type");
+
+            return new ApplicationCommandOptionProperties()
+            {
+                Name = this.Name,
+                Description = this.Description,
+                Default = this.Default,
+                Required = this.Required,
+                Type = this.Type,
+                Options = new List<ApplicationCommandOptionProperties>(this.Options.Select(x => x.Build())),
```
Cenngo: I think you should add a simple null check on this line. In my testing i found out "Linq" throws an error while trying to add a sub-command's parameters, which cannot have any "application command options".

Something like this should do the trick:
```csharp
Options = Options != null ? new List<ApplicationCommandOptionProperties>(this.Options.Select(x => x.Build())) : null,
```

quinchs: Null check has been added [Here](https://github.com/discord-net/Discord.Net/blob/f4a8b219ed6d9de643154a9ea7f6b34e8591f984/src/Discord.Net.Commands/Builders/SlashCommandBuilder.cs#L393)

Mijago:
Hi, anything i could do to help you with this PR?
I dont have much spare time but as I would love to see this feature, I'd like to help, even if it would only be concept work.

quinchs:
Providing ideas on how you would want to write slash command and how a slash command service should work would be ideal if you don't have a lot of time. I have also run out of free time because of work and school so I can only collect ideas as of now. 

If you want to just write code you can make a PR on my fork and I'll check it out but be warned writing a command service requires the bravest of nerds due to the godsent amount of reflection.

ginomessmer:
Hi @quinchs. Thanks for your work so far. 

> If you want to just write code you can make a PR on my fork and I'll check it out but be warned writing a command service requires the bravest of nerds due to the godsent amount of reflection.

I'm new to this repository but I would love to take a look. Could you please briefly describe the current state of your work and elaborate on the pending work? That would help others to take over your progress if necessary.

quinchs:
> Could you please briefly describe the current state of your work and elaborate on the pending work?

The support for slash commands (like creating/receiving them) is done in #1717 while a SlashCommandService still needs to be worked on. 

Me and @TheHeartOfFire have worked a tad on the service and we decided we want to support class based slash commands and attribute based, what I mean by this is that the service will allow command as seen below

`Attribute based`
```cs
[Command("item")]
[Summary("Search for an item in the manifest.")]
public async Task ShowItems(
    [Remainder]
    [Name("Name")]
    [Summary("Specify the name of the item here. You can use RegExp to give a more specific search term.")]
    string name)
{
    await ShowItemsLang(name, "en");
}
```

`Class based`
```cs
//                                                 allow DiscordSocketClient or DiscordShardedClient
public class MySlashCommand : SlashCommandBase<DiscordSocketClient>
{
    private DiscordSocketClient client;
    public MySlashCommand(DiscordSocketClient client)
        : base()
    {
        // Init
        this.client = client;

        // Maybe listen for this command with an event?
        this.Executed += HandleExecute;

        // Register the command here with all the command args?
        // Would use the SlashCommandBuilder here? https://github.com/quinchs/Discord.Net/blob/SlashCommandService/Discord.Net.SlashCommands/Builders/SlashCommandBuilder.cs
        base.Create("some arguments");
    }

    // Instead of passing the raw interaction, maybe use a context like https://github.com/quinchs/Discord.Net/blob/SlashCommandService/src/Discord.Net.Commands/SlashCommands/Types/SlashCommandContext.cs
    public async Task HandleExecute(SocketInteraction arg)
    {
        await arg.RespondAsync("Success!");
    }
}
```

Although is could all be changed I feel like most people would use one of these methods, and reflection could boil methods down to a single class like this https://github.com/quinchs/Discord.Net/blob/SlashCommandService/src/Discord.Net.Commands/SlashCommands/Types/SlashCommandModule.cs then just execute based on that. None of the reflection is done yet as I didn't get time to work on it.

TL;DR: reflection needs to be written, basically everything needs to be written.

quinchs:
Merged @SlenderPlays PR on my repo because it looks like a good start to work off of.

MoonieGZ:
Heads up, been trying this around a little bit and it hits an error in
Discord.Net.WebSocket\Entities\Interaction\SocketApplicationCommand.cs line 59

model.Options can be null and adding a null check seems to fix it.

quinchs:
Il take a look, keep in mind this pr is not done and may contain lots of bugs

MoonieGZ:
I am aware, I'm very happy with the experiments I'm able to run with it so far, thank you so much for your work!

quinchs:
Thank you for using it!

erxkk:
Quick heads up: after somebody had problems with the pr throwing a null ref internally:
[SlashCommands/SlashCommandServiceHelper.cs#L42](https://github.com/quinchs/Discord.Net/blob/SlashCommandService/src/Discord.Net.Commands/SlashCommands/SlashCommandServiceHelper.cs#L42) does not account for Interfaces, which [have no BaseType](https://docs.microsoft.com/en-us/dotnet/api/system.type.basetype?view=net-5.0#property-value) and thus throws a null ref if an interface is declared inside the assembly.
I did not try out the pr myself, but this is my conclusion after they sent some debugger info.

webmilio:
> 
> 
> Quick heads up: after somebody had problems with the pr throwing a null ref internally:
> [SlashCommands/SlashCommandServiceHelper.cs#L42](https://github.com/quinchs/Discord.Net/blob/SlashCommandService/src/Discord.Net.Commands/SlashCommands/SlashCommandServiceHelper.cs#L42) does not account for Interfaces, which [have no BaseType](https://docs.microsoft.com/en-us/dotnet/api/system.type.basetype?view=net-5.0#property-value) and thus throws a null ref if an interface is declared inside the assembly.
> I did not try out the pr myself, but this is my conclusion after they sent some debugger info.

Should be fixed (on another branch). Instead of checking for `null` we check for `!typeInfo.IsAbstract && !typeInfo.IsInterface` .

eva-cosma:
Just as an FYI, I've written my implementation of the slash command service with the goal to work, with little regard to programming norms. My biggest "sin" in all of this has to be [The Command Service Helper](https://github.com/quinchs/Discord.Net/blob/3e9b027f55fa974baadd3d3ae478fb8b87c8f73e/src/Discord.Net.Commands/SlashCommands/SlashCommandServiceHelper.cs). It's just an excuse to put all of the ugly code in. Personally, if I were to do it again I would split it in two - one for transforming the code that we get from the assembly into [INFO's](https://github.com/quinchs/Discord.Net/tree/SlashCommandService/src/Discord.Net.Commands/SlashCommands/Info) and another to handle them properly - in this section, I would include just the registration bit, but I can totally see preconditions and other stuff like that be introduced here as well.

Also, it really needs to be cleaned up. [This whole block](https://github.com/quinchs/Discord.Net/blob/3e9b027f55fa974baadd3d3ae478fb8b87c8f73e/src/Discord.Net.Commands/SlashCommands/SlashCommandServiceHelper.cs#L255) could be totally put into a function that takes as a parameter the type of attribute we want to test for. Also, a lot of stuff is public when in reality we don't really want it to be.

To finish off, there are a lot of bugs, and I'm not super proud of this code, but I will let you folk with way more experience in building an API handle all of my war crimes against programming itself 😃 .

MoonieGZ:
Potentially dumb question, I notice in the sample there is a comment section about registering commands manually for example only certain commands in certain guilds...

Is this possible with this PR? If so, any pointers on implementing it? I'm prepping my servers for when this becomes "official".

Ideally (just throwing out ideas right now, don't burn me please), it'd be really nice to have some tags/attributes to restrict it to certain guild IDs

quinchs:
Currently idk. Will it be added? Yes.

eva-cosma:
> Potentially dumb question, I notice in the sample there is a comment section about registering commands manually for example only certain commands in certain guilds...
> 
> Is this possible with this PR? If so, any pointers on implementing it? I'm prepping my servers for when this becomes "official".

Technically you **can** do that right now but I didn't bother to test it. The purpose of the [BuildCommands method](https://github.com/quinchs/Discord.Net/blob/3e9b027f55fa974baadd3d3ae478fb8b87c8f73e/src/Discord.Net.Commands/SlashCommands/SlashCommandService.cs#L138) is to build the commands and give you the option to manually register them.
Here's a bit of pseudo-C#.
```c#
// Build the commands
List<SlashCommandCreationProperties> commands = await _commands.BuildCommands();

ulong guild1 = ... ;
ulong guild2 = ... ;

// Register half of the commands to guild1
for(int i = 0; i < commands.Count/2; i++)
{
    await socketClient.Rest.CreateGuildCommand(commands[i], guild1);
}

// and the other half to guild2
for(int i = commands.Count/2; i < commands.Count; i++)
{
    await socketClient.Rest.CreateGuildCommand(commands[i], guild2);
}
```

As I've said before, I didn't test this and there is a non-zero chance that there is a bug out there somewhere. I'm a bit swamped right now so I can't test it myself.

I also recommend taking a look at [this class](https://github.com/quinchs/Discord.Net/blob/SlashCommandService/src/Discord.Net.Core/Entities/Interactions/SlashCommandCreationProperties.cs).

> Ideally (just throwing out ideas right now, don't burn me please), it'd be really nice to have some tags/attributes to restrict it to certain guild IDs

Don't worry about being burned. I did want to implement a tag such as
```c#
[Guild(0301...)]
[Guild(0204...)]
public async Task PingAsync()
{
...
}
```

But the issue is that all of the uids have to be constant, so you can't set the guild ids from a config file or from anywhere for that matter. There might be a way to do this programmatically, but I'm not sure.  I guess for people that make their bot for a particular discord server this might work, but otherwise I don't think this could scale. 

I guess if there is a demand for it then it might be useful to implement.

MoonieGZ:
Appreciate the answer @SlenderPlays my current thought process is to have a "normal" bot command to register either individual commands or all of them, that way you're getting the guild ID from the context in which the command is run (therefore not in code)... However, I can see this causing problems when the bot is brought offline for example.

Then again this is all theoretical because the person whose server I'm trying to implement these on has never gotten back to me to give me the permissions to even create slash commands so doesn't really matter if I don't find a solution right away...

src/Discord.Net.Commands/SlashCommands/Attributes/Choice.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        /// <summary>
+        ///     The internal value of this choice.
+        /// </summary>
+        public int? choiceIntValue = null;
+
+        /// <summary>
+        ///     The display value of this choice.
+        /// </summary>
+        public string choiceName;
```
ginomessmer, line 28: Heads up: Public members should be [pascal case by convention](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions). Also, not sure if the internal values (`choiceStringValue`, `choiceIntValue`) should be public.

```suggestion
        /// <summary>
        ///     The internal value of this choice.
        /// </summary>
        private string _choiceStringValue;

        /// <summary>
        ///     The internal value of this choice.
        /// </summary>
        private int? _choiceIntValue = null;

        /// <summary>
        ///     The display value of this choice.
        /// </summary>
        public string ChoiceName { get; set; }
```

src/Discord.Net.Commands/SlashCommands/Attributes/CommandGroup.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        /// <summary>
+        ///     The name of this slash command.
+        /// </summary>
+        public string groupName;
+
+        /// <summary>
+        ///     The description of this slash command.
+        /// </summary>
+        public string description;
```
ginomessmer, line 23: Same principle here:

```suggestion
        /// <summary>
        ///     The name of this slash command.
        /// </summary>
        public string GroupName { get; set; }

        /// <summary>
        ///     The description of this slash command.
        /// </summary>
        public string Description { get; set; }
```

src/Discord.Net.Commands/SlashCommands/CommandRegistration/ExistingCommandOptions.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Discord.SlashCommands
+{
+    public enum ExistingCommandOptions
+    {
+        OVERWRITE,
+        KEEP_EXISTING
+    }
```
ginomessmer, line 13: [By convention](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/enum), enums should be written in pascal case.

```suggestion
    public enum ExistingCommandOptions
    {
        Overwrite,
        KeepExisting
    }
```

src/Discord.Net.Commands/SlashCommands/Info/SlashParameterInfo.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public object Parse(SocketInteractionDataOption dataOption)
+        {
+            switch (Type)
+            {
+                case ApplicationCommandOptionType.Boolean:
+                    if (Nullable)
+                        return (bool?)dataOption;
+                    else
+                        return (bool)dataOption;
```
ginomessmer, line 23: Can be simplified:

```suggestion
                    return Nullable ? (bool?) dataOption : (bool) dataOption;
```

src/Discord.Net.Commands/SlashCommands/Types/CommandModule/ISlashCommandModule.cs
```diff
@@ -0,0 +1,21 @@
...
+
+    public interface ISlashCommandModule
+    {
+        void SetContext(IDiscordInteraction interaction);
+
+        //void BeforeExecute(CommandInfo command);
+
+        //void AfterExecute(CommandInfo command);
+
+        //void OnModuleBuilding(CommandService commandService, ModuleBuilder builder);
```
ginomessmer, line 19: What is this used for? Consider removing it to minimize confusion.

src/Discord.Net.Rest/API/Common/InteractionFollowupMessage.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Threading.Tasks;
+
+namespace Discord.API
+{
+    internal class InteractionFollowupMessage
+    {
+        public string Content { get; set; }
+        public Optional<string> Username { get; set; }
+        public Optional<string> AvatarUrl { get; set; }
+        public Optional<bool> TTS { get; set; }
```
ginomessmer, line 15:
```suggestion
        public Optional<bool> TextToSpeechEnabled { get; set; }
```

ginomessmer (changes requested):

ginomessmer:
Honestly, this is my first code review. I've taken a look at this PR and the work is really great so far. 

> Just as an FYI, I've written my implementation of the slash command service with the goal to work, with little regard to programming norms. 

@SlenderPlays I'd be happy to help out and refactor some bits here and there. How can I help?

SlashCommandsExample/Program.cs
```diff
@@ -0,0 +1,26 @@
...
+using Discord;
+using Discord.Commands;
+using Discord.SlashCommands;
+using Discord.WebSocket;
+
+namespace SlashCommandsExample
+{
+	class Program
+	{
+        static void Main(string[] args)
```
aevitas, line 17: Nit: You can declare `Main` as `Task Main(string[] args)` and simply `await` your `RunAsync()` call to avoid mixing synchronous and asynchronous code.

Still34: It doesn't *really* matter here, as they both ultimately compile down to the same IL internally; at least, as far as this example is concerned.

Cenngo:
After 3 days worth of speed programming i finally got a working prototype that incorporates the REST, Web Socket, Message Component and Command Service functionalities. I'm currently in the process of creating the documentation but the code needs a reasonable amount of polishing. Since I don't have much free time and i had a different vision in mind i created my fork from ground up. Command registration is almost a carbon copy of the original Commands extension. Message component interactions are handled with an [Interaction] attribute.
```cs
[SlashGroup("utility", "utility commands")]
    public class Module : CommandBase<SocketSlashCommandContext>
    {
        [SlashCommand("echo", "repeat after me")]
        public async Task Echo (string text = "empty")
        {
            await Context.Interaction.PopulateAcknowledgement($"You just said: {text}");
        }

        [SlashCommand("animals", "animal commands")]
        public async Task FavouriteAnimal([Choice("Dog", "dog"), Choice("Cat", "cat")]string text)
        {
            await Context.Interaction.PopulateAcknowledgement($"your favourite animal is {text}");
        }

        [SlashGroup("user", "user management tools")]
        [SlashCommand("tag", "tag a user")]
        public async Task TagUser(IChannel channel, IUser user)
        {
            var socketUser = user as SocketUser;
            await (channel as SocketTextChannel)?.SendMessageAsync($"{socketUser?.Mention}: anan");
        }
    }

    public class SecondModule : CommandBase<SocketSlashCommandContext>
    {
        [SlashCommand("ping", "recieve a pong")]
        public async Task Ping ( )
        {
            var builder = new MessageUIBuilder();
            var select = new SelectMenuBuilder("select_song")
                .AddOption(new SelectOption("Never gonna give you up", "rickroll"))
                .AddOption(new SelectOption("I want it that way", "b99"))
                .WithPlaceholder("Select a song")
                .WithMaxValue(2)
                .Build();

            builder.AddRow(select);
            await Context.Interaction.PopulateAcknowledgement("pong", messageComponents: builder.Build());
        }

        [Interaction("select_song")]
        public async Task Song ( params string[] values )
        {
            await Context.Interaction.SendFollowupAsync($"you selected {string.Join(',', values)}");
        }
```
I'll keep working on this but if anyone wants to try it out you check out the fork [here](https://github.com/Cenngo/Discord.Net)

quinchs:
@Cenngo If you could make a PR over on [DNet Labs](https://github.com/Discord-Net-Labs/Discord.Net-Labs) that would be great!

Cenngo:
> @Cenngo If you could make a PR over on [DNet Labs](https://github.com/Discord-Net-Labs/Discord.Net-Labs) that would be great!

Sure, i will. Though since the last commit on this PR is from 6 months ago, i may have misjudged the state of development. I didn't know that repo existed.

quinchs:
Closing as this is implemented in #1923

JKamsker:
Is there a replacement for this? @quinchs 
Manually switching and seperately building commands isn't nice at all :(

Cenngo:
> Is there a replacement for this? @quinchs Manually switching and seperately building commands isn't nice at all :(

Head over to [D.Net-Labs](https://github.com/Discord-Net-Labs/Discord.Net-Labs) for that.

PR #1761, opened 2021-01-31
Fixed a deadlock in DiscordShardedClient during a failed Identify

asmejkal:
Fixed a deadlock in DiscordShardedClient during a failed Identify due to receiving an InvalidSession opcode. Happens when two DiscordShardedClients attempt to identify at once. 

The handling logic of InvalidSession seemed to only consider the case of receiving it in response to Resume, in which the AcquireIdentifyLockAsync is free to grab. But during Identify, the AcquireIdentifyLockAsync is already taken and won't be released until the client is Ready. So I added a special case in the InvalidSessions handler for when no session has been established yet (= it was received in response to Identify).

Please review thoroughly, the locking logic wasn't easy to navigate for a first-timer.

### DIFF
--- a/src/Discord.Net.WebSocket/DiscordSocketClient.cs
+++ b/src/Discord.Net.WebSocket/DiscordSocketClient.cs
@@ -554,25 +554,36 @@ namespace Discord.WebSocket
                     case GatewayOpCode.InvalidSession:
                         {
                             await _gatewayLogger.DebugAsync("Received InvalidSession").ConfigureAwait(false);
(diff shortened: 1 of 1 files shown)


asmejkal:
Or perhaps a better solution would be to not hog the Identify semaphore while waiting for the connection to be Ready?

SubZero0:
Thanks for reporting this. The connection timeout should not let it deadlock endlessly, but it shouldn't happen at all.

PR #1766, opened 2021-02-02
feature: Bring your own MessageCache

230Daniel:
This change will allow library users to create their own implementation of the MessageCache through the DiscordSocketConfig.

My use case is to disable the caching of new messages, and instead only cache messages which my bot is likely to request again in the future. Instead of implementing this functionality directly, it was recommended that I allow the library users to bring their own MessageCache which I agree makes much more sense.

I've not worked much with interfaces and abstract classes before, so please inform me of any changes I could make to keep this change consistent with other parts of the library and working properly.


quinchs:
A feature like this will be coming soonish for caching: Cache providers. That makes this PR out of date with that.

Issue #1764, opened 2021-02-01
Thoughts on adding a method to un-cache guild users?


230Daniel:
When I re-wrote my bot I tried to keep its memory usage to an absolute minimum. The biggest memory hog is certainly the user cache in Discord.NET, so I optimised this by only caching users for the guilds that needed their users cached. For example, those with my role persist feature enabled. I noticed that as my bot stayed online for longer and longer, the memory usage crept up fairly significantly.

I did a test with 780 servers (514431 total users), and the original cached user count of 76861 was up to 89073 after 20 hours of running. The 12212 extra users meant that my bot was using up 130 MB more memory after 20 hours than when it started up. It's my understanding that users are cached once they send a message or do something else in the guild.

I would love to be able to un-cache these users to free up some memory because I can imagine this becoming a bigger problem as my bot grows. A simple way of doing this would be allowing the library users to remove items from `SocketGuild.Users`, but I understand that this might be confusing for some developers who don't completely understand the caching system.

Does anybody have a good idea about how this could be implemented? Or should it not be implemented at all? I'd love to work on the implementation myself, although I'm not very familiar with how the library works.

k-boyle:
There's an internal method on SocketGuld to purge the users (which I don't think gets exposed elsewhere like the purge channel does, not that that method should ever be called) which you could use reflection to call, whether the lib survives after that's called is a different issue, but worth trying I guess. iirc Paulo said he tried to get the lib into a state where it can work cacheless but was a mess

230Daniel:
Ah cool, thanks for the pointer. That method has 2 references, one is called when the guild is removed and the other is called from public method DiscordSocketClient.PurgeUserCache() which is just a foreach loop calling the PurgeGuildUsers method.

It appears that the PurgeGuildUserCache also removes each user from some kind of global cache, so I would assume that the method was designed to be called for all guilds at once, and that if it's only called for a few guilds then problems will occur.

However, if it's safe to call the method for the one guild when it's removed, surely it must be safe to call it at any time? I've added a way to call the method and have added a script to my bot to clear all unnecessary users every 15 seconds. I'll see if I spot any issues, but everything has been working great so far.

PR #1767, opened 2021-02-02
feature: Add method to clear guild user cache

230Daniel:
This pr adds the public `SocketGuild.ClearUserCache(optional predicate)` method, allowing library users to remove all users or select users from a guild's user cache.

My reason for this change is memory usage optimisation. As gateway events come through the user cache gets filled and never resets. This leads to a gradual increase in memory usage which is more significant depending on the popularity of your bot. Adding this method lets us prevent this issue.

The optional predicate argument allows us to further customise which users are cleared from the cache. For example, I need the VoiceStateUpdated handler to give me both a before and after voice channels, which requires the users to be cached. So when clearing the cache for each guild, I only select the users which aren't in a voice channel.

I'm not very familiar with the library so it's possible that using this method could break something, but I've been testing it for a couple of days and I've not been able to find any issues so far.

There are no breaking changes.

issue: #1764


230Daniel:
I noticed that in the unmodified library when a new SocketGlobalUser is created to be added to a guild cache, there are two calls to `AddRef()`, yet when the user is removed from the cache (for example when leaving the server), `RemoveRef()` is only called once. This means that the SocketGlobalUser stays around despite no longer being used.

Could anyone please inform on why `AddRef()` is called twice? Attached is a video of me debugging, showing that AddRef is called twice.

https://user-images.githubusercontent.com/49287178/107054686-5c06c600-67c8-11eb-8c9e-aa8801afe4c7.mov

230Daniel:
The above issue has now been fixed by Paulo and I've renamed the method from ClearGuildUserCache to PurgeUserCache to be more consistent.

quinchs (approved):

MrCakeSlayer (approved):

PR #1773, opened 2021-02-09
feature: get private methods(backward compatibility)

AntonioZn:
Breaking change fix. At the moment if a command contains private method it is not going to be handled. This change will fix this.
Improvement of: https://github.com/discord-net/Discord.Net/pull/1521


SubZero0:
Looks good. Thanks.

PR #1776, opened 2021-02-14
feature: Add enterTyping optional parameter to CommandService.ExecuteAsync()

### LINKED ISSUE #1775: EnterTypingState before command execution but only for valid commands?
Given the example from [the intro to command service guide](https://docs.stillu.cc/guides/commands/intro.html):
```csharp
var context = new SocketCommandContext(_client, message);
await _commands.ExecuteAsync(context, argPos, _provider);
```

I can then wrap this snippet with a `using context.Channel.EnterTypingState()`:
```csharp
using (context.Channel.EnterTypingState())
{
    var context = new SocketCommandContext(_client, message);
    await _commands.ExecuteAsync(context, argPos, _provider);
}
```

This will cause the client to enter typing state as long as the author's message meets the criteria for executing commands, regardless if the command is actually valid.

Is there a way to use `EnterTypingState()` for only valid commands without using `BeforeExecute()` from `ModuleBase` nor `TriggerTypingAsync()`?

blai30:
## Abstract
This resolves #1775 by adding an optional parameter to the `CommandService.ExecuteAsync()` method(s).

### Additions
Optional `bool enterTyping = false` parameter is added to `CommandService.ExecuteAsync()` that will allow users to specify whether or not the client should enter typing state in the channel where a valid command was invoked. The call to `context.Channel.EnterTypingState()` will return an `IDisposable` assigned to the local variable `var typing`. By declaring `var typing` with the `using` statement, the `IDisposable` will get disposed once `ExecuteAsync()` finishes running.

### Tests
This was tested with the sample `02_commands_framework`.

### DIFF
--- a/src/Discord.Net.Commands/CommandService.cs
+++ b/src/Discord.Net.Commands/CommandService.cs
@@ -483,25 +483,27 @@ namespace Discord.Commands
         /// <param name="context">The context of the command.</param>
         /// <param name="argPos">The position of which the command starts at.</param>
         /// <param name="services">The service to be used in the command's dependency injection.</param>
+        /// <param name="enterTyping">Whether the client should enter typing state.</param>
         /// <param name="multiMatchHandling">The handling mode when multiple command matches are found.</param>
         /// <returns>
         ///     A task that represents the asynchronous execution operation. The task result contains the result of the
         ///     command execution.
         /// </returns>
-        public Task<IResult> ExecuteAsync(ICommandContext context, int argPos, IServiceProvider services, MultiMatchHandling multiMatchHandling = MultiMatchHandling.Exception)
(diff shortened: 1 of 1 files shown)


blai30:
Changed commit message.

SubZero0:
This can be easily done without this, there's no reason to add it (much less a breaking change).

blai30:
Yeah sorry about that. I have since found a much better way of doing this and was going to close this myself but totally forgot about it..

This is how I did it:
```csharp
// abstract class that inherits ModuleBase<SocketCommandContext>
protected override void BeforeExecute(CommandInfo command)
{
    base.BeforeExecute(command);
    using IDisposable typing = Context.Channel.EnterTypingState();
}
```

PR #1778, opened 2021-02-14
Removed broken URI check from EmbedBuilder

asmejkal:
The `Uri.IsWellFormedUriString()` method that is used for URL validity checking in EmbedBuilder has always been plagued with problems and causes both false positives and negatives. Details on this issue can be found [here](https://github.com/dotnet/runtime/issues/21626). 

This issue makes some images with a perfectly valid URL impossible to use in Embeds with Discord.Net. The Discord API already throws a 400 error for image URLs it considers invalid, so I believe this check is mostly redundant anyway.

Following some discussion on the Discord I should add this: Ideally the URI check should be "fixed", but speaking from experience, doing a proper URL check in C# is non-trivial, especially since there's a big chance Discord doesn't do it 100% correctly on their side either, and ideally the behaviors should match. Unless anyone wants to deal with implementing that, I believe removing it is a lesser evil.


Still34:
cc @SubZero0 @Emzi0767, opinions?

Emzi0767:
AFAIK the Uri constructor works ok, so perhaps changing types to Uri as a means of constraining input is something to investigate. Discord is also additionally restrictive on what is valid and what is not.

Emzi0767:
> The Discord API already throws a 400 error for image URLs it considers invalid

By that logic, the ratelimit handling can also be removed as Discord will just return a 429. That is not valid reasoning. Sure, the current method is flawed, but instead of removing it, find a suitable replacement.

asmejkal:
> > The Discord API already throws a 400 error for image URLs it considers invalid
> 
> By that logic, the ratelimit handling can also be removed as Discord will just return a 429. That is not valid reasoning. Sure, the current method is flawed, but instead of removing it, find a suitable replacement.

That's not a good equivalence as rate limiting is actually required by Discord, has a documented way of implementing it, and it's not about validating input. This is more like sending a bogus emoji in AddReaction or doing any other stupid thing you can do with the REST API. 

I don't see a reason for a REST API wrapper to pre-validate absolutely everything the server may reject. Especially in this case where this kind of pre-validation will not be perfect. Unless the client-side validation logic is exactly the same as server-side (impossible to guarantee), there will be the following problems:
1. There might be edge-cases where the URL still gets rejected by the server (leads to a surprise if you're relying on the client-side validation and not handling the server-side validation exception).
2. There might be cases where the client-side validation is more restrictive, making you unable to use URLs you'd otherwise be able to (as evidenced by this bug).

If you still feel like this is a no-go without client-side validation, I can at least replace the check with `Uri.TryCreate` + `UriKind.Absolute`. It might still cause problem 1 (e.g. with regard to allowed schemes), but not problem 2 as it's most likely less restrictive than Discord's validation. The parameters should've been Uri types anyway, but I suspect changing that would be a breaking change.

Emzi0767:
> I don't see a reason for a REST API wrapper to pre-validate absolutely everything the server may reject.

Sending high volumes of invalid requests results in API bans. That should be a reason enough.

> That's not a good equivalence as rate limiting is actually required by Discord

So is this, and the ratelimiter in dnet is far from perfect.

asmejkal:
> > I don't see a reason for a REST API wrapper to pre-validate absolutely everything the server may reject.
> 
> Sending high volumes of invalid requests results in API bans. That should be a reason enough.

Not sure that's true, according to the docs at least. Those mention just 401, 403, and 429, which would mean it applies to making unauthorized requests (by not checking permissions) and not honouring the rate limits.

SubZero0:
If it actually hinders using valid urls for Discord and the DotNet team is taking their time to fix it, I'm fine with removing this check.

If Discord will deny it, it's up to the user to not keep doing the same request since it's all in their control (unlike the ratelimiter that's almost entirely internal), we help preventing where we can, but denying a valid request is no good.

Issue #1703, opened 2020-12-04
API v8 upgrade


Misha12:
Actually is in configuration API v6, but this API is deprecated. Is planned upgrade to v8? Is there any problem?

SubZero0:
The latest nightly (3.0) is now on v9, so I'll be closing this issue.

PR #1787, opened 2021-03-02
Updated API version #1703 

matespinda01:
Hopefully updated the API verison. Looks like this will do it, but correct me if I missed something.
related to #1703


Emzi0767:
This will not do it. There's many different breaking changes that need to be addressed. I suggest you [read the docs](https://discord.com/developers/docs/change-log#api-and-gateway-v8).

LedgendODST:
Give this guy a banana 🍌 lmao this is big brain lmao

AntiTcb:
Good attempt, but sadly, there's more to it than bumping up the number.

PR #1791, opened 2021-03-04
Fix: Update Webhook Channel ID if channel ID has changed.

### LINKED ISSUE #1447: Modifying a webhook's channel doesn't actually modify it?
I'm trying to create a generic webhook for use across the entire server, as opposed to an obnoxious number of webhooks per channel. I create the singular webhook and reassign it's Channel, but it still posts in the original specified channel in server settings. ChannelID reassignment doesn't work either.

```
using (var client = new DiscordWebhookClient("https://discordapp.com/api/webhooks/whatever"))
{
await client.ModifyWebhookAsync(x => x.Channel = BotClient.GetChannel(Context.Channel.Id) as SocketTextChannel);
await client.ModifyWebhookAsync(x => x.ChannelId = Context.Channel.Id);
await client.SendMessageAsync("test", false, null);
}
```

AntiTcb:
Fixes #1447.

### DIFF
--- a/src/Discord.Net.Rest/Entities/Webhooks/RestWebhook.cs
+++ b/src/Discord.Net.Rest/Entities/Webhooks/RestWebhook.cs
@@ -12,10 +12,10 @@ namespace Discord.Rest
         internal ITextChannel Channel { get; private set; }
 
         /// <inheritdoc />
-        public ulong ChannelId { get; }
-        /// <inheritdoc />
         public string Token { get; }
 
+        /// <inheritdoc />
+        public ulong ChannelId { get; private set; }
         /// <inheritdoc />
         public string Name { get; private set; }
         /// <inheritdoc />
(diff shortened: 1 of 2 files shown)


SubZero0 (changes requested):
`Channel` property is still outdated and might not match the `ChannelId` anymore.

AntiTcb:
> `Channel` property is still outdated and might not match the `ChannelId` anymore.

It'll have to remain outdated, there's no means of getting a fresh Channel object without hitting the API.
If we want to remove the property altogether, that should be done in a separate PR.

SubZero0:
> It'll have to remain outdated, there's no means of getting a fresh Channel object without hitting the API.

It's fine if we can't remove it (much less in a patch), but it should be noted that it might be outdated and using the `ChannelId` property is more reliable as a remark or similar so there's less chance of people getting an unexpected value from it.

SubZero0:
I just noticed the Channel is internal and not used anywhere so nevermind.

PR #1797, opened 2021-03-14
[Docs] - Updated to use async main

Hitmasu:
Currently, we use GetAwaiter().GetResult() to run our MainAsync method.  In C# 7.1 or above, we can create an async main method:

```cs
public static async Task Main(string[] args){}
```

This way, we can simplify some things on documentation like this:

```cs
public static void Main(string[] args)
	=> new Program().MainAsync().GetAwaiter().GetResult();

public async Task MainAsync()
{
}
```
-----
After PR:

```cs
public static async TaskMain(string[] args) 
	=> await new Program().MainAsync();

public async Task MainAsync()
{
}
```

### DIFF
--- a/docs/guides/getting_started/first-bot.md
+++ b/docs/guides/getting_started/first-bot.md
@@ -80,15 +80,11 @@ recommended for these operations to be awaited in a
 properly established async context whenever possible.
 
 To establish an async context, we will be creating an async main method
-in your console application, and rewriting the static main method to
-invoke the new async main.
+in your console application.
 
 [!code-csharp[Async Context](samples/first-bot/async-context.cs)]
 
-As a result of this, your program will now start and immediately
-jump into an async context. This allows us to create a connection
-to Discord later on without having to worry about setting up the
-correct async implementation.
+As a result of this, your program will now start into an async context.
 
 > [!WARNING]
 > If your application throws any exceptions within an async context,
--- a/docs/guides/getting_started/samples/first-bot/async-context.cs
+++ b/docs/guides/getting_started/samples/first-bot/async-context.cs
@@ -1,7 +1,7 @@
 public class Program
 {
-	public static void Main(string[] args)
-		=> new Program().MainAsync().GetAwaiter().GetResult();
+	public static async Task Main(string[] args)
+		=> await new Program().MainAsync();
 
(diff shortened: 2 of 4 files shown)


Hamsterland:
On another note there should be an example using a generic host too.

Still34:
This was a deliberate choice to not implement async main as a doc example. Due to the way async main is implemented at core runtime, exceptions will not propagate into the debugger properly, which is why we are currently sticking to the legacy method until the .NET team improves upon this.

Joe4evr:
> Due to the way async main is implemented at core runtime, exceptions will not propagate into the debugger properly

Waitwhat? I've been using Async Main since it landed and never had any kind of problem with exceptions and debugging.

HaloFour:
As "async Main" is really just C# syntax candy for emitting an entry point that calls `GetAwaiter().GetResult()`, I'm sure that the C# team would be interested to hear in any friction that teams have experienced in adopting it.

Still34:
Reopening

SubZero0:
As Halo said, the compiler will literally do the same as we current do under the hood.

I would leave the docs up to Still, but if you are going to change it, there's no need to async/await, you can return the Task itself.

Hitmasu:
> I would leave the docs up to Still, but if you are going to change it, there's no need to async/await, you can return the Task itself.

Yeah, returning the Task itself it's better and more clean

```cs
static Task Main(string[] args) => new Program().MainAsync();

public async Task MainAsync()
{

}
```

Still34:
Didn't have time to do a proper response earlier:
My bad, I thought we had a discussion about this a while ago back when async main was first rolled out; at one point this change was pushed through then rolled back due to negative community feedback. Perhaps things have changed. I'll follow up to make sure the behaviors are indeed as expected before merging.

AntiTcb:
> Didn't have time to do a proper response earlier:
> My bad, I thought we had a discussion about this a while ago back when async main was first rolled out; at one point this change was pushed through then rolled back due to negative community feedback. Perhaps things have changed. I'll follow up to make sure the behaviors are indeed as expected before merging.

From what I remember, it was an issue of VS2017 just not being updated properly for it, but likely, a later update or VS2019 handles it nicely.

PR #1801, opened 2021-03-16
Feature: Add missing property to MESSAGE_REACTION_ADD event

Waterball12:
# Description

This pull request adds a missing property for the MESSAGE_REACTION_ADD event

## Changes 

I have added the member property to the reaction event, I made it optional because I saw that it's also used in the MESSAGE_REACTION_REMOVE but in that event is not present. In the DiscordSocketClient I added some lines that checks if the member property is present and if yes use it instead of the user in the cache

```diff
Reaction.cs

+++ [JsonProperty("member")]
+++ public Optional<GuildMember> Member { get; set; }
```

## Motivation

Uses the member property when present in the event instead of rely on the cache

## Breaking changes

No breaking changes

### DIFF
--- a/src/Discord.Net.WebSocket/API/Gateway/Reaction.cs
+++ b/src/Discord.Net.WebSocket/API/Gateway/Reaction.cs
@@ -1,4 +1,4 @@
-﻿using Newtonsoft.Json;
+using Newtonsoft.Json;
 
 namespace Discord.API.Gateway
 {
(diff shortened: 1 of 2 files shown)


zbroderson:
~~I dont think Discord actually sends user data through the reaction on MESSAGE_REACTION_ADD.~~

~~This documentation seems to describe the "partial emoji object" the event sends~~
~~https://discord.com/developers/docs/resources/emoji#emoji-object-gateway-reaction-standard-emoji-example~~

~~Could easily be wrong though.~~

Annnnnnnnnd I've been wrong on the internet

quinchs:
> Annnnnnnnnd I've been wrong on the internet

Haha welcome to the interwebs

Issue #1805, opened 2021-03-22
Is there any way to use the HttpClientFactory for providing the HttpClients?


emorell96:
So I have being trying to understand how the library, in particular the .net.rest part of the library handles creation of the HttpClient objects, and it seems I might have found the answer in [Line 31 of DefaultRestClient](https://github.com/discord-net/Discord.Net/blob/1b64d19c845cb7c612a1c52288c8b44cff605105/src/Discord.Net.Rest/Net/DefaultRestClient.cs#L31).

If I understand the code (which I have browsed quite a bit, but I still wanted to confirm with you all since you know this library 10x better), the DefaultRestClientProvider everytime it gets called, it will return a new HttpClient with a new HttpHandler.

So if I were to register a DiscordRestClient as a scoped, or transient service, every time a new instance of DiscordRestClient would be created a new HttpClient and HttpHandler would be created as well leading to exactly the issue talked about in this:

https://www.aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/
https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests

Would there be a way to use the IHttpClientFactory to create HttpClients in this library?

Maybe a different DefaultRestClient that could use the IHttpClientFactory?

I'd be happy to contribute if this is in fact an issue.

Thanks!

SubZero0:
> So if I were to register a DiscordRestClient as a scoped, or transient service, every time a new instance of DiscordRestClient would be created a new HttpClient and HttpHandler would be created as well leading to exactly the issue talked about in this:

Exactly, so you shouldn't. It needs to be added as a singleton as it was designed, then it will only use a single HttpClient as shown as the "fix" in the exact page (wouldn't solve the dns issues but well).

> Maybe a different DefaultRestClient that could use the IHttpClientFactory?

You can already do that by having a custom provider to achieve the same, except it still shouldn't be added as anything but  a singleton.

**Now, why singleton?** All the clients have an internal ratelimiter to prevent you from hitting them and being possibly banned by Discord for it. By creating multiple clients, you are effectively making it useless.

Basically, add it as singleton, you have this option.

quinchs:
Closed in #1806

Issue #152, opened 2016-07-28
Add RefreshTokenAsync() to DiscordRestClient


Auralytical:

RogueException:
Refresh tokens never expire, but access tokens expire every week (I think?) currently. We need an easy way to refresh them for private RPC.

RogueException:
We could also add a new TokenType of Refresh, which will automatically refresh tokens.

RogueException:
We should discuss how we want this first. TokenType.Refresh makes things complicated since it now needs a client secret and client id, in addition to the refresh token. We could pass these in through DiscordRestConfig?

If we ignore this, then simple RPC bots need to refresh tokens themselves... is that too complicated for the average bot?

PR #1806, opened 2021-03-22
Adding support for Refresh Tokens, as well as adding a new HttpClientFactoryRestClient #152 #1805

emorell96:
In a word, this change adds some DI to DiscordRestClient.

The main issue I was having with Discord.Net when I wanted to use in StockDrops website is that it instantiates a new HttpClient and a new HttpClientHandler every time it gets created (see [#1805](https://github.com/discord-net/Discord.Net/issues/1805)).

Ln 31 above gets called everytime the RestClientProvider delegate is called and generates a new HttpClient. Even if this HttpClient is properly disposed this is not great when you might need to use it in an ASP.NET project where you might need a Scope or even Transient object. Since it could lead to Socket Exhaustion. See [Asp.Net Monsters](https://www.aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/) or [Use IHttpClientFactory to implement resilient HTTP requests](https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests).

For that reason, I want a way I could use IHttpClientFactory in my website without having to recode everything that you guys have done in this terrific library. So I decided to fork it and add an extension method AddScopedDiscordRestClient and AddTransientDiscordRestClient.

The extension methods allows for easy use of DI so that a uses in their startup.cs can just use that extension in their configure services method. (see the new sample created)

This is where that got added: src/Discord.Net.Rest/Extensions/ServiceCollectionExtensions.cs

The main issue with this PR is that it adds 2 dependencies: Microsoft.Extensions.DependencyInjection and Microsoft.Extensions.Http to get the AddHttpClient method.

Are these 2 dependencies worth it? IMHO yes. Because they could add a lot of options for cleaner integration with DI.

Plus if for 3.0 you were to add Polly, you could build resiliency using Polly Policies since the IHttpFactory supports them by default. That could help with Discord's API constant disconnections. Something like this: [Implement HTTP call retries with exponential backoff with IHttpClientFactory and Polly policies](https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/implement-http-call-retries-exponential-backoff-polly)

The new src/Discord.Net.Rest/Net/HttpClientFactoryRestClient.cs and src/Discord.Net.Rest/Net/HttpClientFactoryRestClientProvider.cs are almost 100% copies of the default ones, except that I made a few changes:

- The HttpClientFactoryRestClient now takes as argument the already configured HttpClient generated by IHttpClientFactory when instantiating it by the HttpClientFactoryRestClientProvider.
- Since we provide an instantiated HttpClient I removed the part the added the new HttpClient.
- The HttpClientFactoryRestClientProvider is no longer static. It is a class that gets added as either scoped or transient when you call AddTransientDiscordRestClient or AddScopedDiscordRestClient on an IServiceCollection.
- On construction the Instance gets defined by using the HttpClient generated by the IHttpClientFactory.
- Moved the UseProxies parameter to the AddTransientDiscordRestClient  or AddScopedDiscordRestClient  since that's where the message handler gets configured now.

That's basically all the changes. Plus the new methods/classes are all documented with ``<summary>`` etc.


Hawxy:
As much as I'm personally a proponent of `HttpClientFactory` integrations, I see this as far too heavy handed of a retrofit. Usage of Discord.NET as a pure rest client is very low relative to WS/commands so these extensions would not see a lot of use.

Ideally a simpler solution should be sought, such as exposing a `HttpClient` property on the constructor of the rest client and letting it be passed in via the provider, so at least you could wire it up to to the `HttpClientFactory` via a factory lambda as you're doing in your extensions within your own project.

emorell96:
See the latest commit. I changed it so that the provider becomes unnecessary. It now uses the lambda in the extension method.

SubZero0:
As I said in the issue, the rest client was designed as a singleton, the solution you propose here will make the ratelimiter useless and it will still keep the same HttpClient forever so I believe (if my knowledge of how the factory works and you implemented it here) it will still have the same issues as if you created a custom client provider with the same HttpClient for all clients (would prevent the socket exhaustion).

Not only that but this PR also brings a lot of downsides as adding two new depencies that were removed to not make Discord.Net specifically require Microsoft's DI and adding an entire blazor template project to show a single line.

I believe the correct way would be creating a guide on the docs describing how to use Discord.Net in these cases, that imo should be something as: add the rest client as a singleton (as it was designed) with a custom provider that will "create" a client from the factory instead of using a single httpclient instance (to prevent dns issues since socket exhaustion won't happen anyways).

emorell96:
Definitely a better documentation is in order, the Rest client has almost 0 docs or guides. If I had known a lot of this before, I would have done them differently. None of the information you provided here is anywhere in the docs. And a sample with DiscordRestClient is needed; maybe a smaller c# console program, but anyways it's not like when you download the nuget you download the samples as well, the sample would serve as an example for possible uses. There's 0 examples on how to use it.

Just compare for example the docs of Discord.py and Discord.Net they are miles ahead.

Because I know most of the work on the library was meant for bots using WebSockets, but the Rest library is there and it's very useful for integration specially with websites (our use case), and there's 0 samples or docs about this, all is about WebSockets and bots, never about using Rest for example to pull the current user, or other. That's why I thought you'd appreciate a sample to show case that Discord.Net.Rest also exist.

As for the dependencies, I agree. I even mentioned it in the PR. I think the best solution would be the extensions methods offered separately the issue is that without creating a special RestClient that takes an HttpClient, you cannot make the extensions.

I even tried that before, importing the nuget and build the extensions by implementing an IRestClient, IDisposable as an almost carbon copy of the default ones. Now the issue with that was that MultipartFile in line 118 is marked as internal so I cannot reuse it. So the copied RestClient woudln't build.

The solution could be to either make MultipartFile public so that it can be used by external extensions, or integrate an IRestClient that takes an HttpClient in the constructor so that it can be used in the RestClientProvider delagate in outside extensions that would allow for an easy method like AddDiscordRestClient(configuration delagate in here).

I am thinking of adding that configuration so that the Discord Rest client can also do things like take the client Id, and Secret Id so that it can do things like refresh the token, or even pull a new token from the code given by Oauth2.

Anyways I am going to be using my modifications in our websites since it simplifies my life quite a bit, with the exception that the method will add it as a singleton not a transient or scoped method so not to lose the rate limiter.

I would say the big advantage of a method like that would be the possibility of adding things like Polly to the HttpClient making it more resilient. That's definitely why I wanted to do it this way. 

That's why maybe making a library called Discord.Net.Extensions where people could contribute extension methods that add functionality IMHO could be a good idea, a library that is not part of the Core, or the full Discord.Net but that people can download if there's extensions that seem interesting to them. For example right now in our private blazor project we have an HttpClient that handles getting the token from Discord Oauth2 endpoint since Discord.Net doesn't do it, but if in the future I add this functionality instead to Discord.Net.Rest then I could only use Discord.Net.Rest for everything, now the issue is that for that I need the client id and secret the best way to grab this info is through a configuration delegate and that configuration delegate I'd like to grab it as part of an AddDiscordRestClient method so that it's clean, and easy.

Anyhow thanks for your time guys, I will work on adding the functionality of token refresh and token acquisition and we can talk again later. For now I will remake the methods to create a singleton DiscordRestClient with an HttpClient generated through the factory (btw the factory doesn't give you the same HttpClient each time, the factory what it does is to reuse the HttpHandler - the one that leads to socket exhaustion - by grabbing it from a pool of HttpHandlers. )

emorell96:
After a nice day of coding I have added all the features I said previously: basically a GetTokenAsync method that people can call to request either a new token using a refresh token or a code to get a token for the first time. This address the issue [#152](https://github.com/discord-net/Discord.Net/issues/152) and adds it as said in that issue: through DiscordRestConfig. It adds ClientId and ClientSecret as 2 parameters of DiscordRestConfig. They are by default null so they are optional and only needed in the GetTokenAsync method.

The method uses the buckets and the rate limiter as well correctly.

Finally, I have removed the extra dependencies, and moved the extension methods to the sample as a way to show how it would be possible to make your life a little simpler if you needed to add a DiscordRestClient to your services multiple times (like different projects in the same solution, you write your extension once and then use it wherever needed).

I have tested it with that sample (with my own client Id, and Client secret which of course I didn't share with the git), feel free to try it out.

quinchs:
I'm going to have to agree with SubZero0 here, this basically defeats how the current internal ratelimit system works.

Taken from the issue:
"Now, why singleton? All the clients have an internal ratelimiter to prevent you from hitting them and being possibly banned by Discord for it. By creating multiple clients, you are effectively making it useless."

If you really needed to I'm sure its possible to somehow use the `RestClientProvider` in the `DiscordRestConfig`

PR #1832, opened 2021-04-30
Grab correct Uses value for vanity urls

evman182:
The vanity-url endpoint returns uses, and this can be used when populating the model.

### DIFF
--- a/src/Discord.Net.Rest/API/Common/InviteVanity.cs
+++ b/src/Discord.Net.Rest/API/Common/InviteVanity.cs
@@ -6,5 +6,7 @@ namespace Discord.API
     {
         [JsonProperty("code")]
         public string Code { get; set; }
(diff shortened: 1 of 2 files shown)


SubZero0:
Did you test this? I don't have any servers with a vanity url so I can't know if it's working as expected, considering we are also a few versions behind.

evman182:
It worked last I checked. See API response here: 
https://discord.com/developers/docs/resources/guild#get-guild-vanity-url

SubZero0:
I did see that example but the docs are also written for v8 while we are at v6 so it might not match 1:1 so a live test is more reliable.

evman182:
Just tested my branch against my server. It's working

SubZero0:
Thank you, I'll merge it on the next dev.

PR #1838, opened 2021-05-12
fix: Account for team in RequireOwnerAttribute

TheodoorP:
# Abstract

This pull request adds support for teams in the `RequireOwnerAttribute`, which currently only works when the bot's owner is a user. 

## Changes

When the `Team` property in an application isn't null, this code will check whether the user whose permissions are being checked is part of the application's team. If, however, the `Team` property is null the code will fall back to its original behavior.

# Tests

This pull request has been tested live, and [is proven to work with personal and team applications](https://storage.googleapis.com/theodoordev.appspot.com/6430b5d09255476).

# Possible Caveats:

None

### DIFF
--- a/src/Discord.Net.Commands/Attributes/Preconditions/RequireOwnerAttribute.cs
+++ b/src/Discord.Net.Commands/Attributes/Preconditions/RequireOwnerAttribute.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Linq;
 using System.Threading.Tasks;
 
 namespace Discord.Commands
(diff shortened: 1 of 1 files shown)


SubZero0:
This change wasn't done so the behavior of this attribute wouldn't change and people wouldn't be surprised by it, considering this attribute is most likely used to control the flow of sensitive commands. This could be added as a new attribute.

TheodoorP:
I would argue that this is exactly the behaviour that someone would expect from this attribute. Seeing as ([quoting discord here](https://i.imgur.com/5QdlDGi.png)) "team admins have full ownership over the team's applications as if they owned them". Why would having different attributes based on the fact that some bots may have one owner and others may have multiple owners make sense, given that both types of owner have the exact same privilege over the bot?

SubZero0:
Your argument doens't change the fact that it would change the behavior. A person could still have different authorities depending if the person is the main owner or a team member, shouldn't assume what people must or must not do specially in such an attribute most likely used in sensitive commands.

The discord description also mentions "Teams _currently_ support only the Admin role", so from my understading this could change and this assumption could turn false at any time.

TheodoorP:
I've double-checked the [docs for the Team Members object](https://discord.com/developers/docs/topics/teams#data-models-team-members-object) and at the time of writing it specifically states "will always be ["*"]" for the permissions array, which I would assume is related to the role assigned to the user. I could add a check for the wildcard permission if you would prefer that though.

I've also noticed a minor thing that should probably be changed in my current logic, that being that an invited member is also seen as a team owner, which should probably not be the case.

What if instead of merging to the current dev branch I merged into the 3.0.0-dev branch? As I would assume you would be more open to potentially breaking changes there that (in my mind at least) make sense?

quinchs:
I agree with SubZero0 here, a `RequireTeam` attribute would be better fitting then changing the `RequireOwner` attribute

PR #1868, opened 2021-06-05
3.0 serialization source generator

FiniteReality:
Working in a separate branch so that more testing and discussion can be done without screwing up history. (I.e. all of the changes can be squashed when merging.)

This is the fabled 3.0 serialization source generator. It's by no means finished, and it's by no means the prettiest code either. Some decent code review and cleanup suggestions would be nice - this is just the minimum viable product in order to prove we're able to even do this.


FiniteReality:
Design discussion from dev server:
- Current records don't have properties documented
- Primary ctor was primarily for convenience w/ naming and ordering, since parameter name = property name
- If we're specifying properties manually, we don't need a ctor at all and can just use `init` accessors

So, that means we'll do something like this:
```cs
public enum ChannelType
{
    GuildText
}

public record Channel
{
    public ChannelType Type { get; init; }
    public ulong Id { get; init; }
}

public record GuildTextChannel : Channel
{
    public GuildTextChannel()
    {
        Type = ChannelType.GuildText;
    }
}
```

SubZero0:
As discussed in DM, we should have a way to not be tied to the same naming as discord.
It could be solved with:
- Override attribute: overrides the automatic snake case conversion
- Json name attribute: will be used instead of the snake case conversion

I feel the json name attribute would be the best option for a few reasons:
1. If we are going to check every attribute for the override one, we could just go for a sure one that will already have the correct naming.
2. We can use it as a way to describe if the parameter should be ignored or not.
3. Easier to use instead of having to know how the library will convert the property name internally to match it (for future contributions).

PR #1888, opened 2021-07-06
Allow clients to send 'Gateway Voice State Update' command

### LINKED ISSUE #1882: How to change selfMute/selfDeaf after connecting audio client?
Hello,

When connecting to an audio channel, it is possible to set selfMute and selfDeaf params, see: https://github.com/discord-net/Discord.Net/blob/1b64d19c845cb7c612a1c52288c8b44cff605105/src/Discord.Net.Core/Entities/Channels/IAudioChannel.cs#L21

However, I don't see any API for updating those params after connecting to an audio channel. Is it intentional? Or maybe I just missed something?

I know that it is possible to update `Mute` and `Deaf` parameters via SocketGuildUser class:

```
voiceChannel.GetUser(userId).ModifyAsync(props =>
{
    props.Deaf = false;
    props.Mute = false;
});
```

But those props rather relate to 'server-mute/deaf', no client mute/deaf.

ppaneksamsung:
Fixes #1882


Emzi0767:
Instead of giving people a loaded footgun, how all you instead provide a method for setting just the required properties?

ppaneksamsung:
Well, another option is to provide 'ModifyAsync' method (something like https://github.com/discord-net/Discord.Net/blob/1b64d19c845cb7c612a1c52288c8b44cff605105/docs/_overwrites/Common/ObjectProperties.Overwrites.md), which will allow to modify only the required properties. This approach seems to be widely used by this project. What do you think?

Emzi0767:
That would be the correct approach

ppaneksamsung:
I see one drawback of this 'ModifyAsync' approach. `self_mute` and `self_deaf` fields in the `Voice State Update` command are not marked as optional in the official docs (https://github.com/discord/discord-api-docs/blob/master/docs/topics/Gateway.md#update-voice-state), so we always have to send all required fields. This means that we will have to maintain a state on our side to handle scenarios like:

```
await voiceChannel.ModifyAsync(props =>
{
    props.SelfDeaf = true;
});
await voiceChannel.ModifyAsync(props =>
{
    props.SelfMute = false;
});
```

This state should be probably maintained by `SocketGuild` class. This is not a big deal, but I just wanted to share this minor issue.

quinchs:
`AudioChannelProperties` doesn't seem fitting for modifying the current users self mute/deaf, why not add this to `GuildUserProperties` and have some precondition to check if were setting `SelfMute` or `SelfDeaf` on the current user.

PR #1902, opened 2021-07-30
feature: Add SocketUnknownChannel

SubZero0:
This is another way of resolving, partially, the issue with Threads.
It has a lot less changes than the other ( #1901 ), but with the disavantage that nothing will apparently break for who's using Discord.Net.

This solution adds a new channel type, `SocketUnknownChannel`, that inherits from a `ISocketMessageChannel` so making it easy to do basic functionality (send messages, files, get messages, etc), but has the drawback that now the `Name` property could be `null`, as the only information received about it is the id.

So while it would fix the issue of not being able to process messages without a cached channel, it might cause a few unexpected issues for the developer, also by adding the possibility of a third type of "basic" channel (private, guild, unknown).

I still need to check the structure for the parsers and attributes to see if they are still working as expected.

### Possible changes
- Rename `SocketUnknownChannel` to `SocketUnknownMessageChannel`

## Changes
- Add a new channel type: `SocketUnknownChannel`
This channel might be present inside message objects returned by the events MessageReceived and MessageUpdated.

### DIFF
--- a/src/Discord.Net.WebSocket/DiscordSocketClient.cs
+++ b/src/Discord.Net.WebSocket/DiscordSocketClient.cs
@@ -1262,7 +1262,7 @@ namespace Discord.WebSocket
                                     var data = (payload as JToken).ToObject<API.Message>(_serializer);
                                     var channel = GetChannel(data.ChannelId) as ISocketMessageChannel;
 
-                                    var guild = (channel as SocketGuildChannel)?.Guild;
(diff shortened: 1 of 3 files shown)


quinchs:
The idea for an unknown entity type is good for when discord initially releases new entity types, people running older version don't have to update to fix breaking changes. I think we should have different types of unknown channels ex: SocketUnknownVoiceChannel, SocketUnknownTextChannel, etc or have a generalized SocketUnknownChannel with just the id.

PR #1906, opened 2021-08-07
feature: Add Get***Url methods

Shane4368:
**SemVer: minor**
- Add Get***Url methods to allow image format and size to be specified
- Add Lottie to enum ImageFormat
- Deprecate URL properties

*Reference: https://discord.com/developers/docs/reference#image-formatting-cdn-endpoints*

### DIFF
--- a/src/Discord.Net.Core/CDN.cs
+++ b/src/Discord.Net.Core/CDN.cs
@@ -12,23 +12,35 @@ namespace Discord
         /// </summary>
         /// <param name="teamId">The team identifier.</param>
         /// <param name="iconId">The icon identifier.</param>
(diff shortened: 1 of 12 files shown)


quinchs:
Implemented in #1923

Shane4368:
After looking at the code (example: [L95](https://github.com/discord-net/Discord.Net/blob/dev/src/Discord.Net.Core/CDN.cs#L95)), some of the features in this PR are still not implemented.

This PR was intended to deprecate _all_ URL properties and replace them with methods so the format (example: png, gif, jpeg) and size (example: 32, 128, 1024) could be specified.
`RestGuild#IconUrl` would become `RestGuild#GetIconUrl(ImageFormat, ushort)` for example.

PR #1921, opened 2021-09-13
feature: MoveAsync and DisconnectAsync

yannikHoeflich:
This allows library users to move and disconnect other discord users more easily.

My reason for this change is, that this is much faster to work with and also more clearly what it does than working with `ModifyAsync`

# Changes
- **IGuildUser**: added new method `MoveAsync` (to move the user to another voice channel)
- **IGuildUser**: added new method `DisconnectAsync` (to disconnect the user from its current voice channel)


quinchs:
Implemented in #1923

PR #1943, opened 2021-11-24
feature: Handle bidirectional usernames

### LINKED ISSUE #1829: Handle bidirectional unicode usernames in user ToString
Say there's a user with name that contains RTL characters, e.g. `"שרלוק"`

The current implementation of `SocketUser.ToString()` (https://github.com/discord-net/Discord.Net/blob/5b244f215b76ec7c43e2329b72a62ddd6f6e32e2/src/Discord.Net.WebSocket/Entities/Users/SocketUser.cs#L114) and `RestUser.ToString()` (https://github.com/discord-net/Discord.Net/blob/5b244f215b76ec7c43e2329b72a62ddd6f6e32e2/src/Discord.Net.Rest/Entities/Users/RestUser.cs#L105) naively concatenate that with `#` and the discriminator.

Unfortunately, the resulting rendered output then looks completely wrong: `"שרלוק#1234"`

image

The desired output should maintain the `#{Discriminator}` without being affected by the RTL text: `"⁦שרלוק⁩#1234"`

image

---

Wrapping the username in Unicode bidi isolation control characters fixes this. For example:

```csharp
$"\u2066{Username}\u2069#{Discriminator}"
```

That starts a LTR-based isolated run, puts the username in there, and ends the run. An alternative is to start the run with `\u2068`, which would make the base either LTR or RTL based on the first character in the run... but I think Discord probably defaults LTR so `\u2066` seems more appropriate to me.

See also: https://www.w3.org/International/questions/qa-bidi-unicode-controls

quinchs:
## Summary
This PR fixes #1829 by implementing bi-directional isolation characters when concatenating the username and discriminator.

### DIFF
--- a/src/Discord.Net.Core/Format.cs
+++ b/src/Discord.Net.Core/Format.cs
@@ -106,7 +106,7 @@ namespace Discord
         }
 
         /// <summary>
-        ///     Formats a users username + discriminator while maintaining bidirectional unicode
+        ///     Formats a user's username + discriminator while maintaining bidirectional unicode
         /// </summary>
         /// <param name="user">The user whos username and discriminator to format</param>
         /// <returns>The username + discriminator</returns>
--- a/src/Discord.Net.Core/Format.cs
+++ b/src/Discord.Net.Core/Format.cs
@@ -104,5 +104,15 @@ namespace Discord
             var newText = Regex.Replace(text, @"(\*|_|`|~|>|\\)", "");
             return newText;
         }
+
+        /// <summary>
+        ///     Formats a users username + discriminator while maintaining bidirectional unicode
+        /// </summary>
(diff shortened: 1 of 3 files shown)


src/Discord.Net.Core/Format.cs
```diff
@@ -104,5 +104,15 @@ public static string StripMarkDown(string text)
             var newText = Regex.Replace(text, @"(\*|_|`|~|>|\\)", "");
             return newText;
         }
+
+        /// <summary>
+        ///     Formats a users username + discriminator while maintaining bidirectional unicode
```
MrCakeSlayer: users => user's

MrCakeSlayer (changes requested):

Issue #1878, opened 2021-06-26
Audio binaries link to dead domain


Emzi0767:
The link provided on [Sending Voice](https://docs.stillu.cc/guides/voice/sending-voice.html) page from the documentation links to a dead domain: https://discord.foxbot.me/binaries/ (WARNING: link defunct, domain now leads to domains flagged as dangerous)

There's a mention of this in a now-defunct issue (thusly, with terrible visibility), https://github.com/discord-net/Discord.Net/issues/1110#issuecomment-868881095. Provided there are also links to builds provided for another library, which require renaming for use in D.Net. It would be a good idea to either update the link to point to the binaries linked there (with necessary instructions), or rehost them somewhere else, in a manner which does not require modifying them.

Provided below is a tag quote from Discord API:

> **Windows 64-bit**: https://dsharpplus.github.io/natives/vnext_natives_win32_x64.zip
> **Windows 32-bit**: https://dsharpplus.github.io/natives/vnext_natives_win32_x86.zip
> **See also**: https://dsharpplus.github.io/natives/index.html
> **Discord.Net**: https://discord.foxbot.me/binaries/ and see `?ahh binaries`
> If you want to host these, make sure to link relevant opus and sodium licenses, and, if possible, please mention that I (Emzi0767) built them. Muchas gracias.

PR #1944, opened 2021-11-24
Add Voice binaries

quinchs:
## Summary
This PR adds the `voice-natives` folder containing the zipped binaries for opus and libsodium provided by @Emzi0767 #1878. It also updates the [Sending voice](https://docs.stillu.cc/guides/voice/sending-voice.html) documentation with the url pointing to the binaries


Emzi0767:
After all these years...

quinchs:
> After all these years...

Better late then never!

MrCakeSlayer (approved):

PR #1967, opened 2021-12-15
Pack metapackage with Discord.Net.Core version

adamgauthier:
To avoid any mismatch between both versions and duplication in the code,
the .nuspec file is simplified. The build process now ensures the
version is set to the exact version Discord.Net.Core is packed with.

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -91,16 +91,6 @@ jobs:
           if: startsWith(github.ref, 'refs/tags/')
           run: echo "IsTagBuild=true" >> $GITHUB_ENV
           
-        - name: Generate Beta Suffix
-          if: env.IsTagBuild == 'true' && contains(github.ref_name, '-')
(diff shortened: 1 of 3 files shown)


quinchs:
@SubZero0 Thoughts?

adamgauthier:
For context, there is currently an issue where the metapackage is packed with dependencies of format `version-buildnumber` instead of the actual version of the pushed dependencies (`version-dev-buildnumber`). This leads to errors when attempting to use the metapackage:

```
NU1603    Discord.Net 3.0.0-dev-20211215.1 depends on Discord.Net.Commands (>= 3.0.0-20211215.1) but Discord.Net.Commands 3.0.0-20211215.1 was not found. An approximate best match of Discord.Net.Commands 3.0.0-dev-20210524.5 was resolved.
```

This change would make it impossible to happen in the future (assuming all packages are in sync with Discord.Net.Core)

PR #1970, opened 2021-12-18
Fix NRE on http error handling

adamgauthier:
In some cases, the code is not able to deserialize API.DiscordError, but
it attempts to use its message to throw an exception. This change uses
null as the exception reason in that case.


quinchs:
This was fixed in https://github.com/Discord-Net-Labs/Discord.Net-Labs/pull/362 but I'll merge it in here

quinchs (approved):

adamgauthier:
Ah I see, sorry it is a bit difficult to follow what changes have been made with the fork and everything

quinchs:
No worries

Issue #1975, opened 2021-12-19
RequireRoleAttribute checks for existence of Role inside a guild, not on the user.


JMWren:
Hey there,

When testing the new interactions I found that the RequireRole precondition for interactions was passing every time, even when the user did not have the specified role. 

It seems that the RequireRoleAttribute precondition for interactions is checking the existence of the Role inside of the guild when supplied with a string value, rather than for the user. This causes the condition to succeed if the role exists in the guild.

![image](https://user-images.githubusercontent.com/29394737/146692501-8c30fc19-b273-4a55-ac3d-e6472622d0ed.png)

Here is the code sample I was running.

![image](https://user-images.githubusercontent.com/29394737/146692721-094a89dc-c9a8-4935-9de1-9534166fe279.png)

quinchs:
cc: @Cenngo

Cenngo:
This should be an easy fix. @quinchs Where do you want me to push it? Here or Labs

quinchs:
Push it here and il make a labs commit

PR #1978, opened 2021-12-20
return error when precondition fails

Bakersbakebread:
return `Task.FromResult()` when condition is false for user not having role required


Cenngo:
sadly this doesnt fix the issue with this precondition on its own, L#62 also needs to be updated. Ill push the fix for #1975 shortly.

Cenngo:
> sadly this doesnt fix the issue with this precondition on its own, L#62 also needs to be updated. Ill push the fix for #1975 shortly.

#1980

PR #2021, opened 2022-01-07
Add SuppressEmbeds() method

### LINKED ISSUE #2020: Suppress Embeds
Before 3.0.0 there used to be a method similar to `context.message.suppressEmbeds()` that would remove embeds on messages from other users; However, it appears that it has been removed. I have also tried to do `await context.Message.ModifyAsync(item => { item.Flags = MessageFlags.SuppressEmbeds; });` but that results in the error `System.ArgumentException: Argument cannot be blank. (Parameter 'Content')`. 

According to the [Discord API docs](https://discord.com/developers/docs/resources/channel#edit-message), other users are allowed to edit flags as long as they have the Manage messages permission and the SuppressEmbeds flag can be set/unset after sending the message.

bman46:
I added a SuppressEmbeds() method back to allow for suppressing other users embeds. It is based off of the ModifyAsync() method. 
Fixes #2020

### DIFF
--- a/src/Discord.Net.Core/Entities/Messages/IMessage.cs
+++ b/src/Discord.Net.Core/Entities/Messages/IMessage.cs
@@ -270,7 +270,14 @@ namespace Discord
         ///     A task that represents the asynchronous removal operation.
         /// </returns>
         Task RemoveAllReactionsForEmoteAsync(IEmote emote, RequestOptions options = null);
-
+        /// <summary>
+        ///     Suppresses embeds for a specified message.
+        /// </summary>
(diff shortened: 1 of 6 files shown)


SubZero0:
This method was removed because this endpoint doesn't exist anymore. If there's an issue with ModifyAsync, it should be fixed instead of trying to readd something that was removed from the library.

At worst, if quinchs really wanted it, would be as an extension method, not readded to the interface.

bman46:
Ok, makes sense. I wasn't sure why it was removed and couldn't find anything in the change logs mentioning it. I'll try to fix the ModifyAsync method instead.

PR #2044, opened 2022-01-15
Use ApplicationId for interactions

quinchs:
## Summary
This PR fixes the interaction api methods using the bot id instead of the application id, for newer bots this isn't an issue but for older bots the application id isn't always the bot id.


moiph:
LGTM

PR #2062, opened 2022-01-27
Add return statement to precondition handling

Cenngo:
Adds a return statement to the Discord.Commands' command handling method to return the failed `PreconditionResult` instead of `MatchResult`, which shouldn't be exposed to the user in this case.

Waiting for @MrCakeSlayer.


quinchs:
@MrCakeSlayer Status?

MrCakeSlayer (approved):

PR #2077, opened 2022-02-03
Patch; Clone being created on updated entity

csmir:
Discovery of this bug was assisted by @MichailiK


MichailiK (approved):
For some context: This fixes a bug where the before/cached fields of the `UserUpdated` and `GuildMemberUpdated` events returned the after/new values of some properties (specificically, properties stored in the `GlobalUser`).

PR #2078, opened 2022-02-03
Patch; Webhook regex doesn't respect case-sensitive url's

### LINKED ISSUE #2074: Couldn't parse Discord Canary webhook URLs
Relevant stack trace :
```
System.ArgumentException: The given webhook Url was not in a valid format. The webhook Id could not be parsed. (Parameter 'webhookUrl')
   at Discord.Webhook.DiscordWebhookClient.ParseWebhookUrl(String webhookUrl, UInt64& webhookId, String& webhookToken)
   at Discord.Webhook.DiscordWebhookClient..ctor(String webhookUrl, DiscordRestConfig config)
```
The webhook URL in question was `https://canary.discord.com/api/webhooks/937967020320899092/BWKbVhNqu5shH9t96qWQKlzYisoHmM7PIMv88-IokTdR-lwHX7fNeWKk_Y2VBtSJpD2b`
image
(I made sure to invalidate the webhook URL for security purposes, it's fine to pass this around).

csmir:
Fixes #2074.

### DIFF
--- a/src/Discord.Net.Webhook/DiscordWebhookClient.cs
+++ b/src/Discord.Net.Webhook/DiscordWebhookClient.cs
@@ -33,7 +33,7 @@ namespace Discord.Webhook
             : this(webhookUrl, new DiscordRestConfig()) { }
 
         // regex pattern to match webhook urls
-        private static Regex WebhookUrlRegex = new Regex(@"^.*(discord|discordapp)\.com\/api\/webhooks\/([\d]+)\/([a-z0-9_-]+)$", RegexOptions.Compiled | RegexOptions.IgnoreCase | RegexOptions.CultureInvariant);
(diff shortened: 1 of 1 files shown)


MichailiK (changes requested):
Because `RegexOptions.IgnoreCase` is specified, this does not change anything.

csmir:
Ah, I probably missed that. It seems to match the string with or without canary in Regex101, so other causes do not make sense to me?

csmir:
image
As implied here, the webhook ID seems to match group 2 with index 3. This seems executed quite well. The canary case is entirely ignored with or without this present.

image

image

Error thrown in issue 2074 implies that the ulong is mismatched. Here it implies that that would be impossible.

csmir:
Upon testing now, it appears that ptb & canary work fine regularly. Quite a waste of effort to check internally before testing.

Issue #1248, opened 2019-01-28
v3: acceptable breaking changes


foxbot:
This issue serves as a discussion hub for what types of changes we would like to make in v3.

Chris-Johnston:
From #1247, [The addition of the `IPresence#ActiveClients` `get` property, since this is information that is included as part of the presence update payload.](https://github.com/RogueException/Discord.Net/pull/1247/files#diff-9a85adbcccffdf1a55c6f22b2070480bR22)

Kyle-Undefined:
An Activity overhaul would be nice, with the way Discord is going, it doesn't make sense to keep calling Spotify, Visual Studio, etc Integrations as "Games".

AntiTcb:
#974 in a nutshell. The separation between the sync methods on the socket entities and the async methods on the interfaces are cumbersome to explain to newcomers, and there are quite a couple of implementations where the `CacheMode` parameter doesn't actually do anything, most notably in explicit interface implementations that just map to the cache.

Chris-Johnston:
Not necessarily a breaking change to the public API, but thoughts on splitting up existing tests into a set of unit tests and integration tests, so that we can at least run some tests in CI on PRs?

Yucked:
Move TokenType and Token to config class instead of LoginAsync. And, perhaps replace AsyncEvents with Func<T>?

Joe4evr:
> Not necessarily a breaking change to the public API, but thoughts on splitting up existing tests into a set of unit tests and integration tests, so that we can at least run some tests in CI on PRs?

As an aside to this: I had toyed with the idea of providing a TestSuite library that could effectively mock the Discord API surface locally so users can also make tests and debug their routines much easier. I can't count the number of times I could prevent a bug in my addons if that were available.

FiniteReality:
I'd say fixing a whole bunch of issues with the command service is in order. There are a *lot* of pitfalls and issues with it, so much so that some of us have written our own.

I think a good start would be to open up another issue for the commands to ask what people would like most. However, I think a good base design would be something like [Finite.Commands](https://gitlab.com/FiniteReality/Finite.Commands) considering it contains the work of a lot of our internal design discussions, but making it somewhat more tied to Discord.Net and pointing out other options if people require it.

Chris-Johnston:
Since DNet has moved to an org now, do you think that commands should still be part of the Discord.Net repository, or that they should have their own repository? (and does this matter?)

[...]

PR #2079, opened 2022-02-03
Feature: Add IEnumerable collection parameters instead of arrays

quinchs:
## Summary
As discussed in https://github.com/discord-net/Discord.Net/issues/1248#issuecomment-619884488, having a `IEnumerable<>` parameter allows more flexibility on the developers end when using this extension.

This shouldn't be a breaking change as any array parameters will work for `IEnumerable`.


csmir:
More needing to be pushed to this? As far as I can see there are no further changes to be made.

PR #2109, opened 2022-02-14
Fixes unused creation of REST clients for DiscordShardedClient shards.

### LINKED ISSUE #2086: Sharding creates multiple instances of DiscordRestClient while only using one.
The ` DiscordShardedClient ` handles REST over a single ` DiscordRestClient `, which makes perfect sense. However, each shard creates its own rest client, while the .Rest property of the sharded client defaults to the first shard. A better approach to create rest clients should be implemented, excess types being instanced like this seems like a waste of resources.

csmir:
Fixes #2086


csmir:
No instance is populated pre-setting the `Shards` value? If this is so, then all we have to do is check if there are any shards, as _shards and thus Shards will not contain anything before the first shard has been initialized

csmir:
Check for parentclient seems to be better, as it is null if no earlier shard exists and is sorted by index, not by shardId.

quinchs (approved):

PR #2130, opened 2022-02-23
Optimising code

DenVot:
I've removed redundant using-s and optimised some loops


csmir (commented):
Swapping out regular loops for linq has negative effects on performance and should not be done if not required. Other changes unnecessary influence code other people made without contributing to any change.

csmir (commented):
This seems like a very low effort way to get contributor status on a project, I do not approve of this PR due to reason included above, as well as the redundancy of this commit.

PR #2131, opened 2022-02-24
Support Sending Message Flags

CottageDwellingCat:
This pr add support for sending flags on message creation, according to discord's docs only the `SuppressEmbeds` flag is supported

see discord-net-labs/discord.net-labs#437


quinchs:
This change will require all 3rd party libs to update if they reference the `SendMessage` function, I'm going to hold this one off until a minor change

PR #2137, opened 2022-02-25
Add self_video to VoiceState

JoshCollis:
---
I've added the 'self_video' property to VoiceState.  This is included in the discord API but wasn't included in the library.  It is used to detect if a user has their webcam enabled


src/Discord.Net.WebSocket/Entities/Users/SocketVoiceState.cs
```diff
@@ -25,6 +25,7 @@ private enum Flags : byte
             SelfMuted = 0x08,
             SelfDeafened = 0x10,
             SelfStream = 0x20,
+            SelfVideo = 0x30,
```
quinchs: Should be 0x40 for flags

quinchs (changes requested):

quinchs (approved):

Issue #2150, opened 2022-03-02
Add Integration Create, Update and Delete events


Misha-133:
https://discord.com/developers/docs/topics/gateway#integration-create

right now it only spams  `Unknown Dispatch (INTEGRATION_UPDATE)` in logs

FeroxFoxxo:
Should be closed? This issue was addressed in #1755
Are you sure you're updated to the latest version of DNet?

Misha-133:
> Should be closed, check https://github.com/discord-net/Discord.Net/issues/1755

That's the thing - it's not implemented yet

FeroxFoxxo:
How odd, let me have a look into it rq.

FeroxFoxxo:
You're completely right, I'm not sure why those threads were marked as resolved. I'll quickly write a fix and pr it!

FeroxFoxxo:
Fixed in #2168. Sorry it took so long! The integration codebase hadn't been changed in *years*, and was broken due to being outdated :P

PR #2168, opened 2022-03-05
 Fixed Integration model from GuildIntegration and added INTEGRATION gateway events

FeroxFoxxo:
This PR introduces the INTEGRATION_CREATED, INTEGRATION_UPDATED and INTEGRATION_DELETED events, fixing #2150.

It also fixes some rather big changes to the integration model, which I assume is due to a change in the object being sent by Discord? I noticed GuildIntegration referenced a lot in the code, rather than Integration.

Since the integration part of the API isn't used like, at all, I'm not sure if anyone would be affected by these changes, but nevertheless, they are breaking! Although looking at the old models I don't think they've even worked in the past few years..-

Some questions, before approval:

I can add caching of integrations, but as this is not sent to the client on the GUILD_CREATE event, we would have to run a second RESTful request on getting this information, say; if the integration intent is enabled. As such, this is a point that should probably be discussed further, and thus I'm currently returning a guild object, the id of the integration, and the optional id of the application on the INTEGRATION_DELETED event, as the related integration object isnt cached.

Secondly, I saw that Optional<ulong> threw an error for RoleId in the Integration model I had made, as Discord returns `null` for this in the JSON on `INTEGRATION_CREATION`. Is this a bug on our side, by means of a JSON type converter, or the API? The API does state this can be optional, hence why I used the related Optional<> type. The issue stems from reader.value being null on the ulong type converter that Optional runs. By any means, I have since changed this to ulong? in my push here.

Note: there are two separate types of integrations sent in the integration object: Discord and Twitch/Others. Discord integrations do not contain a big chunk of fields in Integration, hence why there are so many optional fields, as is specified per the Discord API docs.

Apparently the endpoints for creating, modifying and syncing integrations were removed aswell, with the only option left being to get and delete them. Please tell me if I'm wrong here though!

I also noticed the docs had stated the IntegrationAccount as returning a string for its ID, while we had it as an int. This is due to platforms like YouTube, whose support was added in 2020, not using ints to store IDs like Twitch does, and thus this code being untouched since 2016 has probably been the reason it has been broken.

There may be some bugs as this is my first large PR, so bear with me :P


src/Discord.Net.Core/Entities/Integrations/IIntegration.cs
```diff
@@ -3,15 +3,16 @@
 namespace Discord
 {
     /// <summary>
-    ///     Holds information for a guild integration feature.
+    ///     Holds information for an integration feature.
+    ///     Nullable fields not provided for Discord bot integrations, but are for Twitch etc.
```
quinchs, line 7: To prevent some headaches with nullables maybe we could make an interface system under this with the correct properties that are non null, ex:
```md
IIntegration
- IDiscordBotIntegration
- ITwitchIntegration
- etc.
```

src/Discord.Net.Rest/Entities/Integrations/RestIntegrationAccount.cs
```diff
@@ -0,0 +1,26 @@
+using Model = Discord.API.IntegrationAccount;
+
+namespace Discord.Rest
+{
+    public class RestIntegrationAccount : IIntegrationAccount
```
quinchs, line 8: This needs documentation

src/Discord.Net.Rest/Entities/Integrations/RestIntegrationApplication.cs
```diff
@@ -0,0 +1,36 @@
+using Model = Discord.API.IntegrationApplication;
+
+namespace Discord.Rest
+{
+    public class RestIntegrationApplication : RestEntity<ulong>, IIntegrationApplication
```
quinchs, line 8: This needs documentation

quinchs (changes requested):
> I can add caching of integrations, but as this is not sent to the client on the GUILD_CREATE event, we would have to run a second RESTful request on getting this information, say; if the integration intent is enabled. As such, this is a point that should probably be discussed further, and thus I'm currently returning a guild object, the id of the integration, and the optional id of the application on the INTEGRATION_DELETED event, as the related integration object isnt cached.

I personally think that if it isn't provided on the `GUILD_CREATE` payload then we shouldn't cache it due to state mismatch.

> Secondly, I saw that Optional threw an error for RoleId in the Integration model I had made, as Discord returns null for this in the JSON on INTEGRATION_CREATION. Is this a bug on our side, by means of a JSON type converter, or the API? The API does state this can be optional, hence why I used the related Optional<> type. The issue stems from reader.value being null on the ulong type converter that Optional runs. By any means, I have since changed this to ulong? in my push here.

`Optional<>` doesn't deal with nulls, its strictly for whether the value is in the json or not, I would try using and `Optional<ulong?>`.

src/Discord.Net.Rest/Entities/Integrations/RestIntegrationAccount.cs
FeroxFoxxo: Fixed!

src/Discord.Net.Rest/Entities/Integrations/RestIntegrationApplication.cs
FeroxFoxxo: Fixed!

PR #2203, opened 2022-03-21
Update intro.md

GreenScreen410:


OrionFOTL:
So they're supported on mobile now? Do you mind posting a screenshot, or linking to an announcement if there is any? It seems like there's no support on Discord 118.11 Stable on Android 11, or that's just me.

quinchs:
Seems like they're stull unsupported.

PR #2210, opened 2022-03-26
Add a way to remove type readers from the interaction/command service.

### LINKED ISSUE #2189: CommandService should have a RemoveTypeReader method
Could we get a way to remove custom type readers from the command/interaction services?

CommandService already has a method AddTypeReader but it doesn't have a matching RemoveTypeReader method. This method, if implemented, should also undo the default type reader overrides.

Thanks

quinchs:
## Summary
This PR implements methods to remove type readers from the interaction service as well as the command service.


quinchs:
@Cenngo waiting on your review

Cenngo (approved):

PR #2212, opened 2022-03-29
Add build overrides to Discord.Net

quinchs:
## Summary

This PR adds a way for library maintainers to create build overrides for dnet.

### Why is this useful?
When people encounter bugs that's hard or near impossible to debug locally or if we want to test out a patch before releasing it without having to go thru the hastle of getting people to either use myget or clone the repo and build we can use build overrides. They are much easier for the end developer to add to their codebase:
```cs
public async Task MainAsync()
{
  // hook into the log function
  BuildOverrides.Log += (buildOverride, message) => 
  {
    Console.WriteLine($"{buildOverride.Name}: {message}");
    return Task.CompletedTask;
  };

  // add your overrides
  await BuildOverrides.AddOverrideAsync("example-override-name");
}
```

### Wouldn't this be insecure/unsafe?

The current way I've set this up is that only library maintainers can create and distribute build overrides. To add a layer of transparency I've also setup a [repository](https://github.com/discord-net/Discord.Net.BuildOverrides) containing the override server code and the override code.

### Can overrides be used to token log my bot?

Theoretically yes. By default overrides don't receive client information ex: bot id, guilds, etc. But due to their nature they can override methods that expose this info, again overrides are solely created by library maintainers and are used strictly to fix or diagnose bugs.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "overrides/Discord.Net.BuildOverrides"]
+	path = overrides/Discord.Net.BuildOverrides
+	url = https://github.com/discord-net/Discord.Net.BuildOverrides
--- a/Discord.Net.sln
+++ b/Discord.Net.sln
@@ -34,7 +34,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "_InteractionFramework", "sa
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "_WebhookClient", "samples\WebhookClient\_WebhookClient.csproj", "{B61AAE66-15CC-40E4-873A-C23E697C3411}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "IDN", "samples\idn\idn.csproj", "{4A03840B-9EBE-47E3-89AB-E0914DF21AFB}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "idn", "samples\idn\idn.csproj", "{4A03840B-9EBE-47E3-89AB-E0914DF21AFB}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tests", "Tests", "{C7CF5621-7D36-433B-B337-5A2E3C101A71}"
 EndProject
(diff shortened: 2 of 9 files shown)


Emzi0767:
Ayyyy, finally the ability to inject arbitrary, unsigned, not attested, unverified code into random bots!

SylveonDeko:
> Ayyyy, finally the ability to inject arbitrary, unsigned, not attested, unverified code into random bots!

At the users own discretion

Emzi0767:
> > Ayyyy, finally the ability to inject arbitrary, unsigned, not attested, unverified code into random bots!
> 
> At the users own discretion 

And not always the intended user! Isn't it wonderful?

SylveonDeko:
We don't need passive aggressiveness here, nor is there any need to attack the lib for it, choose another repo, not this one

Emzi0767:
You obviously need someone to tell you that adding the ability to execute any arbitrary code *without any safeguards whatsoever* is a bad idea. It is in fact up there with making a logger that can interpret user input to load classes from the internet.

MrCakeSlayer:
@Emzi0767 Build Overrides are only created by library maintainers meaning random user's can't just inject code into your project. If your worried about maintainer's code then you shouldn't use the library to begin with as they could just inject said code into the core project to begin with.
@Sylveon76 this isn't the place for comments like that

Emzi0767:
> meaning random user's can't just inject code into your project

That is what log4j maintainers were thinking as well

MrCakeSlayer:
As is the thought with any project or software on the internet. If you are afraid to use build overrides, then simply don't enable any. They are purely there to help diagnose and patch bugs without a full release test. No one is forcing you to enable them.

quinchs:
> > meaning random user's can't just inject code into your project
> 
> That is what log4j maintainers were thinking as well

They're completely optional, you have to explicitly add them to your codebase. I've put in as much security measures into the server and system as I reasonably could

Emzi0767:
But you didn't. It's a straight download & load, nothing else. There's not even a shred of security thinking there. Your only layer of defence is trivially bypassed without modifying the assembly, and sometimes without ever touching the box at all.

quinchs:
> But you didn't. It's a straight download & load, nothing else. There's not even a shred of security thinking there. Your only layer of defence is trivially bypassed without modifying the assembly, and sometimes without ever touching the box at all.

How would you propose I improve the system?

Emzi0767:
Get rid of it. Seriously.

Emzi0767:
I apologize for my earlier rude behaviour. I had some time to calm down, and I also apologized to quin directly. I've also offered some ideas for improvements, mostly around the areas I commented on.

PR #2244, opened 2022-04-14
Add files to the parameters

Diego-VP20:
# Changes

Changed the order of two parameters (just to make it a bit prettier)
Added the parameter of type `IAttachment`.


csmir (approved):
lgtm

PR #2245, opened 2022-04-14
Add doc for Position field

Diego-VP20:
# Changes
Add docstring for the `Position` field.


quinchs:
image
The `Position` property has docs already from `IRole`

PR #2246, opened 2022-04-14
[DOCS] Group commands example 

Misha-133:
This PR adds an example with command and sub-command groups to interaction framework inro
It should explain how to use `Group` attribute

# Changes
- added codeblock with an example for interaction framework intro

image

### DIFF
--- a/docs/guides/int_framework/samples/intro/groupmodule.cs
+++ b/docs/guides/int_framework/samples/intro/groupmodule.cs
@@ -1,0 +1,15 @@
+[Group("group-name", "Group description")]
+public class CommandGroupModule : InteractionModuleBase<SocketInteractionContext>
+{
+    [SlashCommand("ping", "Get a pong")]
+    public async Task PongSubcommand()
+        => await RespondAsync("Pong!");
+
+    [Group("subcommand-group", "Subcommand group description")]
+    public class SubcommandGroupModule : InteractionModuleBase<SocketInteractionContext>
+    {
+        [SlashCommand("echo", "Echo an input")]
+        public async Task PongSubcommand(string input)
+            => await RespondAsync(input);
+    }
+}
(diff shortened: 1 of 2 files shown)


docs/guides/int_framework/samples/intro/groupmodule.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    [SlashCommand("ping", "Get a pong")]
+    public async Task PongSubcommand()
+        => await RespondAsync("Pong!");
+
+    [Group("subcommand-group", "Subcommand group description")]
+    public class SubcommandGroupModule : InteractionModuleBase<SocketInteractionContext>
+    {
+        [SlashCommand("echo", "Echo an input")]
+        public async Task PongSubcommand(string input)
```
csmir: I suggest naming this 'EchoSubcommand' instead for logic

csmir (changes requested):

docs/guides/int_framework/samples/intro/groupmodule.cs
Misha-133: Oops, didn't notice that i forgot to change the name ._.

docs/guides/int_framework/samples/intro/groupmodule.cs
```diff
@@ -0,0 +1,21 @@
+//You can put commands in groups
```
lhjt:
```suggestion
// You can put commands in groups
```

lhjt (changes requested):
Please add a space after the `//` in your comments

docs/guides/int_framework/samples/intro/groupmodule.cs
Misha-133: Fixed 🤔

lhjt (approved):

PR #2248, opened 2022-04-17
Fix SocketGuild not returning the AudioClient

LXGaming:
This PR fixes an issue where the SocketGuild always returns null for IGuild.AudioClient.


quinchs:
I wonder why that was never caught...

PR #2250, opened 2022-04-18
V4: Message builder

quinchs:
## Summary
This PR adds a new builder called `MessageBuilder` which can build a `Message` that is used in `SendMessageAsync` functions.
The upsides to this style is we can merge `SendFile` and `SendMessage` functions as the library can determine whether or not the message contains files.
The builder also makes building reusable message templates easier.

#### Todo

- [x] `IMessageChannel.SendMessageAsync`
- [ ] Interaction responses
- [ ] Module base methods


d4n3436:
There should be some abstraction layer for `MessageBuilder` and `Message` like interfaces, so people can create their own custom builders.

quinchs:
We could just make stuff protected and virtual?

csmir:
@quinchs thoughts on allowing message parameters to be provided with a string operator?

```cs
    public class Message
    {
        private Message(string input)
        {

        }
        
        public static implicit operator Message(string input)
            => new Message(input);
    }

    internal class Example
    {
        public void ExampleMethod()
        {
            Send("test");

            void Send(Message message)
            {

            }
        }
    }
```

csmir:
Also with the same design as above, an embed could serve as operator for message:

```cs
        private Message(Embed input)
        {

        }

        public static implicit operator Message(Embed embed)
            => new Message(embed);
```

csmir:
With the mention of 'format', another idea to consider is a `Format` method, much like `string.Format` which runs over all entries of the builder looking for `{}` format entries and `params object[] values` as part of the `Format` method in these places. If not enough values are provided, escape and leave other format entries sit, or allow a bool to clear other format notations out of the string.

quinchs:
@Rozen4334 lgtm

PR #2251, opened 2022-04-18
V4: State and Cache providers

### LINKED ISSUE #2238: Proposal: State and Cache providers
## Proposal

Memory usage is a big issue within the library, developers have minimal control on how entities are stored in state. The proposal is to allow developers to fully customize the way Discord.Net stores entities in memory.

Foxbot has the first part of this covered with the [`IStateProvider`](https://github.com/discord-net/Discord.Net/tree/0cf14da3dfcf0bc069dcc6d212dca130a71f34e1/src/Discord.Net/State) allowing users to control how entities are fetched-stored to and from the cache. The state provider allows for downloading entities if they don't exist within the cache for almost all gateway models and events.

The [`ICacheProvider`](https://github.com/Discord-Net-Labs/Discord.Net-Labs/issues/130) is something I've been thinking about for *way too long*. I've finalized my idea on implementing this into Discord.Net.
For each entity that *can* be cached ex `IGuildMember` there will be a `IGuildMemberModel` interface that allows us to have multiple ways to implement converting to and from strong-typed entities to models. We can make the api model `Discord.API.Member` and make it inherit the `IGuildMemberModel` interface and take in the interface for constructing a `SocketGuildMember` or `RestGuildMember` entity. Developers can create their own models that inherit `IGuildMemberModel` for example a class that converts itself to a protobuf buffer and pass that to the state provider to use to create an entity.

The term "state provider" and "cache provider" are two separate things, the state provider determines how entities are retrieved and stored while the cache provider stores to and fetches from the cache. Take this flow diagram as some more explanation behind this:

dnet-state-cache-flow

### Things to note
The state provider will have to return the interface types of entities ex `IGuild`, `IUser`, etc due to the ability to fetch from rest if the behavior is enabled, this will cause a lot of gateway events to have to use interfaces due to this.

quinchs:
## Summary
This PR implements state and cache providers as proposed in #2238.

The current api of the providers is up for debate, I've implemented it how I see fit but feedback and reviews are always welcome.

#### Todo:
- [x] API design
- [x] Internal reference cache
- [ ] Default providers
    - [x] Default concurrent providers
    - [ ] Redis cache provider
- [ ] Entities
    - [x] Users
    - [x] Presence
    - [ ] Channels
    - [ ] Guilds
    - [ ] Messages
    - [ ] Emotes


src/Discord.Net.Core/Cache/CacheableEntityExtensions.cs
```diff
@@ -0,0 +1,111 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Discord
+{
+    internal static class CacheableEntityExtensions
```
Cenngo, line 9: Is there any reason why these methods arent returning the generic type parameters? We could downcast it later.

Cenngo (approved):

src/Discord.Net.Core/Cache/CacheableEntityExtensions.cs
quinchs: Not really, its just whats there atm. it could change once I implement the rest of the entities

src/Discord.Net.WebSocket/DiscordSocketConfig.cs
```diff
@@ -25,6 +25,17 @@ namespace Discord.WebSocket
     /// </example>
     public class DiscordSocketConfig : DiscordRestConfig
     {
+        /// <summary>
+        ///     Gets or sets the cache provider to use
```
lhjt:
```suggestion
        ///     Gets or sets the cache provider to use.
```
Just for consistency.

src/Discord.Net.WebSocket/Entities/Users/SocketGuildUser.cs
```diff
@@ -6,29 +6,33 @@
...
+        private ulong _guildId;
 
-        internal override SocketGlobalUser GlobalUser { get; set; }
         /// <summary>
         ///     Gets the guild the user is in.
         /// </summary>
-        public SocketGuild Guild { get; }
+        public Lazy<SocketGuild> Guild { get; } // TODO: convert to LazyCached once guilds are cached.
+        /// <summary>
+        ///     Gets the guilds id that the user is in.
```
lhjt:
```suggestion
        ///     Gets the ID of the guild that the user is in.
```

lhjt (changes requested):
Just a few copyedits

MrCakeSlayer (approved):

AddressXception:
Is this still on the roadmap? I'm building something and using this library but running into a few limitations that I need to work around. I left my comments in this discussion here:

https://github.com/discord-net/Discord.Net/discussions/2508

thanks!

PR #2265, opened 2022-04-28
[Docs] Adding permission docs for interaction framework

csmir:
Also removes a random sln file that I noticed when switching views in vs


docs/guides/int_framework/permissions.md
```diff
@@ -0,0 +1,59 @@
...
+### Nesting permissions
+
+Alternatively, permissions can also be nested.
+It will look for all uses of `DefaultMemberPermissions` up until the highest level class.
+The `EnabledInDm` attribute can be defined at top level as well,
+and will be set up for all of the commands & nested modules inside this class.
+
+[!code-csharp[Permission stacking](samples/permissions/perm-nesting.cs)]
+
+The amount of nesting you can do is realistically endless, although really deep nesting is not adviced as the lookup will only take longer.
```
lhjt:
```suggestion
The amount of nesting you can do is realistically endless, although really deep nesting is not advised as the lookup will only take longer.
```

lhjt (changes requested):
Other than this, looks good.

lhjt (approved):

quinchs (approved):

docs/guides/int_framework/permissions.md
```diff
@@ -0,0 +1,59 @@
...
+### Nesting permissions
+
+Alternatively, permissions can also be nested.
+It will look for all uses of `DefaultMemberPermissions` up until the highest level class.
+The `EnabledInDm` attribute can be defined at top level as well,
+and will be set up for all of the commands & nested modules inside this class.
+
+[!code-csharp[Permission stacking](samples/permissions/perm-nesting.cs)]
+
+The amount of nesting you can do is realistically endless, although really deep nesting is not advised as the lookup will only take longer.
```
Cenngo: actually it doesnt, all modules are built at startup. any RegisterCommands*Async method call after modules are loaded to the interaction service will only perform a simple bitwise OR operation for that and that all.

Cenngo (changes requested):

Cenngo (approved):

PR #2270, opened 2022-04-29
Treat warnings as errors and set warning level to 5

quinchs:
## Summary
This PR sets the warning level to five and treats warnings as errors. The motivation behind this is to make our codebase more consistent and to help remove invalid summaries or small warnings.


joker-119:
Typo in summary "treats errors as warnings" should be the other way around

PR #2273, opened 2022-04-29
Fix: RequireOwnerPrecondition ignoring teams

exsersewo:
If an application is contained within a team the Owner prop on Application will be a team user thus the application owner id will not match the invokers id if they are a member of the team


quinchs:
see https://github.com/discord-net/Discord.Net/pull/1838

exsersewo:
Ah my bad, I didn't notice the proposal for `RequireTeam` instead.

PR #2289, opened 2022-05-12
Added UpdateAsync to SocketModal

serkanyazicioglu:
At the moment Modals cannot update initial interacted message. I added a UpdateAsync method similar to message components.

### DIFF
--- a/src/Discord.Net.WebSocket/Entities/Interaction/Modals/SocketModal.cs
+++ b/src/Discord.Net.WebSocket/Entities/Interaction/Modals/SocketModal.cs
@@ -174,6 +174,91 @@ namespace Discord.WebSocket
             HasResponded = true;
         }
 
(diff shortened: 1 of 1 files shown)


Almighty-Shogun:
Modals can also do this with:
```cs
await DeferAsync();
await ModifyOriginalResponse(x => {});
```

serkanyazicioglu:
Yeah checked a lot of combinations but missed this one I guess. Tried it now and it's working.

Not sure about making this non-standard. If this method will not be added at least your recommendation should be documented, it can be really helpful for others.

quinchs:
iirc modals were non-deferrable and non-updatable? Will check with a few people first.

quinchs:
apparently this is a feature so it lgtm

quinchs (approved):

PR #2293, opened 2022-05-14
Fix Application commands are disabled to everyone except admins by default

### LINKED ISSUE #2292: [Bug]: Application commands are disabled to everyone except admins by default
### Check The Docs

### Verify Issue Source

### Check your intents

### Description

On Discord.Net 3.6.0+, all application commands (except sub-commands) are being registered with `default_member_permissions` set to 0 by default. [According to Discord](https://discord.com/developers/docs/interactions/application-commands#application-command-object), setting this to 0 means the commands will be disabled to everyone except admins by default. This can be proven by requesting the app commands and inspecting the JSON response:

```json
[
    {
        "id": "972032022631567420",
        "application_id": "id",
        "version": "974181358614233090",
        "default_permission": true,
        "default_member_permissions": "0", // Message command, disabled to everyone except admins
        "type": 3, 
        "name": "Reverse Image Search",
        "description": "",
        "dm_permission": true
    },
    {
        "id": "972032022631567421",
        "application_id": "id",
        "version": "974181358614233091",
        "default_permission": true,
        "default_member_permissions": null, // Slash command group, enabled to everyone
        "type": 1, 
        "name": "img",
        "description": "Image search commands.",
        "dm_permission": true,
        "options": []
    },
    {
        "id": "972032022631567427",
        "application_id": "450865985545830401",
        "version": "974181358614233097",
        "default_permission": true,
        "default_member_permissions": "0", // Top-level slash command, disabled to everyone except admins
        "type": 1,
        "name": "help",
        "description": "Displays some help.",
        "dm_permission": true
    },
]
```
Downgrading to 3.5.0 fixes the issue.

### Version

3.6.1

### Working Version

3.5.0

### Logs

None

### Sample

None

### Packages

None

Cenngo:
Closes #2292.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,1 +1,3 @@
+github: quinchs
 open_collective: discordnet
+custom: https://paypal.me/quinchs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Changelog
 
+## [3.6.1] - 2022-04-30
+### Added
+- #2272 add 50080 Error code (503e720)
+
+### Fixed
+- #2267 Permissions v2 Invalid Operation Exception (a8f6075)
+- #2271 null user on interaction without bot scope (f2bb55e)
+- #2274 Implement fix for Custom Id Segments NRE (0d74c5c)
+
+### Misc
+- 3.6.0 (27226f0)
+
+
 ## [3.6.0] - 2022-04-28
 ### Added
 - #2136 Passing CustomId matches into contexts (4ce1801)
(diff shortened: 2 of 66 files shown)


Cenngo:
@d4n3436 feel free to test this for yourself. Now DefaultMemberPermission field is left as unspecified for cases where it used be 0 and the field is now actually returned as null by the backend.

d4n3436:
I tested it and it's working. I guess there won't be a way to explicitly set `default_member_permissions` to 0 since there's already  an Administrator permission, right?

Cenngo:
> I tested it and it's working. I guess there won't be a way to explicitly set `default_member_permissions` to 0 since there's already an Administrator permission, right?

yea i was thinking about exposing a way for setting the field to 0 but then i realized that wouldnt have much use because of GuildPermission.Administrator like you said

quinchs (approved):

PR #2305, opened 2022-05-16
(4.x) Tackle warnings, errors & messages in source at large scale

csmir:
This PR removes, modifies & introduces a huge amount of code refactoring to the DNet code base. This PR does not go paired with our current targets for out-of-support frameworks. 

## Fixes

Removes all warnings and a large amount of messages in the codebase in respect to code conventions. Also resolves some very minor errors in the codebase with assignment in some locations.

## Breaking

`RequestOptions` has been removed from `LoginAsync` as it doesn't use the parameter at all.
`RequestOptions` have been removed from `RestInteraction` serializing methods as they don't use it at all.

## Internal

There is no reference to `ClientBucketType` anywhere, it has been removed from the `RestApiClient`.
Changes explicit variables to implicit where it is clear what the type the variable presents.
Swaps `SubString` for range operator & `AsSpan`.

### DIFF
--- a/samples/InteractionFramework/InteractionHandler.cs
+++ b/samples/InteractionFramework/InteractionHandler.cs
@@ -37,7 +37,10 @@ namespace InteractionFramework
         }
 
         private async Task LogAsync(LogMessage log)
-            => Console.WriteLine(log);
+        {
+            Console.WriteLine(log);
+            await Task.CompletedTask;
+        }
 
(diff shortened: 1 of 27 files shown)


MrCakeSlayer (approved):

csmir:
Closing as state of PR is stale & I do not plan to follow up in this approach.

PR #2329, opened 2022-05-26
Fixes voice perms not retaining text perms.

csmir:


quinchs:
moving the case in the switch before text seems to work and is a bit cleaner

quinchs (approved):

PR #2336, opened 2022-05-28
Add support for attachments on interaction response type 7

d4n3436:
This PR adds support for modifying attachments on interaction response type 7 (`SocketMessageComponent.UpdateAsync`).

### DIFF
--- a/src/Discord.Net.WebSocket/Entities/Interaction/MessageComponents/SocketMessageComponent.cs
+++ b/src/Discord.Net.WebSocket/Entities/Interaction/MessageComponents/SocketMessageComponent.cs
@@ -261,20 +261,41 @@ namespace Discord.WebSocket
                 }
             }
 
-            var response = new API.InteractionResponse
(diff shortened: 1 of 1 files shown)


csmir:
Could you also cover this for rest message components?

d4n3436:
> Could you also cover this for rest message components?

That would require a rewrite of REST interactions due to the way they are designed.

csmir:
Could you clarify? The Update() method does the exact same as UpdateAsync() except that it does not directly hook into the response callback like socket interactions. There is no need to make any edits other than the action itself

d4n3436:
> The Update() method does the exact same as UpdateAsync() except that it does not directly hook into the response callback like socket interactions.

What does that mean? Update() returns a JSON string that is used to respond to the HTTP request and it can't support files because `UploadInteractionFileParams` is not meant to be serialized to a JSON string (unlike `InteractionResponse`).

quinchs (approved):

PR #2338, opened 2022-05-30
Make AddOptions more user friendly

CottageDwellingCat:
- You can now add 0 options without an exception, this is helpful when adding user-generated commands that may have no options.
- You can now call AddOptions on an option, this is helpful when adding multiple subcommands at once from a user-generated source.


Almighty-Shogun:
Would also be cool to make options disabled. Idk if it is already added since I can't find anything related to it

CottageDwellingCat:
That does not appear supported by discord (I checked the [documentation on command options](https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-option-structure)) you might want to create a feature request with them [here](https://github.com/discord/discord-api-docs/discussions/new). This pr changes the AddOptions() method on command builders to make it easier to use.

quinchs (approved):

PR #2345, opened 2022-06-03
Fix a bug in EmbedBuilder.Length when there is an EmbedField with no Value

Hampo:
Fixes a bug where 'EmbedBuilder.Length' will throw an exception of type 'System.NullReferenceException' when a field doesn't have a value.


csmir (approved):

src/Discord.Net.Core/Entities/Messages/EmbedBuilder.cs
```diff
@@ -150,7 +150,7 @@ public int Length
                 int authorLength = Author?.Name?.Length ?? 0;
                 int descriptionLength = Description?.Length ?? 0;
                 int footerLength = Footer?.Text?.Length ?? 0;
-                int fieldSum = Fields.Sum(f => f.Name.Length + f.Value.ToString().Length);
+                int fieldSum = Fields.Sum(f => f.Name.Length + (f.Value == null ? 0 : f.Value.ToString()?.Length ?? 0));
```
Cenngo: the initial null check seems redundant. `f.Value?.ToString().Length ?? 0` should be enough. Can you please verify whether that works or not and update your code if so.

Cenngo (changes requested):

csmir:
Could you please continue this PR and take up @Cenngo 's review?

Cenngo (approved):

quinchs (approved):

PR #2347, opened 2022-06-03
[Feature] Embed comparison

Misha-133:
### Changes:
Added `GetHashCode()`, `Equals()`, `==` operator and `!=` operator to following classes:
- `Embed` 
- `EmbedFooter`
- `EmbedAuthor`
- `EmbedImage`
- `EmbedThumbnail`
- `EmbedField`
- `EmbedBuilder`
- `EmbedVideo`
- `EmbedFooterBuilder`
- `EmbedAuthorBuilder`
- `EmbedFieldBuilder`
- `EmbedProvider`

### DIFF
--- a/src/Discord.Net.Core/Entities/Messages/Embed.cs
+++ b/src/Discord.Net.Core/Entities/Messages/Embed.cs
@@ -119,7 +119,7 @@ namespace Discord
         /// <param name="embed">The <see cref="Embed"/> to compare with the current <see cref="Embed"/></param>
         /// <returns></returns>
         public bool Equals(Embed embed)
-            => GetHashCode() == embed.GetHashCode();
+            => GetHashCode() == embed?.GetHashCode();
 
         /// <inheritdoc />
         public override int GetHashCode()
--- a/src/Discord.Net.Core/Entities/Messages/Embed.cs
+++ b/src/Discord.Net.Core/Entities/Messages/Embed.cs
@@ -102,12 +102,26 @@ namespace Discord
         public static bool operator !=(Embed left, Embed right)
             => !(left == right);
 
+        /// <summary>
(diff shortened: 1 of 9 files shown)


quinchs (approved):

d4n3436:
`GetHashCode` shouldn't be overriden on mutable types. The hash will change every time the object is mutated, which will break code that relies on it (like dictionaries or hashsets).

Another problem I see is that an anonymous object is created every time a hash code is computed, creating unnecesary memory allocations. This can be fixed by using `HashCode.Combine` or tuples instead.

Misha-133:
> `GetHashCode` shouldn't be overriden on mutable types. The hash will change every time the object is mutated, which will break code that relies on it (like dictionaries or hashsets).
> 
> Another problem I see is that an anonymous object is created every time a hash code is computed, creating unnecesary memory allocations. This can be fixed by using `HashCode.Combine` or tuples instead.


I will look into first issue you've mentioned when get home

But the second one - first approach I tried used `System.HashCode` which did not compile for .NET Framework 4.6.1 & .NET Standard 2.0
There is nuget package `Microsoft.Bcl.HashCode` which brings support for `HashCode` in .NET Standard 2.0 but I didn't find any way to get it working in framework 4.6.1 so I had to use anonymous objects.

d4n3436:
> But the second one - first approach I tried used `System.HashCode` which did not compile for .NET Framework 4.6.1 & .NET Standard 2.0

Use tuples.

Misha-133:
Issues should be solved now

csmir:
Adding xmldoc addressing what is being compared would be desirable.

csmir (approved):

MrCakeSlayer (approved):

PR #2401, opened 2022-07-31
Update intro.md - Include the part where the handler is created.

MartenM:
Updated intro.md to include the part where you need to setup the interactions handler.


csmir:
Theres a lot of things in this PR that need to be reorganized. The code snippet needs to be delegated to a separate file and the location of your paragraph is in the wrong location by a couple 100 lines. I would like to review this PR, but there are quite a lot of things to go through, so I will just quickly walk over this in another iteration of the idea.

PR #2410, opened 2022-08-01
small docs fix (components -> component) to match API

Aida-Enna:
Just what it says on the tin 👍


csmir:
`components` is the API default. This PR makes no sense.

Aida-Enna:
Ah, I was looking at a project with a 2.X version of the API, where it used ``component`` instead of ``components``. My mistake.

PR #2421, opened 2022-08-07
Fix placeholder length being hardcoded

gepbird:
When the developer assigns a string to a `TextInputBuilder.Placeholder`, its length is checked against a hardcoded value. Extracting it out to a public variable makes it more consistent with the codebase and the developer can check if the placeholder is too long without having to hardcode it.

### DIFF
--- a/src/Discord.Net.Core/Entities/Interactions/MessageComponents/ComponentBuilder.cs
+++ b/src/Discord.Net.Core/Entities/Interactions/MessageComponents/ComponentBuilder.cs
@@ -1198,6 +1198,7 @@ namespace Discord
 
     public class TextInputBuilder
     {
+        public const int MaxPlaceholderLength = 100;
         public const int LargestMaxLength = 4000;
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


SylveonDeko:
..... It's done this way because it's a discord restriction

gepbird:
> ..... It's done this way because it's a discord restriction

@Pusheon 

I don't understand that. 
Discord restricts placeholder length to maximum 100 characters, what does it have to do with not extracting it to a variable?
In [`SelectMenuBuilder`](https://github.com/discord-net/Discord.Net/blob/a9116d658ba3314bd0930f576e312f59969b8a4f/src/Discord.Net.Core/Entities/Interactions/MessageComponents/ComponentBuilder.cs#L734) it is extracted.

csmir (approved):

MrCakeSlayer (approved):

MrCakeSlayer:
@Pusheon it doesn't matter if it's a Discord limit or a DNet limit. the purpose of this PR is the make the hardcoded value accessible to developers so they can check input lengths on their side without also hardcoding the value in their code (which would cause issues if it ever changed in DNet).

PR #2436, opened 2022-08-22
ForumChannel GetActiveThreads issue and more changes

viktorchernikov:
@Ori#9999 from Discord.Net discord server figured out that when you try to get all active threads of forum channel, instead of returning only threads inside this channel, the method returns all guild's threads.

All methods responsible for getting thread channels were removed from Forum Channels and instead were moved to Guild Channel. These methods will return only threads of needed channel.

Original method that allows to get all threads in guild still exists.


quinchs:
I don't think moving the `GetThreads` methods to the `IGuildChannel` is a good idea as `IGuildChannel` represent *any* guild channel, not all channels have thread support.

PR #2437, opened 2022-08-23
adding scheduled events to audit log

SaculRennorb:
I'll be frank, I just made this because i needed the `user` and `affected_id` for ScheduledEvent updates, as the corresponding event does not provide a field of who updated the ScheduledEvent. 

Was sad to see that apparently noone added auditLog ids 100-102 (the ones related to ScheduledEvents) and there does not seen to be a generic way of just getting an arbitrary field from those log entries (can't even really do reflection magic because of how the data is set up).

This isn't tested very well and might also still contain some copy paste problems, as well as problems related to me not knowing the inner workings of discord.net, also I didn't bother too much with getting used to style / structures of discord.net. Feel free to reject this if you don't want to put in work, as neither do I. It seems to work, but also I just wanted to have the ids.

### DIFF
--- a/src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventUpdateAuditLogData.cs
+++ b/src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventUpdateAuditLogData.cs
@@ -87,7 +87,13 @@ namespace Discord.Rest
         ///     Gets the snowflake id of the event.
         /// </summary>
         public ulong Id { get; }
+        /// <summary>
+        ///     Gets the state before the change.
+        /// </summary>
         public ScheduledEventInfo Before { get; }
+        /// <summary>
+        ///     Gets the state after the change.
+        /// </summary>
         public ScheduledEventInfo After { get; }
     }
 }
--- a/src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventUpdateAuditLogData.cs
+++ b/src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventUpdateAuditLogData.cs
@@ -1,0 +1,93 @@
+using System;
+using System.Linq;
+using Discord.API;
+
+using Model = Discord.API.AuditLog;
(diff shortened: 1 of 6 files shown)


src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventUpdateAuditLogData.cs
```diff
@@ -0,0 +1,93 @@
...
+            return new ScheduledEventUpdateAuditLogData(id, before, after);
+        }
+
+        // Doc Note: Corresponds to the *current* data
+
+        /// <summary>
+        ///     Gets the snowflake id of the event.
+        /// </summary>
+        public ulong Id { get; }
+        public ScheduledEventInfo Before { get; }
```
quinchs, line 93: Missing docnote

src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventUpdateAuditLogData.cs
```diff
@@ -0,0 +1,93 @@
...
+        public ScheduledEventInfo After { get; }
```
quinchs, line 97: Missing docnote

quinchs (changes requested):

src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventUpdateAuditLogData.cs
SaculRennorb: added a doc text

src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventUpdateAuditLogData.cs
SaculRennorb: added a doc text

quinchs (approved):

quinchs:
@SaculRennorb please update from dev

PR #2453, opened 2022-09-03
Null check command localizations

### LINKED ISSUE #2450: [Bug]: Slash Command Builder doc throws exception
### Check The Docs

### Verify Issue Source

### Check your intents

### Description

Following the hello world doc [here ](https://github.com/discord-net/Discord.Net/blob/370bdfa3c630152bfcd851c918f5d49ab6a3116d/docs/guides/int_basics/application-commands/slash-commands/responding-to-slash-commands.md)for creating a slash command via dotnet results in an exception being thrown for guildCommand.Build().

FWIW others in that discord are experiencing the same. 

### Version

3.8.0

### Working Version

_No response_

### Logs

System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=Discord.Net.Core
  StackTrace:
   at Discord.ApplicationCommandProperties.set_NameLocalizations(IReadOnlyDictionary`2 value)
   at Discord.SlashCommandBuilder.Build()
   at VRisingBot.Services.CommandHandler.<Client_Ready>d__7.MoveNext() in C:\Dev\VRisingBot\VRisingBot\VRisingBot\Services\CommandHandler.cs:line 84

### Sample

```cs
 public class CommandHandler
    {
        private readonly DiscordSocketClient client;
        private readonly CommandService commands;
        private readonly IServiceProvider services;

        public CommandHandler(DiscordSocketClient client, CommandService commands, IServiceProvider services)
        {
            this.commands = commands;
            this.services = services;
            this.client = client;
        }

        public async Task InstallCommandsAsync()
        {
            client.Ready += Client_Ready;
            client.SlashCommandExecuted += SlashCommandHandler;
        }    

        private async Task SlashCommandHandler(SocketSlashCommand command)
        {
            await command.RespondAsync($"You executed {command.Data.Name}");
        }

        public async Task Client_Ready()
        {
            // Let's build a guild command! We're going to need a guild so lets just put that in a variable.
            var guild = client.GetGuild(1013836369673597091);

            // Next, lets create our slash command builder. This is like the embed builder but for slash commands.
            var guildCommand = new SlashCommandBuilder();
            // Note: Names have to be all lowercase and match the regular expression ^[\w-]{3,32}$
            guildCommand.WithName("first-command");

            // Descriptions can have a max length of 100.
            guildCommand.WithDescription("This is my first guild slash command!");

            // Let's do our global command
            var globalCommand = new SlashCommandBuilder();
            globalCommand.WithName("first-global-command");
            globalCommand.WithDescription("This is my first global slash command");

            try
            {
                
                // Now that we have our builder, we can call the CreateApplicationCommandAsync method to make our slash command.
                await guild.CreateApplicationCommandAsync(guildCommand.Build());

                // With global commands we don't need the guild.
                await client.CreateGlobalApplicationCommandAsync(globalCommand.Build());
                // Using the ready event is a simple implementation for the sake of the example. Suitable for testing and development.
                // For a production bot, it is recommended to only run the CreateGlobalApplicationCommandAsync() once for each command.
            }
            catch (Exception exception)
            {
                // If our command was invalid, we should catch an ApplicationCommandException. This exception contains the path of the error as well as the error message. You can serialize the Error field in the exception to get a visual of where your error is.
                var json = JsonConvert.SerializeObject(exception.Message, Formatting.Indented);

                // You can send this error somewhere or just print it to the console, for this example we're just going to print it.
                Console.WriteLine(json);
            }
        }
    }
```

### Packages

Azure.Identity
Azure.REsourceManager
Azure.ResourceManager.ContainerInstance
Microsoft.Extensions.DepdencyInjection

essoperagma:
Fixes https://github.com/discord-net/Discord.Net/issues/2450

Localization dictionaries are not checked against null and throw exception when building commands.
This PR adds necessary checks so that exceptions are avoided.
Includes a simple unit test to prevent regressions.

### DIFF
--- a/src/Discord.Net.Core/Entities/Interactions/ApplicationCommandOption.cs
+++ b/src/Discord.Net.Core/Entities/Interactions/ApplicationCommandOption.cs
@@ -106,13 +106,17 @@ namespace Discord
             get => _nameLocalizations;
             set
             {
-                foreach (var (locale, name) in value)
+                if (value != null)
                 {
-                    if(!Regex.IsMatch(locale, @"^\w{2}(?:-\w{2})?$"))
-                        throw new ArgumentException($"Invalid locale: {locale}", nameof(locale));
+                    foreach (var (locale, name) in value)
+                    {
+                        if(!Regex.IsMatch(locale, @"^\w{2}(?:-\w{2})?$"))
+                            throw new ArgumentException($"Invalid locale: {locale}", nameof(locale));
 
-                    EnsureValidOptionName(name);
+                        EnsureValidOptionName(name);
+                    }
                 }
+
                 _nameLocalizations = value;
             }
         }
@@ -126,13 +130,17 @@ namespace Discord
             get => _descriptionLocalizations;
             set
             {
-                foreach (var (locale, description) in value)
+                if (value != null)
                 {
-                    if(!Regex.IsMatch(locale, @"^\w{2}(?:-\w{2})?$"))
-                        throw new ArgumentException($"Invalid locale: {locale}", nameof(locale));
+                    foreach (var (locale, description) in value)
+                    {
+                        if(!Regex.IsMatch(locale, @"^\w{2}(?:-\w{2})?$"))
+                            throw new ArgumentException($"Invalid locale: {locale}", nameof(locale));
 
-                    EnsureValidOptionDescription(description);
+                        EnsureValidOptionDescription(description);
+                    }
                 }
+
                 _descriptionLocalizations = value;
             }
         }
--- a/src/Discord.Net.Core/Entities/Interactions/ApplicationCommandOptionChoice.cs
+++ b/src/Discord.Net.Core/Entities/Interactions/ApplicationCommandOptionChoice.cs
@@ -55,18 +55,21 @@ namespace Discord
             get => _nameLocalizations;
             set
             {
-                foreach (var (locale, name) in value)
+                if (value != null)
                 {
-                    if (!Regex.IsMatch(locale, @"^\w{2}(?:-\w{2})?$"))
-                        throw new ArgumentException("Key values of the dictionary must be valid language codes.");
-
-                    switch (name.Length)
+                    foreach (var (locale, name) in value)
                     {
-                        case > 100:
-                            throw new ArgumentOutOfRangeException(nameof(value),
-                                "Name length must be less than or equal to 100.");
-                        case 0:
-                            throw new ArgumentOutOfRangeException(nameof(value), "Name length must at least 1.");
+                        if (!Regex.IsMatch(locale, @"^\w{2}(?:-\w{2})?$"))
+                            throw new ArgumentException("Key values of the dictionary must be valid language codes.");
+
+                        switch (name.Length)
+                        {
+                            case > 100:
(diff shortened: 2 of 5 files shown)


quinchs (approved):

Cenngo (approved):

Cenngo:
Now that i got the chance to better investigate the #2450, i think it can be fixed with fewer additions/deletions. Your unit test would be a nice addition to the library and nice catch on SlashCommandBuilder#L913 but i think the following fix would be a more elegant solution. Could you test this out and let me know?
image

essoperagma:
Sounds reasonable!
Maybe we should also consider enabling nullable reference types in projects to help contributors be more conscious about what can and cannot be null. Though I agree that it maybe annoying when working with models where everything is OK to be null.

metalglove:
This is quite a major bug. 
The `SlashCommandBuilder` class has literally become unusable. When will this fix be merged and a new version be published?

essoperagma:
@Cenngo Now I sat down to get this done, but noticed that your method will leave the properties vulnerable.

That would be somewhat OK if the setters are inaccessible (cannot be called by the consumers of the library). But the types are public and so are the property setters. Which means that a consumers will get an exception if they runs this:

```csharp
new ApplicationCommandOptionProperties()
{
    NameLocalizations = null
}
```

So, my recommendation will be to keep this as is.
Let me know if you think otherwise 🙂

Cenngo:
true. I thought props classes were immutable. @quinchs any particular reason why built ApplicationCommand classes feture public setters unlike built MessageComponents, Embeds for instance?

akaJuliaan:
@metalglove as long as this PR is not merged you can avoid the error by doing something like this:
```cs
command.WithNameLocalizations(new Dictionary<string, string>());
command.WithDescriptionLocalizations(new Dictionary<string, string>());
```

metalglove:
> @metalglove as long as this PR is not merged you can avoid the error by doing something like this:
> 
> ```cs
> 
> command.WithNameLocalizations(new Dictionary<string, string>());
> 
> command.WithDescriptionLocalizations(new Dictionary<string, string>());
> 
> ```

That doesn't work since the null reference error still occurs when it tries to iterate over the backing dictionary which is null.

At L#913 as @Cenngo corrected in a screenshot above.

essoperagma:
@quinchs @Cenngo In case you are waiting for me, I don't have merge rights. Could one of you get this merged?

quinchs:
>  any particular reason why built ApplicationCommand classes feture public setters unlike built MessageComponents, Embeds for instance?
@Cenngo That doesn't seem right... They shouldn't?

quinchs:
also @Cenngo this ones yours to merge etc

Cenngo:
> any particular reason why built ApplicationCommand classes feture public setters unlike built MessageComponents, Embeds for instance?
> @Cenngo That doesn't seem right... They shouldn't?

All of them have internal ctors but class properties are still publicly settable which doesnt fit into the builder pattern, I agree with you on that. Looking at the current state of 3.8.0 i think we shouldnt introduce a breaking change like that atm. I can look more into fixing the builder pattern of app commands later but in the mean time Im merging this so we can at least realease a patch.

PR #2483, opened 2022-09-26
[Bugfix] fixed an NRE when event was changed from in channel to external or vice versa

SaculRennorb:
This fixes a nre from my earlier contribution that occurred when an scheduled-event update audit log entry contained a change that changed the event type from internal to exernal. I forgot to handle the fact that channel ids and location names can actually have a null value (only one of them can actually have a non null value at the same time). Also there now is a way to differentiate between those values not having changed (the property is null in that case) and the value changing from/to a null value (the optional will be unset / the id will be 0).


src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventInfo.cs
```diff
@@ -50,7 +50,7 @@ public class ScheduledEventInfo
         /// <summary>
         ///     Gets the metadata for the entity associated with the event.
         /// </summary>
-        public string Location { get; }
+        public Optional<string>? Location { get; }
```
Rekkonnect: Optional should not be nullable

src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventInfo.cs
```diff
@@ -60,7 +60,7 @@ public class ScheduledEventInfo
         /// </summary>
         public string Image { get; }
 
-        internal ScheduledEventInfo(ulong? guildId, ulong? channelId, string name, string description, DateTimeOffset? scheduledStartTime, DateTimeOffset? scheduledEndTime, GuildScheduledEventPrivacyLevel? privacyLevel, GuildScheduledEventStatus? status, GuildScheduledEventType? entityType, ulong? entityId, string location, int? userCount, string image)
+        internal ScheduledEventInfo(ulong? guildId, ulong? channelId, string name, string description, DateTimeOffset? scheduledStartTime, DateTimeOffset? scheduledEndTime, GuildScheduledEventPrivacyLevel? privacyLevel, GuildScheduledEventStatus? status, GuildScheduledEventType? entityType, ulong? entityId, Optional<string>? location, int? userCount, string image)
```
Rekkonnect: Same here

src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventUpdateAuditLogData.cs
```diff
@@ -35,13 +36,13 @@ internal static ScheduledEventUpdateAuditLogData Create(BaseDiscordClient discor
...
             var entityId = entry.Changes.FirstOrDefault(x => x.ChangedProperty == "entity_id");
-            var entityMetadata = entry.Changes.FirstOrDefault(x => x.ChangedProperty == "entity_metadata");
+            var location = entry.Changes.FirstOrDefault(x => x.ChangedProperty == "location");
             var userCount = entry.Changes.FirstOrDefault(x => x.ChangedProperty == "user_count");
             var image = entry.Changes.FirstOrDefault(x => x.ChangedProperty == "image");
 
             var before = new ScheduledEventInfo(
                 guildId?.OldValue.ToObject<ulong>(discord.ApiClient.Serializer),
-                channelId?.OldValue.ToObject<ulong?>(discord.ApiClient.Serializer),
+                channelId == null ? null : channelId.OldValue?.ToObject<ulong?>(discord.ApiClient.Serializer) ?? 0,
```
Rekkonnect, line 43: Ugly expression, should be similarly written to the code below

src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventUpdateAuditLogData.cs
```diff
@@ -51,16 +52,15 @@ internal static ScheduledEventUpdateAuditLogData Create(BaseDiscordClient discor
                 status?.OldValue.ToObject<GuildScheduledEventStatus>(discord.ApiClient.Serializer),
                 entityType?.OldValue.ToObject<GuildScheduledEventType>(discord.ApiClient.Serializer),
                 entityId?.OldValue.ToObject<ulong?>(discord.ApiClient.Serializer),
-                entityMetadata?.OldValue.ToObject<GuildScheduledEventEntityMetadata>(discord.ApiClient.Serializer)
-                    ?.Location.GetValueOrDefault(),
+                location == null ? null : location.OldValue?.ToObject<string>(discord.ApiClient.Serializer) ?? Optional<string>.Unspecified,
```
Rekkonnect: Same here

src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventUpdateAuditLogData.cs
```diff
@@ -51,16 +52,15 @@ internal static ScheduledEventUpdateAuditLogData Create(BaseDiscordClient discor
...
                 userCount?.OldValue.ToObject<Optional<int>>(discord.ApiClient.Serializer)
                     .GetValueOrDefault(),
                 image?.OldValue.ToObject<Optional<string>>(discord.ApiClient.Serializer)
                     .GetValueOrDefault()
             );
             var after = new ScheduledEventInfo(
                 guildId?.NewValue.ToObject<ulong>(discord.ApiClient.Serializer),
-                channelId?.NewValue.ToObject<ulong?>(discord.ApiClient.Serializer),
+                channelId == null ? null : channelId.NewValue?.ToObject<ulong?>(discord.ApiClient.Serializer) ?? 0,
```
Rekkonnect, line 61: Same here

src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventUpdateAuditLogData.cs
```diff
@@ -70,8 +70,7 @@ internal static ScheduledEventUpdateAuditLogData Create(BaseDiscordClient discor
                 status?.NewValue.ToObject<GuildScheduledEventStatus>(discord.ApiClient.Serializer),
                 entityType?.NewValue.ToObject<GuildScheduledEventType>(discord.ApiClient.Serializer),
                 entityId?.NewValue.ToObject<ulong?>(discord.ApiClient.Serializer),
-                entityMetadata?.NewValue.ToObject<GuildScheduledEventEntityMetadata>(discord.ApiClient.Serializer)
-                    ?.Location.GetValueOrDefault(),
+                location == null ? null : location.NewValue?.ToObject<string>(discord.ApiClient.Serializer) ?? Optional<string>.Unspecified,
```
Rekkonnect: Same here

Rekkonnect (changes requested):
Some code style changes

src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventInfo.cs
SaculRennorb: The optional is nullable to represent 3 states.
- null => no change
- empty optional => change to null value
- non empty optional => change to non null value

How would you like me to express these in a different way? 
If i combine 'change to null' and 'no change' into just making the field a `string?` you always have to check
`if(old.Location != null || new.Location != null) /*change occured*/;`.
None of the other fields behave that way.

src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventUpdateAuditLogData.cs
SaculRennorb: you literally cannot convert this in a null coalescing expression because of bracket and calling rules

src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventUpdateAuditLogData.cs
SaculRennorb: you literally cannot convert this in a null coalescing expression because of bracket and calling rules

src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventUpdateAuditLogData.cs
SaculRennorb: you literally cannot convert this in a null coalescing expression because of bracket and calling rules

src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventUpdateAuditLogData.cs
SaculRennorb: you literally cannot convert this in a null coalescing expression because of bracket and calling rules

src/Discord.Net.Rest/Entities/AuditLogs/DataTypes/ScheduledEventInfo.cs
SaculRennorb: @Rekkonnect any further comments?

Rekkonnect: `Optional<T>` and by consequence, `Optional<string>` on its own represents 3 states:
- `HasValue: false`: the object is not null, and denotes that there is no value of `string`
- `HasValue: true`: the object is not null, and denotes that there is a value of `string`, which is either `null` or not `null`.

What you do is introduce a fourth state, which you treat the exact same way as `HasValue: false`.

Sorry for the delayed response, I forgot

SaculRennorb: you are actually 100% correct. ive addressed the optional situation, but the ternary expressions cannot really be improved as far as i can tell.

SaculRennorb:
Hey @Rekkonnect , could you have another look at this?

Rekkonnect:
LGTM nice edits

csmir (approved):

PR #2487, opened 2022-10-08
[Fix] Make `ref` parameter of `message.HasCharPrefix`, `message.HasStringPrefix`, `message.HasCharPrefix` to be `out`

vladflotsky:
I think it`s better in terms of logic.

### DIFF
--- a/src/Discord.Net.Commands/Extensions/MessageExtensions.cs
+++ b/src/Discord.Net.Commands/Extensions/MessageExtensions.cs
@@ -16,35 +16,44 @@ namespace Discord.Commands
         /// <returns>
         ///     <c>true</c> if the message begins with the char <paramref name="c"/>; otherwise <c>false</c>.
         /// </returns>
-        public static bool HasCharPrefix(this IUserMessage msg, char c, ref int argPos)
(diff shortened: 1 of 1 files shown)


Cenngo:
What makes you think that replacing `ref` with `out` is better? argPos might already have an assigned value when control enters the method. `out` is only used when a parameter is definitely going to be reassigned or initialized in the method which isnt the case here.

vladflotsky:
At a minimum, these methods don't need a value of dev's argPos.

KubaZ2:
> What makes you think that replacing `ref` with `out` is better? argPos might already have an assigned value when control enters the method. `out` is only used when a parameter is definitely going to be reassigned or initialized in the method which isnt the case here.

What about `int.TryParse`? The second parameter is `out`, not `ref`.

I was always wondering why there is `ref` in this methods, but I was thinking that there is a reason. But now, I am seeing that there is no reason to use `ref`.

csmir (commented):
> This change is breaking and would need to be pushed to v4x.

ref is used in cases where the user has already done a check on a previous segment and passes a different arg position into the method. I do not approve of this change myself.

PR #2493, opened 2022-10-18
[Draft] Switch from Newtonsoft.Json to System.Text.Json

TheStachelfisch:
# Switching to System.Text.Json
This is a very early draft to switch from Newtonsoft.Json to System.Text.Json

**Draft Roadmap:**

- [ ] Migrate data classes (Partly done. Some features like `MemberSerialization = MemberSerialization.OptIn` need to be migrated)
- [ ] Migrate Clients and API classes
- [ ] Tests

Help with migrating is appreciated!

--- 

### Why switch?
Over the years System.Text.Json has come a long way and improved a lot, even beyond Newtonsoft.Json. Some major reasons of switching to System.Text.Json include:
- A major increase in performance when serializing and deserializing Data
- Cutting down on a 3rd party library and replacing it with an official one
- More robust in terms of security and reliability
- Performance that is tailored towards the targeted framework

### Additions with switching
A lot of code will need to be migrated over to System.Text.Json, many things that used to be specified with attributes when using Newtonsoft.Json will now need to be added as a setting in the serializer/deserializer.

### Problems with switching
No problems should arise if everything is tested correctly after the code was migrated over.

### DIFF
--- a/src/Discord.Net.Core/Discord.Net.Core.csproj
+++ b/src/Discord.Net.Core/Discord.Net.Core.csproj
@@ -11,7 +11,8 @@
     <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    <!-- For .Net 3.0 and earlier the System.Text.Json package has to be installed -->
+    <PackageReference Condition=" '$(TargetFramework)' == 'netstandard2.0' Or '$(TargetFramework)' == 'netstandard2.1' " Include="System.Text.Json" Version="7.0.0-rc.2.22472.3" />
     <PackageReference Include="System.Interactive.Async" Version="5.0.0" />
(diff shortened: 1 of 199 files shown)


JKamsker:
Have you done any benchmarks which support your Performance claims?
I mean, yes, System.Text.Json is faster than Newtonsoft, but does it really matter in this case?
Discord ratelimits requests anyways, so i dont quite see a huge benefit in migrating.

TheStachelfisch:
> Have you done any benchmarks which support your Performance claims? I mean, yes, System.Text.Json is faster than Newtonsoft, but does it really matter in this case? Discord ratelimits requests anyways, so i dont quite see a huge benefit in migrating.

There are official benchmarks from Microsoft which compared Newtonsoft.Json to System.Text.Json, System.Text.Json generally performed better in serializing and deserializing, especially with small data and very big data.

For Discord big data can be chunks of downloaded members or just simple small messages.

But after consulting with Quinch, a switch to System.Text.Json is too much trouble to be worth it. So I'll also be closing this Draft with this comment.

Issue #2381, opened 2022-07-08
[Bug]: "401 Unauthorized" on several REST client methods


sjyasuosj:
### Check The Docs

- [X] I double checked the docs and couldn't find any useful information.

### Verify Issue Source

- [X] I verified the issue was caused by Discord.Net.

### Check your intents

- [X] I double checked that I have the required intents.

### Description

Using DiscordRestClient errors for several endpoints with "401 Unauthorized". 

I'm able to successfully log in using bearer token and verified that it pulls my profile data by looking at ` CurrentUser` property. However, as soon as I call any of the API methods to pull any other data, it errors out and gives me "401 Unauthorized". 

So far I've tried:
`GetDMChannelsAsync`, `GetGuildsAsync`, `GetGroupChannelsAsync`, `GetPrivateChannelsAsync` and `GetGuildUserAsync`.

All of those failed.

### Version

3.7.2

### Logs

      An unhandled exception has occurred while executing the request.
      Discord.Net.HttpException: The server responded with error 401: 401: Unauthorized
         at Discord.Net.Queue.RequestBucket.SendAsync(RestRequest request)
         at Discord.Net.Queue.RequestQueue.SendAsync(RestRequest request)
         at Discord.API.DiscordRestApiClient.SendInternalAsync(String method, String endpoint, RestRequest request)
         at Discord.API.DiscordRestApiClient.SendAsync[TResponse](String method, String endpoint, BucketId bucketId, ClientBucketType clientBucket, RequestOptions options)
         at Discord.API.DiscordRestApiClient.GetGuildAsync(UInt64 guildId, Boolean withCounts, RequestOptions options)
         at Discord.Rest.ClientHelper.GetGuildsAsync(BaseDiscordClient client, Boolean withCounts, RequestOptions options)

### Sample

```
await restClient.LoginAsync(TokenType.Bearer, accessToken);
var guilds = await restClient.GetDMChannelsAsync(); <-- fails
```

### Packages

N/A

csmir:
Does this issue persist in the latest source nightly?

Misha-133:
Closing due to lack of followup

PR #2516, opened 2022-11-21
Bearer token fixes

tg2k:
This attempts to fix a couple issues I ran into in testing with bearer tokens. The prefixed tokens were rejected with a 401 Unauthorized, and without the prefix the constructor of DiscordRestApiClient threw an exception due to the null token (which got set later).

In addition, if a DM involved a deleted user, the channel would show up as having no recipients. My quick fix for this was simply not to add the private channel.

### DIFF
--- a/src/Discord.Net.Rest/DiscordRestApiClient.cs
+++ b/src/Discord.Net.Rest/DiscordRestApiClient.cs
@@ -75,7 +75,8 @@ namespace Discord.API
             RestClient = _restClientProvider(baseUrl);
             RestClient.SetHeader("accept", "*/*");
             RestClient.SetHeader("user-agent", UserAgent);
-            RestClient.SetHeader("authorization", GetPrefixedToken(AuthTokenType, AuthToken));
+            if (!string.IsNullOrEmpty(AuthToken))
+                RestClient.SetHeader("authorization", GetPrefixedToken(AuthTokenType, AuthToken));
         }
         /// <exception cref="ArgumentException">Unknown OAuth token type.</exception>
         internal static string GetPrefixedToken(TokenType tokenType, string token)
@@ -83,7 +84,7 @@ namespace Discord.API
             return tokenType switch
             {
                 TokenType.Bot => $"Bot {token}",
-                TokenType.Bearer => $"Bearer {token}",
+                TokenType.Bearer => $"{token}",  // Bearer tokens are not prefixed
                 _ => throw new ArgumentException(message: "Unknown OAuth token type.", paramName: nameof(tokenType)),
             };
         }
--- a/src/Discord.Net.WebSocket/DiscordSocketApiClient.cs
+++ b/src/Discord.Net.WebSocket/DiscordSocketApiClient.cs
@@ -234,8 +234,16 @@ namespace Discord.API
                 {
                     if (!_isExplicitUrl && _gatewayUrl == null)
                     {
-                        var gatewayResponse = await GetBotGatewayAsync().ConfigureAwait(false);
-                        _gatewayUrl = FormatGatewayUrl(gatewayResponse.Url);
+                        if (AuthTokenType == TokenType.Bot)
+                        {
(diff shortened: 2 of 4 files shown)


theiam79:
Hoping this fixes #2381 as I'm still having the same problem

src/Discord.Net.Rest/DiscordRestApiClient.cs
```diff
@@ -75,15 +75,16 @@ internal void SetBaseUrl(string baseUrl)
...
+                RestClient.SetHeader("authorization", GetPrefixedToken(AuthTokenType, AuthToken));
         }
         /// <exception cref="ArgumentException">Unknown OAuth token type.</exception>
         internal static string GetPrefixedToken(TokenType tokenType, string token)
         {
             return tokenType switch
             {
                 TokenType.Bot => $"Bot {token}",
-                TokenType.Bearer => $"Bearer {token}",
+                TokenType.Bearer => $"{token}",  // Bearer tokens are not prefixed
```
quinchs, line 87: Is this correct? as per [the discord api docs](https://discord.com/developers/docs/reference#authentication), the authorization header follows this format: `Authorization: TOKEN_TYPE TOKEN`.

Also, changing this may break other usecases of bearer tokens

tg2k: I don't know whether it is technically correct. What I do know is that:

- The browser on discord.com/app uses an unprefixed token (with the v9 api)
- Doing this made HTTP errors (401 Unauthorized provided my notes are correct) go away
- I have not personally noticed any ill effects from this change nearly 4 weeks later

Am I using an incorrect token? I was following [this StackOverflow post](https://stackoverflow.com/questions/72605540/getting-bearer-token-for-discord-api).

quinchs: That kind of token is a User Account token, which per the discord TOS is not allowed to be used by 3rd party tools (such as your code and this library).

theiam79: I'll be honest, I hadn't messed with the API directly much. I'm trying to use discord as an authentication provider and then call the API with the subsequent claims/tokens

tg2k: > That kind of token is a User Account token, which per the discord TOS is not allowed to be used by 3rd party tools (such as your code and this library).

Thanks. I'll take a second look when I'm able to circle back to this again.

theiam79: @quinchs where are you seeing this in the ToS? As far as I can tell, it's valid to use a user token to hit the API so long as it's the user initiating the action/it's not automated. 

> Automating normal user accounts (generally called "self-bots") outside of the OAuth2/bot API is forbidden, and can result in account termination if found.

Perhaps it'd be best to add a new token type/a different rest client if the library was to support that.

SylveonDeko:
Lookup the term "self botting", that's what you're attempting to do, and it's not allowed, hence the reason they constantly update the API to disallow user tokens from most endpoints

src/Discord.Net.Rest/DiscordRestApiClient.cs
SylveonDeko: Lookup the term "self botting", that's what you're attempting to do, and it's not allowed, hence the reason they constantly update the API to disallow user tokens from most endpoints

theiam79: @SylveonDeko I'm not automating anything - the user authenticates via discord, and I need to retrieve a list of their guilds so they can choose one for some of my bots functionality. This is happening inside of a web app, not inside of a bot command. If there's some other way I'm supposed to accomplish this I'm all ears, but I fail to see how that's "self botting"

NeutraleNull: so in short if one wants this fixed one has to fork all of this and do it themself?

PR #2524, opened 2022-12-06
Fixed syntax error in example AutocompleteHandler

Fridious:
It only fix a syntax error in a code example in the docs -> https://discordnet.dev/guides/int_framework/autocompletion.html#creating-autocompletehandlers


Misha12 (approved):

MrCakeSlayer:
There isn't an error here, the current version is correct.

PR #2532, opened 2022-12-13
[Fix] AddXToGuildAsync with deleteMissing=false causes all existing to require Admin

### LINKED ISSUE #2515: [Bug]: Incorrect command permissions after using AddCommandsToGuildAsync with deleteMissing=false
### Check The Docs

### Verify Issue Source

### Check your intents

### Description

If you use `AddCommandsToGuildAsync` with `deleteMissing` parameter set to `false` (or `AddModulesToGuildAsync`) already existing commands will start to require `Administator` permissions.

As far as I noticed the only difference between them inside `BulkOverwriteGuildCommandsAsync` method is `DefaultMemberPermission` property.
The old command (that will will start to require `Administator` permission) have `0`.
image
While the new one have `null`.
image

And changing it to `null` here resolves the problem. Which is kinda strange, `0` should mean no permissions according to the docs.

### Version

3.8.1

### Working Version

_No response_

### Logs

-

### Sample

```cs
// add a command with no permissions
await _interaction.AddCommandsToGuildAsync(guild, true, command1);
// add a second one
await _interaction.AddCommandsToGuildAsync(guild, false, command2);

// command1 now requires Administrator permission
```

### Packages

-

cjbonn:
This PR fixes #2515 - the Administrator overwriting permission issue

### DIFF
--- a/src/Discord.Net.Interactions/Utilities/ApplicationCommandRestUtil.cs
+++ b/src/Discord.Net.Interactions/Utilities/ApplicationCommandRestUtil.cs
@@ -227,7 +227,7 @@ namespace Discord.Interactions
                     Name = command.Name,
                     Description = command.Description,
                     IsDefaultPermission = command.IsDefaultPermission,
-                    DefaultMemberPermissions = (GuildPermission)command.DefaultMemberPermissions.RawValue,
+                    DefaultMemberPermissions = command.DefaultMemberPermissions.RawValue == 0 ? new Optional<GuildPermission>() : (GuildPermission)command.DefaultMemberPermissions.RawValue,
                     IsDMEnabled = command.IsEnabledInDm,
                     IsNsfw = command.IsNsfw,
                     Options = command.Options?.Select(x => x.ToApplicationCommandOptionProps())?.ToList() ?? Optional<List<ApplicationCommandOptionProperties>>.Unspecified,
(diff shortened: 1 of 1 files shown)


src/Discord.Net.Interactions/Utilities/ApplicationCommandRestUtil.cs
```diff
@@ -223,7 +223,7 @@ public static ApplicationCommandProperties ToApplicationCommandProps(this IAppli
                     Name = command.Name,
                     Description = command.Description,
                     IsDefaultPermission = command.IsDefaultPermission,
-                    DefaultMemberPermissions = (GuildPermission)command.DefaultMemberPermissions.RawValue,
+                    DefaultMemberPermissions = command.DefaultMemberPermissions.RawValue == 0 ? new Optional<GuildPermission>() : (GuildPermission)command.DefaultMemberPermissions.RawValue,
```
cjbonn, line 226: Looking at this more, would it make more sense to overwrite as such?
`DefaultMemberPermissions = command.DefaultMemberPermissions.SanitizeGuildPermissions(),`

The issue opened uses the Optional, which I don't believe is necessary.

Cenngo, line 230: Is it possible to use the approach in L108 here as well. It would be nice if we can keep everything consistent

cjbonn: > Is it possible to use the approach in L108 here as well. It would be nice if we can keep everything consistent

L108's Builder uses `GuildPermission?`, while these XCommandProperties.DefaultUserPermissions uses `Optional<GuildPermission>`. The Sanitize helper method is geared towards the former, not the latter. It's already a bit inconsistent (the Builder on L108 ends up converting to Optional<> in the constructor anyway). I can definitely make a secondary method for this data type however - let me know.

csmir (approved):

src/Discord.Net.Interactions/Utilities/ApplicationCommandRestUtil.cs
cjbonn: @Cenngo any additional feedback here, or leave as-is ?

PR #2469, opened 2022-09-14
[Feature] Add missing properties in forum & thread channels

Misha-133:
This PR adds support for missing channel properties and improves channel modification

Also it fixes a bug in `DiscordRestClient`

### Changes
- [x] new properties for editing channels:
  - `ThreadChannelProperties`
  - `ForumChannelProperties`
- [x] move `Locked` and `Archived` props from `TextChannelProperties` into `ThreadChannelProperties` **Might be breaking**
- [x] add `AppliedTags` property to:
  -  `IThreadChannel`
  - `RestThreadChannel`
  - `SocketThreadChannel`
  - `ThreadChannelProperties`
- [x] add `DefaultSlowModeInterval` and `ThreadCreationInterval` properties
  - to forum channels
  - to `ForumChannelProperties`
- [x] add support for `default_reaction_emoji`
- [x] add `ChannelFlags`
- [x] add `ForumTagBuilder`
- [x] add comparison operators on forum tags
- [x] add `CreateForumChannelAsync` methods to guilds
- [x] add `ForumChannels` property to `SocketGuild`
- [x] add support for `default_sort_order`\
- [x] support creating post with applied tags (also fixes #2503) **Might be breaking**

### Fixes
- [x] Fix `DiscordRestClient.GetChannelAsync` returning `null` if getting forum channel
- [x] Fix #2475
- [x] Fix misleading comment on `ApplicationCommandOptionType.Integer` 
- [x] Fix #2502

PR #2546, opened 2022-12-25
[Fix] Creating post in threads with required tags

Misha-133:
Apparently I made some oversights in #2469 PR which went unnoticed until now

### Changes
- fix throwing exception on thread creation in forums that require posts to have at least 1 tag


csmir (approved):

ZhJiYuan:
Here is my code.
Can you please tell me what is the reason, thank you very much.

```
try
{
    IForumChannel channel = discord.GetChannel(1056545567339196497);
    await channel.CreatePostAsync(title: "title", text: "hello", tags: new[] { channel.Tags[0], channel.Tags[1] });
}
catch (Exception ex)
{
    Console.WriteLine(ex.Message);
}
```

Error code：The channel test requires posts to have at least one tag.

PR #2557, opened 2023-01-03
Update customId template generator to escape regex metachars used in …

### LINKED ISSUE #2555: [Bug]: Button regex does not work
### Check The Docs

### Verify Issue Source

### Check your intents

### Description

So, for some reason TreatAsRegex is off by default, which was not mentioned and probably many many bots needed refactoring, however even with TreatAsRegex enabled, for some reason a simple regex like ```[ComponentInteraction("trigger.*.runin.*$*", true), CheckPermissions]``` Does not work.

For some reason, it goes through all preconditions just fine but never runs the actual method

### Version

3.9

### Working Version

3.8.1

### Logs

No stack traces

### Sample

_No response_

### Packages

None

Cenngo:
Fixes #2555.


cjbonn (approved):
Tested on my end, lgtm

csmir (approved):

PR #2564, opened 2023-01-07
[Feature] RespondWithModal() which accepts an IModal instance as template

Cenngo:
Example usage:
```csharp
[SlashCommand("summon-modal", "desc")]
public async Task SummonModalAsync()
{
    Context.Interaction.RespondWithModal("modal_id", new TestModal
    {
        FirstField = "Initial Value1",
        SecondField = "Initial Value2"
    });
}
```

### DIFF
--- a/src/Discord.Net.Interactions/Extensions/IDiscordInteractionExtensions.cs
+++ b/src/Discord.Net.Interactions/Extensions/IDiscordInteractionExtensions.cs
@@ -45,12 +45,12 @@ namespace Discord.Interactions
         }
 
         /// <summary>
-        ///     Respond to an interaction with a <see cref="IModal"/> and fills the value fields of the modal using the property values of the provided
+        ///     Respond to an interaction with an <see cref="IModal"/> and fills the value fields of the modal using the property values of the provided
         ///     instance.
         /// </summary>
         /// <typeparam name="T">Type of the <see cref="IModal"/> implementation.</typeparam>
         /// <param name="interaction">The interaction to respond to.</param>
-        /// <param name="modal"><see cref="IModal"/> instance to get field values from.</param>
+        /// <param name="modal">The <see cref="IModal"/> instance to get field values from.</param>
         /// <param name="options">The request options for this <see langword="async"/> request.</param>
         /// <param name="modifyModal">Delegate that can be used to modify the modal.</param>
         /// <returns></returns>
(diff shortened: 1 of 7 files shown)


src/Discord.Net.Interactions/Extensions/IDiscordInteractionExtensions.cs
```diff
@@ -44,6 +44,44 @@ public static async Task RespondWithModalAsync<T>(this IDiscordInteraction inter
             await SendModalResponseAsync(interaction, customId, modalInfo, options, modifyModal);
         }
 
+        /// <summary>
+        ///     Respond to an interaction with a <see cref="IModal"/> and fills the value fields of the modal using the property values of the provided
```
csmir:
```suggestion
        ///     Respond to an interaction with an <see cref="IModal"/> and fills the value fields of the modal using the property values of the provided
```
🤓

src/Discord.Net.Interactions/Extensions/IDiscordInteractionExtensions.cs
```diff
@@ -44,6 +44,44 @@ public static async Task RespondWithModalAsync<T>(this IDiscordInteraction inter
...
+        ///     instance.
+        /// </summary>
+        /// <typeparam name="T">Type of the <see cref="IModal"/> implementation.</typeparam>
+        /// <param name="interaction">The interaction to respond to.</param>
+        /// <param name="modal"><see cref="IModal"/> instance to get field values from.</param>
```
csmir:
```suggestion
        /// <param name="modal">The <see cref="IModal"/> instance to get field values from.</param>
```

src/Discord.Net.Interactions/Extensions/IDiscordInteractionExtensions.cs
```diff
@@ -44,6 +44,44 @@ public static async Task RespondWithModalAsync<T>(this IDiscordInteraction inter
...
+        /// <param name="options">The request options for this <see langword="async"/> request.</param>
+        /// <param name="modifyModal">Delegate that can be used to modify the modal.</param>
```
csmir, line 55: Why is `modifyModal` after `options`? Common library behavior places options last.

csmir (changes requested):

src/Discord.Net.Interactions/Extensions/IDiscordInteractionExtensions.cs
Cenngo: factory delegate was an afterthought and was added a good amount of time after the Modals commit so i didnt want it to break existing code bases

csmir: This should be marked to be changed for v4. I will open an issue. Next to that, lgtm

csmir (approved):

PR #2575, opened 2023-01-27
[Feature] add `Thread` property to `RestMessage`

Misha-133:
This PR adds a `RestThreadChannel` property to `RestMessage`.

### Changes
- add a `RestThreadChannel Thread` property to `RestMessage`.

### DIFF
--- a/src/Discord.Net.Rest/Entities/Messages/RestMessage.cs
+++ b/src/Discord.Net.Rest/Entities/Messages/RestMessage.cs
@@ -80,6 +80,12 @@ namespace Discord.Rest
         /// <inheritdoc/>
         public MessageType Type { get; private set; }
 
+
+        /// <summary>
+        ///     Gets the thread that was started from this message.
+        /// </summary>
+        public RestThreadChannel Thread { get; private set; }
+
         /// <inheritdoc />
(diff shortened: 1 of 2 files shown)


src/Discord.Net.Rest/Entities/Messages/RestMessage.cs
```diff
@@ -80,6 +80,12 @@ public abstract class RestMessage : RestEntity<ulong>, IMessage, IUpdateable
         /// <inheritdoc/>
         public MessageType Type { get; private set; }
 
+
+        /// <summary>
+        ///     Gets the thread that was started from this message.
+        /// </summary>
+        public RestThreadChannel Thread { get; private set; }
```
csmir, line 90: Null if it doesn't exist? Add a remark defining that behavior

csmir (changes requested):

src/Discord.Net.Rest/Entities/Messages/RestMessage.cs
Misha-133: I've pushed a fix... should be resolved? 
*I was really struggling to write a good enough comment and i'm really not sure how to describe it in a better way*

csmir (approved):

PR #2579, opened 2023-02-01
[Feature] Rework `Thread` property of a message

Misha-133:
This PR is a followup to previous #2575 PR; 

### Changes
- move `Thread` property to `IUserMessage`
- add `Thread` property to `SocketUserMessage` (since it might be present in component interaction or context command payload)


csmir:
Why is it user message only? System messages also support nested threads, and should thus also be allowed to have access to this getter:
image

Misha-133:
> Why is it user message only? System messages also support nested threads, and should thus also be allowed to have access to this getter:
> image
> 

Huh, seems like I've missed this case.

csmir (approved):

PR #2580, opened 2023-02-02
[Feature] Thread user member & thread members pagination

Misha-133:
This PR adds support for the guild member property of thread member & thread member pagination.

### Changes
- `RestThreadChannel.GetUsersAsync()` now wraps `RestThreadChannel.GetThreadUsersAsync()` to keep old code working
- add `RestThreadChannel.GetThreadUsersAsync()` method which returns an `IAsyncEnumerable`
- add `GuildUser` property to `IThreadUser`

### DIFF
--- a/src/Discord.Net.Core/DiscordConfig.cs
+++ b/src/Discord.Net.Core/DiscordConfig.cs
@@ -214,5 +214,13 @@ namespace Discord
         ///     If set to <see langword="false"/>, this value will be "Discord#1234".
         /// </remarks>
         public bool FormatUsersInBidirectionalUnicode { get; set; } = true;
+
+        /// <summary> 
+        ///     Returns the max thread members allowed to be in a request. 
+        /// </summary>
+        /// <returns>
+        ///     The maximum number of thread members that can be gotten per-batch.
+        /// </returns>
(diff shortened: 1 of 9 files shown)


csmir:
Discord API changes are considered exempt from our semantic versioning, since there is no way to prevent these changes from happening on Discord's end. Though, I would like some input here. Do we keep the old method, and allow it as standard for top 100 members? This way, people who upgrade won't need any refactors

Misha-133:
> Discord API changes are considered exempt from our semantic versioning, since there is no way to prevent these changes from happening on Discord's end. Though, I would like some input here. Do we keep the old method, and allow it as standard for top 100 members? This way, people who upgrade won't need any refactors

I mostly agree with your statement
My point is that having multiple similarly named method would create some unnecessary confusion between them.

But now thinking about it - we could have the old method wrap the new one so it would keep its full functionality while also creating a new method that would allow to have more control over the process.
I think this would be a better way of implementing this functionality
I'll make necessary changes shortly.

csmir:
I advice marking the old method as obsolete. This way people will be notified to make changes, and future us can go and remove it completely for v4.

csmir (approved):

PR #2598, opened 2023-02-13
Fix ModuleBuilder.AddSlashCommand/AddAutocompleteCommand typo

dedmen:
Method name was typo'ed, probably when it was created


csmir:
This is breaking. This would need to be moved to v4.x...

PR #2605, opened 2023-02-19
[Feature]  Modify guild features

Misha-133:
This PR adds support for modifying guild features & adds a few missing ones.

### Changes
- add missing guild features
- add `GuildFeatures` json writer
- add `GuildFeature` property to `GuildProperties`

### DIFF
--- a/src/Discord.Net.Core/Entities/Guilds/GuildProperties.cs
+++ b/src/Discord.Net.Core/Entities/Guilds/GuildProperties.cs
@@ -113,5 +113,10 @@ namespace Discord
         ///     Gets or sets if the boost progress bar is enabled.
         /// </summary>
         public Optional<bool> IsBoostProgressBarEnabled { get; set; }
+
+        /// <summary>
+        ///     Gets or sets the guild features enabled in this guild.
+        /// </summary>
+        public Optional<GuildFeature> Features { get; set; }
     }
 }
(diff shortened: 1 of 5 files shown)


csmir (changes requested):
Would it not be wise to check the mutable features, to protect the developer from editing before unavailable ones get sent to discord? Maybe I read over this check, but I do not see it anywhere.

Misha-133:
> Would it not be wise to check the mutable features, to protect the developer from editing before unavailable ones get sent to discord? Maybe I read over this check, but I do not see it anywhere.

Discord ignores immutable features
is this kind of validation needed in this case?

csmir:
Just a comment now is fine then, I will merge it

src/Discord.Net.Core/Entities/Guilds/GuildProperties.cs
```diff
@@ -113,5 +113,10 @@ public class GuildProperties
         ///     Gets or sets if the boost progress bar is enabled.
         /// </summary>
         public Optional<bool> IsBoostProgressBarEnabled { get; set; }
+
+        /// <summary>
+        ///     Gets or sets the guild features enabled in this guild.
```
csmir:
```suggestion
        ///     Gets or sets the guild features enabled in this guild. Features that are not mutable will be ignored.
```

csmir (dismissed):

csmir (approved):

PR #2644, opened 2023-03-30
[Feature]: ComponentBuilder component removal methods

cjbonn:
This PR adds two helper methods to ComponentBuilder to assist in the removal of components.

- `ComponentBuilder.RemoveComponent(customId)` will remove a component based on its custom id.
- `ComponentBuilder.RemoveComponentsOfType(componentType)` will remove all components of a given type (ie: button, menu, etc).

Build() will now automatically resolve empty ActionRows.

An example that will always remove the button that is clicked from a message:
```cs
[ComponentInteraction("testBtn:*", ignoreGroupNames: true)]
public async Task HandleButtonClick(string id) {
      ComponentBuilder comp = ComponentBuilder.FromMessage((Context.Interaction as IComponentInteraction).Message);
      comp.RemoveComponent($"testBtn:{id}");
      await (Context.Interaction as IComponentInteraction).UpdateAsync(m => { m.Components = comp.Build(); });
}
```

### DIFF
--- a/src/Discord.Net.Core/Entities/Interactions/MessageComponents/ComponentBuilder.cs
+++ b/src/Discord.Net.Core/Entities/Interactions/MessageComponents/ComponentBuilder.cs
@@ -82,6 +82,28 @@ namespace Discord
         }
 
         /// <summary>
+        ///     Removes all components of the given type from the <see cref="ComponentBuilder"/>.
+        /// </summary>
+        /// <param name="t">The <see cref="ComponentType"/> to remove.</param>
+        /// <returns>The current builder.</returns>
+        public ComponentBuilder RemoveComponentsOfType(ComponentType t)
+        {
+            this.ActionRows.ForEach(ar => ar.Components.RemoveAll(c => c.Type == t));
+            return this;
+        }
+
+        /// <summary>
+        ///     Removes a component from the <see cref="ComponentBuilder"/>.
+        /// </summary>
+        /// <param name="customId">The custom id of the component.</param>
+        /// <returns></returns>
+        public ComponentBuilder RemoveComponent(string customId)
+        {
+            this.ActionRows.ForEach(ar => ar.Components.RemoveAll(c => c.CustomId == customId));
+            return this;
+        }
+
+        /// <summary>
         ///     Adds a <see cref="SelectMenuBuilder"/> to the <see cref="ComponentBuilder"/> at the specific row.
         ///     If the row cannot accept the component then it will add it to a row that can.
         /// </summary>
--- a/src/Discord.Net.Core/Entities/Interactions/MessageComponents/ComponentBuilder.cs
+++ b/src/Discord.Net.Core/Entities/Interactions/MessageComponents/ComponentBuilder.cs
@@ -99,7 +99,7 @@ namespace Discord
         /// <returns>The current builder.</returns>
         public ComponentBuilder RemoveComponent(string customId)
         {
-            this.ActionRows.ForEach(ar => ar.Components.RemoveAll(c => c.CustomId == customId || (c is ButtonComponent b && b.Url == customId)));
+            this.ActionRows.ForEach(ar => ar.Components.RemoveAll(c => c.CustomId == customId));
             return this;
         }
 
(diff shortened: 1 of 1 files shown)


Misha-133:
IMO it would be a good idea to add a method to remove a link-button by a URL, since one doesn't have a custom id

cjbonn:
> IMO it would be a good idea to add a method to remove a link-button by a URL, since one doesn't have a custom id

Good call. What would you suggest an example implementation looks like? `RemoveButtonByURL(url)`?

Misha-133:
> `RemoveButtonByURL(url)`

Looks good to me
👍

cjbonn:
@Misha-133 not sure I can/want to do the link button one - [IMessageComponent](https://github.com/discord-net/Discord.Net/blob/933ea42eaac47094ef77608aa2aa3f6d602ac30d/src/Discord.Net.Core/Entities/Interactions/MessageComponents/IMessageComponent.cs) only exposes custom id & component type to reference. Adding to that interface would mean forcing an additional unused property in all the other components that don't need it (ie: SelectMenu, TextInput, etc).

Misha-133:
> @Misha-133 not sure I can/want to do the link button one - [IMessageComponent](https://github.com/discord-net/Discord.Net/blob/933ea42eaac47094ef77608aa2aa3f6d602ac30d/src/Discord.Net.Core/Entities/Interactions/MessageComponents/IMessageComponent.cs) only exposes custom id & component type to reference. Adding to that interface would mean forcing an additional unused property in all the other components that don't need it (ie: SelectMenu, TextInput, etc).

You could cast the component to a button tho
something like 
```cs
if(component is ButtonComponent button && button.Url == ......)
```

cjbonn:
I was on the right track, just derped and used `as` instead of `is`, forgetting that it can be any component in there. This is why I don't code in the morning :')

src/Discord.Net.Core/Entities/Interactions/MessageComponents/ComponentBuilder.cs
```diff
@@ -81,6 +81,35 @@ internal void AddComponent(IMessageComponent component, int row)
...
+        }
+
+        /// <summary>
+        ///     Removes a component from the <see cref="ComponentBuilder"/>.
+        /// </summary>
+        /// <param name="customId">The custom id of the component.</param>
+        /// <returns>The current builder.</returns>
+        public ComponentBuilder RemoveComponent(string customId)
+        {
+            this.ActionRows.ForEach(ar => ar.Components.RemoveAll(c => c.CustomId == customId || (c is ButtonComponent b && b.Url == customId)));
```
csmir: Why does this remove matching link buttons but not document this behaviour? This may cause unexpected behaviour, being difficult to debug. I would suggest making that an optional feature, or taking it out of this method entirely.

csmir (changes requested):

Misha-133:
Also, shouldn't these methods support regex wildcards? So it'd be possible to do something like `RemoveButton("buttonId_*")`

src/Discord.Net.Core/Entities/Interactions/MessageComponents/ComponentBuilder.cs
cjbonn: fair enough, will split it out

cjbonn:
> Also, shouldn't these methods support regex wildcards? So it'd be possible to do something like `RemoveButton("buttonId_*")`

I despise working with Regex, but I'll take a gander when I'm more free and want to torture myself with it. 🥲

cjbonn:
I took a look, figured I could just reference the Regex implementation that is used in the ComponentInteration attribute, but I cannot find it 😳 thinking we should just use this PR as-is, and have another PR sometime in the future that will expand the functionality of these to support regex.

csmir:
A feature that does not support functionality the library offers out of the box should not be merged. @Cenngo, could you reference the component regex?

Cenngo:
see https://github.com/discord-net/Discord.Net/blob/dev/src/Discord.Net.Interactions/Utilities/RegexUtils.cs#L94-L157

cjbonn:
> see https://github.com/discord-net/Discord.Net/blob/dev/src/Discord.Net.Interactions/Utilities/RegexUtils.cs#L94-L157

This is internal to the Discord.Net.Interaction namespace; ComponentBuilder is in Core... should I just copy all of the non-dependent code into either Core/Utils or ComponentBuilder.cs? or is there a better way to go about this?

Cenngo:
you could add the .Remove methods to the Discord.Interactions assembly as extension methods

cjbonn:
> you could add the .Remove methods to the Discord.Interactions assembly as extension methods

True, but along the lines of reusing that code, it is dependent on ICommandInfo and InteractionServiceConfig.WildCardExpression, both of which I don't think an extension method (or otherwise) would have any context for?

cjbonn:
Alright, well I hoped to add a couple useful functions that assisted me in my project so others could have that functionality too, rather than not at all - but this is not something I intend to spend hours digging into, I have enough on my plate as it is. If this absolutely cannot be included without regex support and be incrementally developed to add regex later down the line, then please go ahead and close this PR. thx

quinchs (approved):

PR #2652, opened 2023-04-03
[Fix] Don't dispose streams in DefaultRestClient

### LINKED ISSUE #2194: [Bug]: Modifying a message with attachments when hitting rate limits attempts to read/close/read the stream passed in
### Check The Docs

### Verify Issue Source

### Check your intents

### Description

When calling ModifyMessageAsync and attempting to change/add attachments, if you are getting rate limited the stream you pass in will be read, sent to discord and then disposed when you get the rate limit error. It will then try to send it again, which will attempt to re-read the stream, which will fail for most streams. To work around this, I wrapped my stream in a wrapper that claimed it was not seekable, since the d.net code will copy a non-seekable stream to its own memoryStream and use (and dispose) that stream.

### Version

3.3.2

### Working Version

_No response_

### Logs

System.ObjectDisposedException: Cannot access a closed Stream.
   at System.IO.StreamHelpers.ValidateCopyToArgs(Stream source, Stream destination, Int32 bufferSize)
   at System.IO.MemoryStream.CopyToAsync(Stream destination, Int32 bufferSize, CancellationToken cancellationToken)
   at Discord.Net.Rest.DefaultRestClient.SendAsync(String method, String endpoint, IReadOnlyDictionary2 multipartParams, CancellationToken cancelToken, Boolean headerOnly, String reason)
   at Discord.Net.Queue.MultipartRestRequest.SendAsync()
   at Discord.API.DiscordRestApiClient.SendMultipartAsync[TResponse](String method, String endpoint, IReadOnlyDictionary2 multipartArgs, BucketId bucketId, ClientBucketType clientBucket, RequestOptions options)
   at Discord.API.DiscordRestApiClient.ModifyMessageAsync(UInt64 channelId, UInt64 messageId, UploadFileParams args, RequestOptions options)
   at Discord.Rest.MessageHelper.ModifyAsync(UInt64 channelId, UInt64 msgId, BaseDiscordClient client, Action1 func, RequestOptions options)
   at Discord.Rest.ChannelHelper.ModifyMessageAsync(IMessageChannel channel, UInt64 messageId, Action1 func, BaseDiscordClient client, RequestOptions options)
   at Discord.WebSocket.SocketTextChannel.ModifyMessageAsync(UInt64 messageId, Action`1 func, RequestOptions options) 

### Sample

```cs
using MemoryStream memStr = new MemoryStream();
memStr.writeByte(123);
memStr.Position = 0;

await channel.ModifyMessageAsync(messageId, m =>
                    {
                        m.Content = "hi there";
                        m.Attachments = new List<FileAttachment>() { new FileAttachment(memStr, "zone_map.png")};
                    });

<.. repeat above ModifyMessageAsync a couple times to trigger a rate limit>

```

### Packages

None

ben-reilly:
## Context
In `DefaultRestClient.SendAsync(...)`, we create an `HttpRequestMessage` into which we put all of the content that we are going to send. `HttpRequestMessage` is disposable, which usually means that we should take responsibility for disposing it.

However, its `Dispose()` method merely disposes of the content objects in its `Content` property, and we don't always want to dispose those. For instance, if the stream was passed in by the function caller, we ought not to dispose of that stream because the caller may intend to reuse it. And the caller may be forced to reuse it if the request comes back with a 429 status and needs to be retried.

And other standard `HttpContent` classes like `StringContent` and `ByteArrayContent` don't need to be disposed.

In fact, [.NET devs have mentioned in the past](https://github.com/aspnet/Security/issues/886#issuecomment-229181249) that it is not necessary to dispose of `HttpRequestMessage` for this reason. And its implementation can be confirmed in [the source for the class](https://github.com/microsoft/referencesource/blob/master/System/net/System/Net/Http/HttpRequestMessage.cs).

## Bug in current code + solution
Since we have been disposing the `HttpRequestMessage` object, in some cases (see #2194) we have been disposing streams that have been passed in. Then subsequent uses of that stream (either intentionally or as a result of a retried 429) result in an `ObjectDisposedException`.

We could continue to dispose the `HttpRequestMessage` and copy all streams provided by the caller so that we only dispose those copies. But doing that means that we read through the stream twice for every request, and we don't really gain anything from it.

I believe the simpler option is to just use the streams passed in by the caller and expect the caller to handle its disposal.

Fixes #2194.

## Trying it out

```cs
var webhookClient = new DiscordWebhookClient(...);
var fileAttachment = new FileAttachment(pathToFile);

// Reuse the same FileAttachment object for two messages
await webhookClient.SendFileAsync(fileAttachment, "Message 1");
await webhookClient.SendFileAsync(fileAttachment, "Message 2");
```


src/Discord.Net.Rest/Entities/Guilds/RestGuild.cs
```diff
@@ -1071,7 +1071,7 @@ public async Task<CustomSticker> CreateStickerAsync(string name, Image image, IE
         public Task<CustomSticker> CreateStickerAsync(string name,  string path, IEnumerable<string> tags, string description = null,
             RequestOptions options = null)
         {
-            var fs = File.OpenRead(path);
+            using var fs = File.OpenRead(path);
```
ben-reilly, line 1074: This function can no longer expect the client/sender to dispose of the object.

src/Discord.Net.Rest/Entities/Interactions/Modals/RestModal.cs
```diff
@@ -228,14 +228,15 @@ public override string Defer(bool ephemeral = false, RequestOptions options = nu
             fileName ??= Path.GetFileName(filePath);
             Preconditions.NotNullOrEmpty(fileName, nameof(fileName), "File Name must not be empty or null");
 
+            using var fileStream = string.IsNullOrEmpty(filePath) ? new MemoryStream(File.ReadAllBytes(filePath), false) : null;
```
ben-reilly: This function can no longer expect the client/sender to dispose of the object.

src/Discord.Net.WebSocket/Entities/Guilds/SocketGuild.cs
```diff
@@ -1561,7 +1561,7 @@ public SocketCustomSticker GetSticker(ulong id)
         public Task<SocketCustomSticker> CreateStickerAsync(string name,  string path, IEnumerable<string> tags, string description = null,
             RequestOptions options = null)
         {
-            var fs = File.OpenRead(path);
+            using var fs = File.OpenRead(path);
```
ben-reilly, line 1564: This function can no longer expect the client/sender to dispose of the object.

ben-reilly (commented):
I did a review of all of the places where `DefaultRestClient.SendAsync()` was called, and these were the only ones that were creating a stream that needed to be closed. All other streams would be coming in from the integrating app.

src/Discord.Net.Rest/Net/DefaultRestClient.cs
ben-reilly, line 1: I recommend reviewing this file with whitespace changes ignored.

src/Discord.Net.Rest/Net/DefaultRestClient.cs
```diff
@@ -101,62 +101,68 @@ public void SetCancelToken(CancellationToken cancelToken)
...
-                if (requestHeaders != null)
-                    foreach (var header in requestHeaders)
-                        restRequest.Headers.Add(header.Key, header.Value);
-                var content = new MultipartFormDataContent("Upload----" + DateTime.Now.ToString(CultureInfo.InvariantCulture));
-                MemoryStream memoryStream = null;
-                if (multipartParams != null)
+                if (stream.CanSeek)
+                {
+                    // Reset back to the beginning; it may have been used elsewhere or in a previous request.
+                    stream.Position = 0;
```
ben-reilly, line 128: I debated between whether to seek back to the beginning before or after the request is processed. I decided to do it before because I think in general we can't assume that a stream has been reset to the beginning, and the integrating app shouldn't assume that about its own streams either.

src/Discord.Net.Rest/Net/DefaultRestClient.cs
```diff
@@ -101,62 +101,68 @@ public void SetCancelToken(CancellationToken cancelToken)
...
+                }
+
+#pragma warning disable IDISP004
+                return new StreamContent(stream);
+#pragma warning restore IDISP004
+            }
+
+            foreach (var p in multipartParams ?? ImmutableDictionary<string, object>.Empty)
```
ben-reilly, line 136: A small thing, but I figured we could reduce the indentation a bit by dropping the `if null` wrapper around this whole block.

src/Discord.Net.Rest/Net/DefaultRestClient.cs
```diff
@@ -101,62 +101,68 @@ public void SetCancelToken(CancellationToken cancelToken)
...
-                                    var stream = fileValue.Stream;
-                                    if (!stream.CanSeek)
-                                    {
-                                        memoryStream = new MemoryStream();
-                                        await stream.CopyToAsync(memoryStream).ConfigureAwait(false);
-                                        memoryStream.Position = 0;
-#pragma warning disable IDISP001
-                                        stream = memoryStream;
-#pragma warning restore IDISP001
-                                    }
```
ben-reilly, line 137: I got rid of this whole `MemoryStream` logic intentionally because I don't think it gains us anything.

If the request is only performed once, then the initial `Stream` will be read fully and we wouldn't have benefited from first copying it into a `MemoryStream`.

If the integrating app tries a second request with the same stream, or the first request fails with a 429 and is retried, then copying the stream a second time won't be possible. In other words, if the incoming stream can't be seeked, then it can't be reused for future requests, and we're gonna be in a pickle here no matter what.

ben-reilly (commented):
Adding some notes on my changes to DefaultRestClient.

src/Discord.Net.Rest/Entities/Interactions/Modals/RestModal.cs
dimson-n:
```suggestion
            using var fileStream = !string.IsNullOrEmpty(filePath) ? new MemoryStream(File.ReadAllBytes(filePath), false) : null;
```

ben-reilly: Thank you! I mis-copied this line.

src/Discord.Net.Rest/Entities/Guilds/RestGuild.cs
```diff
@@ -1071,7 +1071,7 @@ public async Task<CustomSticker> CreateStickerAsync(string name, Image image, IE
...
             return CreateStickerAsync(name, fs, Path.GetFileName(fs.Name), tags,  description,options);
```
dimson-n:
```suggestion
            return await CreateStickerAsync(name, fs, Path.GetFileName(fs.Name), tags,  description,options);
```
For correct dispose of file object from previous line.

src/Discord.Net.Rest/Entities/Guilds/RestGuild.cs
```diff
@@ -1071,7 +1071,7 @@ public async Task<CustomSticker> CreateStickerAsync(string name, Image image, IE
         public Task<CustomSticker> CreateStickerAsync(string name,  string path, IEnumerable<string> tags, string description = null,
```
dimson-n:
```suggestion
        public async Task<CustomSticker> CreateStickerAsync(string name,  string path, IEnumerable<string> tags, string description = null,
```
And `async` for `await`.

dimson-n (changes requested):

src/Discord.Net.Rest/Entities/Guilds/RestGuild.cs
ben-reilly: Thank you, and good catch. The other file needed it too.

ben-reilly:
For what it's worth, I've been using this change in my app for the past week or so and it's been working great (at least when paired with my other PR, #2642). No more send errors after sending tens of thousands of messages; only a small percentage of those contains files to upload, but they are constantly being rate limited and that combo used to guarantee failure.

csmir (approved):

PR #2676, opened 2023-04-27
Allow specifying list of user ids in SocketGuild.DownloadUsersAsync

compujuckel:
Discord supports specifying a list of user IDs in their [Request Guild Members](https://discord.com/developers/docs/topics/gateway-events#request-guild-members) event, but Discord.NET currently does not expose this functionality.

I'm trying to use Discord.NET in a large server (>600k members) and my bot is interested in only a few (100-10,000) of those. Using `DownloadUsersAsync` to get all users takes very long and also shoots up RAM usage to >2GB

This PR is not really ready to merge (the changes work fine for my use case though), I'm mainly looking to get some feedback on what is needed to get this feature into Discord.NET.

### DIFF
--- a/src/Discord.Net.Core/Entities/Guilds/IGuild.cs
+++ b/src/Discord.Net.Core/Entities/Guilds/IGuild.cs
@@ -959,6 +959,17 @@ namespace Discord
         /// </returns>
         Task DownloadUsersAsync();
         /// <summary>
+        ///     Downloads specific users for this guild.
+        /// </summary>
+        /// <remarks>
+        ///     This method downloads all users specified in <paramref name="userIds" /> through the Gateway and caches them.
+        /// </remarks>
+        /// <param name="userIds">The list of Discord user IDs to download</param>
+        /// <returns>
+        ///     A task that represents the asynchronous download operation.
+        /// </returns>
+        Task DownloadUsersAsync(IEnumerable<ulong> userIds);
+        /// <summary>
         ///     Prunes inactive users.
         /// </summary>
         /// <remarks>
--- a/src/Discord.Net.Core/Entities/Guilds/IGuild.cs
+++ b/src/Discord.Net.Core/Entities/Guilds/IGuild.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Discord
@@ -958,17 +959,33 @@ namespace Discord
         ///     A task that represents the asynchronous download operation.
         /// </returns>
         Task DownloadUsersAsync();
+
+        /// <summary>
+        ///     Downloads specific users for this guild with the default request timeout.
+        /// </summary>
+        /// <remarks>
+        ///     This method downloads all users specified in <paramref name="userIds" /> through the Gateway and caches them.
+        ///     Consider using <see cref="DownloadUsersAsync(IEnumerable{ulong}, CancellationToken)"/> when downloading a large number of users.
+        /// </remarks>
+        /// <param name="userIds">The list of Discord user IDs to download.</param>
+        /// <returns>
+        ///     A task that represents the asynchronous download operation.
+        /// </returns>
+        /// <exception cref="OperationCanceledException">The timeout has elapsed.</exception>
+        Task DownloadUsersAsync(IEnumerable<ulong> userIds);
+
         /// <summary>
         ///     Downloads specific users for this guild.
         /// </summary>
         /// <remarks>
         ///     This method downloads all users specified in <paramref name="userIds" /> through the Gateway and caches them.
         /// </remarks>
-        /// <param name="userIds">The list of Discord user IDs to download</param>
+        /// <param name="userIds">The list of Discord user IDs to download.</param>
+        /// <param name="cancelToken">The cancellation token used to cancel the task.</param>
         /// <returns>
         ///     A task that represents the asynchronous download operation.
         /// </returns>
-        Task DownloadUsersAsync(IEnumerable<ulong> userIds);
+        Task DownloadUsersAsync(IEnumerable<ulong> userIds, CancellationToken cancelToken);
         /// <summary>
         ///     Prunes inactive users.
         /// </summary>
--- a/src/Discord.Net.WebSocket/API/Gateway/RequestMembersParams.cs
+++ b/src/Discord.Net.WebSocket/API/Gateway/RequestMembersParams.cs
@@ -6,15 +6,17 @@ namespace Discord.API.Gateway
     [JsonObject(MemberSerialization = MemberSerialization.OptIn)]
     internal class RequestMembersParams
     {
-        [JsonProperty("query")]
-        public string Query { get; set; }
-        [JsonProperty("limit")]
-        public int Limit { get; set; }
         [JsonProperty("guild_id")]
         public ulong GuildId { get; set; }
+        [JsonProperty("query")]
+        public Optional<string> Query { get; set; }
+        [JsonProperty("limit")]
+        public int Limit { get; set; }
+        [JsonProperty("presences")]
+        public Optional<bool> Presences { get; set; }
         [JsonProperty("user_ids")]
-        public IEnumerable<ulong> UserIds { get; set; }
+        public Optional<IEnumerable<ulong>> UserIds { get; set; }
         [JsonProperty("nonce")]
-        public string Nonce { get; set; }
+        public Optional<string> Nonce { get; set; }
     }
 }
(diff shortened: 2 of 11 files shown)


src/Discord.Net.WebSocket/API/Gateway/RequestMembersParams.cs
```diff
@@ -10,8 +10,11 @@ internal class RequestMembersParams
...
         [JsonProperty("limit")]
         public int Limit { get; set; }
-
         [JsonProperty("guild_id")]
-        public IEnumerable<ulong> GuildIds { get; set; }
+        public ulong GuildId { get; set; }
+        [JsonProperty("user_ids")]
+        public IEnumerable<ulong> UserIds { get; set; }
+        [JsonProperty("nonce")]
+        public string Nonce { get; set; }
```
Misha-133: According to [discord docs](https://discord.com/developers/docs/topics/gateway-events#request-guild-members-request-guild-members-structure) `query`, `user_ids` and `nonce` are optional parameters.
Consider wrapping them into `Optional<>`

Additionally, the `presences` field seems to be also missing & should be also added

Misha-133 (changes requested):

src/Discord.Net.WebSocket/API/Gateway/RequestMembersParams.cs
compujuckel: done

quinchs:
I like this idea. @Misha-133 can you stress test this PR and see what you can find for bugs/bottlenecks?

src/Discord.Net.WebSocket/DiscordShardedClient.cs
```diff
@@ -381,6 +381,24 @@ public override async Task DownloadUsersAsync(IEnumerable<IGuild> guilds)
             }
         }
 
+        /// <inheritdoc />
+        /// <exception cref="ArgumentNullException"><paramref name="guild"/> is <see langword="null"/></exception>
+        public override async Task DownloadUsersAsync(IGuild guild, IEnumerable<ulong> userIds)
+        {
+            if (guild == null)
+                throw new ArgumentNullException(nameof(guild));
```
Misha-133:
```suggestion            
            Preconditions.NotNull(guild, nameof(guild));
```

Misha-133:
```cs
        await SocketGuild.DownloadUsersAsync(Array.Empty<ulong>());
```
~~Supplying an empty collection of users to download results in a deadlock~~

UPD: awaiting the method in a event handler causes a deadlock on the gateway thread

Misha-133 (changes requested):

Misha-133:
@quinchs Should `DownloadUsersAsync` methods block the thread until users are actually downloaded?
This causes deadlock if the method is executed in a event handler

`DownloadUsersAsync` method that we had before also has this issue. Shouldn't this be at least documented? Or maybe blocking should be tied to a `bool` parameter

\_\_\_\_\_\_\_\_\_
Everything seems to work fine otherwise

src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -627,30 +631,36 @@ private async Task ProcessUserDownloadsAsync(IEnumerable<SocketGuild> guilds)
...
-                for (int j = 0; j < count; j++, k++)
+                var socketGuild = GetGuild(guild.Id);
+                if (socketGuild != null)
                 {
-                    var guild = cachedGuilds[k];
-                    batchIds[j] = guild.Id;
-                    batchTasks[j] = guild.DownloaderPromise;
+                    await ProcessUserDownloadsAsync(socketGuild, userIds).ConfigureAwait(false);
                 }
+            }
```
Misha-133: This should log an error if the connection state is not `Connected` instead of silently failing

compujuckel:
Once thing that's currently missing from this PR is this:
> Requesting user_ids will continue to be limited to returning 100 members

so this must be chunked when requesting more than 100 members at once

src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -1410,6 +1420,13 @@ private async Task ProcessMessageAsync(GatewayOpCode opCode, int? seq, string ty
                                             guild.CompleteDownloadUsers();
                                             await TimedInvokeAsync(_guildMembersDownloadedEvent, nameof(GuildMembersDownloaded), guild).ConfigureAwait(false);
                                         }
+
+                                        if (data.Nonce.IsSpecified
+                                            && data.ChunkIndex + 1 >= data.ChunkCount
+                                            && _guildMembersRequestTasks.TryRemove(data.Nonce.Value, out var tcs))
```
Misha-133, line 1442: This being the only way of removing request tasks may result in a memory leak if client misses the event (due to temporary connection loss, discord requesting a reconnect, etc.)
I'm not exactly sure what's the best way around this, but I'd recommend adding a timeout to tasks so it retries/fails the task if the client doesn't receive the member chuck in X amount of time. This would also solve the issue of possible deadlock caused by this

src/Discord.Net.WebSocket/DiscordSocketClient.cs
compujuckel: I can add that, I just oriented myself on what the existing `DownloadUsersAsync(IEnumerable<IGuild> guilds)` is doing, which also silently fails.
What would you recommend? Throwing an exception or logging an error?

src/Discord.Net.WebSocket/DiscordSocketClient.cs
compujuckel: What do you think about allowing users to pass a CancellationToken in `DownloadUsersAsync(IEnumerable<ulong> userIds)`? A fixed timeout seems a bit too inflexible IMHO, since downloading >1k users will actually take a while (=> more than a few seconds). If no CancellationToken is given, we could apply a default timeout

src/Discord.Net.WebSocket/DiscordSocketClient.cs
Misha-133: I'd say throwing an exception would be better in this case

src/Discord.Net.WebSocket/DiscordSocketClient.cs
compujuckel: Added a default request timeout and a way for users to supply their own cancellation token.

src/Discord.Net.WebSocket/DiscordSocketClient.cs
```diff
@@ -627,29 +631,51 @@ private async Task ProcessUserDownloadsAsync(IEnumerable<SocketGuild> guilds)
...
+        private async Task ProcessUserDownloadsAsync(SocketGuild guild, IEnumerable<ulong> userIds, CancellationToken cancelToken = default)
+        {
+            var nonce = Interlocked.Increment(ref _guildMembersRequestCounter).ToString();
+            var tcs = new TaskCompletionSource<bool>();
+            using var registration = cancelToken.Register(() => tcs.TrySetCanceled());
+            _guildMembersRequestTasks.TryAdd(nonce, tcs);
+            try
+            {
+                await ApiClient.SendRequestMembersAsync(guild.Id, userIds, nonce).ConfigureAwait(false);
+                await tcs.Task.ConfigureAwait(false);
```
quinchs, line 673: the cancellation token should be integrated with the await of the tcs

quinchs (changes requested):

src/Discord.Net.WebSocket/DiscordSocketClient.cs
compujuckel: that's done a few lines above with the cancellation token registration, unless I'm missing something

src/Discord.Net.Core/Entities/Guilds/IGuild.cs
```diff
@@ -958,6 +959,33 @@ public interface IGuild : IDeletable, ISnowflakeEntity
...
+        /// <remarks>
+        ///     This method downloads all users specified in <paramref name="userIds" /> through the Gateway and caches them.
+        ///     Consider using <see cref="DownloadUsersAsync(IEnumerable{ulong}, CancellationToken)"/> when downloading a large number of users.
+        /// </remarks>
+        /// <param name="userIds">The list of Discord user IDs to download.</param>
+        /// <returns>
+        ///     A task that represents the asynchronous download operation.
+        /// </returns>
+        /// <exception cref="OperationCanceledException">The timeout has elapsed.</exception>
+        Task DownloadUsersAsync(IEnumerable<ulong> userIds);
```
quinchs, line 975: instead of defining 2 methods with the only difference being a cancel token, why not make the token parameter optional?

quinchs (changes requested):

PR #2680, opened 2023-05-11
Revert "Optimize Add/RemoveRolesAsync methods (#2614)"

### LINKED ISSUE #2655: [Bug]: AddRoleAsync() and AddRolesAsync() removes roles after adding them
### Check The Docs

### Verify Issue Source

### Check your intents

### Description

`AddRoleAsync()` and `AddRolesAsync()` removes roles after adding them when using Discord.Net 3.10.0.

After downgrading to Discord.Net 3.9.0, `AddRoleAsync()` and `AddRolesAsync()` functions as intended.

### Version

3.10.0

### Working Version

3.9.0

### Logs

n/a

### Sample

```csharp
public async Task HandleMenuInteraction(SocketMessageComponent messageComponent)
{
    switch(messageComponent.Data.CustomId)
    {
        case "role-selection":
            Console.WriteLine($"{messageComponent.User.Username} has changed their role subscriptions.");

            var values = messageComponent.Data.Values.ToList<string>();

            Console.WriteLine($"{messageComponent.User.Username} has selected the following role subscriptions:");

            foreach(var v in values) {
                Console.WriteLine(v);
            }
            
            var user = (IGuildUser) messageComponent.User;

            if(user != null)
            {
                List<string> allowedRoles = new List<string>()
                {
                    "Games",
                    "Technology",
                    "Programming",
                    "Art",
                    "Music"
                };

                List<ulong> addedRoles = new List<ulong>();
                List<ulong> removedRoles = new List<ulong>();

                string roleSubscriptions = "";

                foreach(var i in allowedRoles) {
                    try {
                        var role = user.Guild.Roles.First(x => x.Name == i);

                        if(values.Contains(role.Name)) {
                            addedRoles.Add(role.Id);
                            roleSubscriptions += $@"
:bell: {role.Name}";
                            Console.WriteLine($"{role.Name} ({role.Id}) was added to {messageComponent.User.Username}.");
                        }
                        else {
                            removedRoles.Add(role.Id);
                            roleSubscriptions += $@"
:no_entry_sign: {role.Name}";
                            Console.WriteLine($"{role.Name} ({role.Id}) was removed from {messageComponent.User.Username}.");
                        }
                    }
                    catch(Exception ex) {
                        Console.WriteLine(ex);
                    }
                }

                await user.AddRolesAsync(addedRoles);
                await user.RemoveRolesAsync(removedRoles);

                await messageComponent.DeferAsync();
                await messageComponent.ModifyOriginalResponseAsync(x => x.Content = $"{user.Mention} is now subscribed to the following channels: {roleSubscriptions}");
            }
            break;
    }
}
```

### Packages

Discord.Net V3.9.0

Misha-133:
this pr resolves #2655

### DIFF
--- a/src/Discord.Net.Rest/Entities/Users/UserHelper.cs
+++ b/src/Discord.Net.Rest/Entities/Users/UserHelper.cs
@@ -81,18 +81,14 @@ namespace Discord.Rest
 
         public static async Task AddRolesAsync(IGuildUser user, BaseDiscordClient client, IEnumerable<ulong> roleIds, RequestOptions options)
         {
-            await client.ApiClient.ModifyGuildMemberAsync(user.GuildId, user.Id, args: new()
-            {
-                RoleIds = user.RoleIds.Except(new[] { user.Guild.Id }).Concat(roleIds).Distinct().ToArray()
-            }, options);
+            foreach (var roleId in roleIds)
+                await client.ApiClient.AddRoleAsync(user.Guild.Id, user.Id, roleId, options).ConfigureAwait(false);
         }
 
         public static async Task RemoveRolesAsync(IGuildUser user, BaseDiscordClient client, IEnumerable<ulong> roleIds, RequestOptions options)
         {
-            await client.ApiClient.ModifyGuildMemberAsync(user.GuildId, user.Id, args: new()
-            {
-                RoleIds = user.RoleIds.Except(new[] { user.Guild.Id }).Except(roleIds).ToArray()
-            }, options);
+            foreach (var roleId in roleIds)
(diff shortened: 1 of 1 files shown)


quinchs (approved):

xXBuilderBXx:
This would be better if AddRole() was split up from the user helper AddRoles() function with its own user helper AddRole() 
And change the comment for AddRoles() to say that it overwrites the user's roles and may lead to race conditions with other bots or something along that line.

Misha-133:
> This would be better if AddRole() was split up from the user helper AddRoles() function with its own user helper AddRole() 

It's been like this for couple years now... I'll just leave it in its current state.

> And change the comment for AddRoles() to say that it overwrites the user's roles and may lead to race conditions with other bots or something along that line.

It does not. That's the point of this revert.

csmir (approved):

PR #2681, opened 2023-05-12
Add support for sending the guild in the PresenceUpdated event

SylveonDeko:
This PR adds the guild to the presenceupdated event to know which guild the presence updated event is fired from so you aren't left guessing.

# This is a breaking change

### DIFF
--- a/src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
+++ b/src/Discord.Net.WebSocket/BaseSocketClient.Events.cs
@@ -531,12 +531,12 @@ namespace Discord.WebSocket
         #region Presence
 
         /// <summary> Fired when a users presence is updated. </summary>
-        public event Func<SocketUser, SocketPresence, SocketPresence, Task> PresenceUpdated
+        public event Func<SocketUser, SocketGuild, SocketPresence, SocketPresence, Task> PresenceUpdated
         {
             add { _presenceUpdated.Add(value); }
             remove { _presenceUpdated.Remove(value); }
         }
-        internal readonly AsyncEvent<Func<SocketUser, SocketPresence, SocketPresence, Task>> _presenceUpdated = new AsyncEvent<Func<SocketUser, SocketPresence, SocketPresence, Task>>();
+        internal readonly AsyncEvent<Func<SocketUser, SocketGuild, SocketPresence, SocketPresence, Task>> _presenceUpdated = new AsyncEvent<Func<SocketUser, SocketGuild, SocketPresence, SocketPresence, Task>>();
 
         #endregion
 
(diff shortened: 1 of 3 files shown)


quinchs:
Exposing a null parameter to events is not the solution here... if you want the guild, check whether or not the user is a `SocketGuildUser`, since presence updates are not tied to a guild and can be sent from users in dms etc.

simply, just do:
```cs
if(user is SocketGuildUser guildUser) 
{
    var guild = guildUser.Guild;
}
```

SylveonDeko:
It exposed global not guilduser in the event

SylveonDeko:
And why should we need an extra cast if discord literally gives us this
info?

Misha-133:
> since presence updates are not tied to a guild and can be sent from users in dms etc.
@quinchs 

According to [documentation](https://discord.com/developers/docs/topics/gateway-events#presence-update) they actually are tied to a guild; and `guild_id` field in event payload is neither optional or nullable.
And from my testing I could not get presence updates for dms

Misha-133:
I'd suggest changing the `SocketUser` parameter in the event to `SocketGuildUser` instead of adding a guild param
This way it wouldn't be a breaking change, since the `SocketGuildUser` inherits from `SocketUser`
and it would eliminate the confusion & need for unneeded extra cast

quinchs:
> It exposed global not guilduser in the event

*if* the guid id is present, the client will get the guild from the cache, then will set the events `user` to `guild.GetUser()`[^1]

> And why should we need an extra cast if discord literally gives us this
info?

because this event is not guild exclusive, the event *can* and *does* fire when presence is updated for a user you do not share a guild with, therefor we cant make it guild exclusive without either
A: making the guild parameter nullable (like your pr does)
B: splitting the events up into a guild specific and non guild specific presence event.
With A: we want to avoid nullable parameters for events, this is just better DX. For B: v2 had presence tied to `UserUpdated` and `GuildUserUpdated`, it was changed to be one single event in v3[^2] do lower the event frequency of `UserUpdated` and `GuildUserUpdated`. Sure an argument can be make to split this event up even more, but why do that when you can just check the user type.

ultimately, dnet uses the cast-check pattern everywhere, and it just makes sense to use it here.

 > According to [documentation](https://discord.com/developers/docs/topics/gateway-events#presence-update) they actually are tied to a guild; and `guild_id` field in event payload is neither optional or nullable.
And from my testing I could not get presence updates for dms

Sure, but what happens if dnet doesn't have the guild cached? the event should still be fired with the global user if its present, although the code doesn't do this (@Misha-133 another feature for you to add :)). 
When this event was split, bots used to receive presence updates from non-guild users too, thats why the event was written like this. In V4, we can defiantly make a change to make the user parameter a `SocketGuildUser`.

[^1]: https://github.com/SylveonDeko/Discord.Net/blob/e2dfd2d6281784ac32b526aa8923d21107637ead/src/Discord.Net.WebSocket/DiscordSocketClient.cs#L1996

[^2]: https://github.com/discord-net/Discord.Net/pull/1945

Issue #2686, opened 2023-05-31
[Bug]: Unable to register more than 1 automod action


ThePheggHerself:
### Check The Docs

- [X] I double checked the docs and couldn't find any useful information.

### Verify Issue Source

- [X] I verified the issue was caused by Discord.Net.

### Check your intents

- [X] I double checked that I have the required intents.

### Description

When trying to create a new automod rule for a guild, if it has more than 1 action listed, it will cause an error stating that it must have atleast 1 action. Along with this, you can also attempt to register an automod rule with 0 actions, resulting in an invalid form body error from the discord API.

### Version

3.10.0

### Logs

[31/05/2023 14:48] ERROR | System.ArgumentException: Auto moderation rule must have at least 1 action (Parameter 'Actions')
   at Discord.Preconditions.AtLeast(Int32 obj, Int32 value, String name, String msg)
   at Discord.Rest.GuildHelper.CreateAutoModRuleAsync(IGuild guild, Action`1 func, BaseDiscordClient client, RequestOptions options)
   at Discord.WebSocket.SocketGuild.CreateAutoModRuleAsync(Action`1 props, RequestOptions options)
   at CorruptionBot.Commands.SettingsCommand.FilterSubcommand.InviteFilter(ITextChannel alertChan) in C:\Users\theph\Documents\Programming\CorruptionBot\CorruptionBot\Commands\SettingsCommand.cs:line 77

### Sample

```cs
[SlashCommand("filter", "Automod testing command")]
public async Task InviteFilter([Summary("Channel", "Channel for the rule to send an alert to")] ITextChannel alertChan)
{
	await Context.Guild.CreateAutoModRuleAsync(x =>
	{
		x.Name = "Test Filter";
		x.TriggerType = AutoModTriggerType.Keyword;
		x.KeywordFilter = new string[] { "cat", "dog", "foo", "bar" };
		x.Actions = new AutoModRuleActionProperties[]{
			new AutoModRuleActionProperties { Type = AutoModActionType.BlockMessage, CustomMessage = "You cannot use these words" },
			new AutoModRuleActionProperties { Type = AutoModActionType.SendAlertMessage, ChannelId = alertChan.Id }
		};
	});
	await RespondAsync("Your new automod rule has been created");
}
```

### Packages

Discord.Net 2.4.0
Discord.InteractivityAddon 2.4.0
Microsoft.Extensions.DependencyInjection 7.0.0
MySql.Data 8.0.31

[...]

PR #2687, opened 2023-05-31
Allow more than 1 action to be registered per automod rule

ThePheggHerself:
Fixes issue #2686 

Swaps the 1st and 2nd parameter around to allow more than 1 action to be registered per rule (And prevents any attempts to register 0 actions per rule)


quinchs (approved):
@Misha-133 you have to test your code works, bugs like this shouldn't get included into dev..

PR #2696, opened 2023-06-07
[Feature] Support for the new username system

### LINKED ISSUE #2693: [Bug]: New usernames are displayed with #0000 discriminator
### Check The Docs

### Verify Issue Source

### Check your intents

### Description

When I convert a SocketUser (who got the new, discriminatorless username) to a string, there is `#0000` at the end.

### Version

3.10.0

### Sample

```cs
SocketUser user; // assume this an account with the new username
Console.WriteLine(user.ToString()); // someone#0000
```

Misha-133:
This PR brings support for the new username system introduced by discord.

### Changes
- update all `ToString()` methods of user entities & their debugger display properties to use the new unique username instead of `username#discriminator` pattern if discriminator value is equals to 0 - the placeholder value.
- made `discriminator` parameter of `BaseSocketClient.GetUser()` method optional
- update `EmbedBuilderExtensions.WithAuthor()`
- update `MentionUtils`
- Add `GlobalName` property to user entities
- Add a new CDN method with support for new default avatar calculation
- Update `UserTypeReader` in `Discord.Commands` to search for `GlobalName`
#### Misc
- update `MentionUtils` to use `<@id>` user mention syntax instead of deprecated `<@!id>` ([docs](https://discord.com/developers/docs/reference#message-formatting-formats))

- closes #2693

*P.S. I think I've found all places that need to be updated for the new system, pls correct me if I missed any*

### DIFF
--- a/src/Discord.Net.Core/Entities/Users/IUser.cs
+++ b/src/Discord.Net.Core/Entities/Users/IUser.cs
@@ -47,11 +47,11 @@ namespace Discord
         /// </returns>
         string GetDefaultAvatarUrl();
         /// <summary>
-        ///     Gets the per-username unique ID for this user. This will return "0000" for users, who have migrated to new username system.
+        ///     Gets the per-username unique ID for this user. This will return "0000" for users who have migrated to new username system.
         /// </summary>
         string Discriminator { get; }
         /// <summary>
-        ///     Gets the per-username unique ID for this user. This will return 0 for users, who have migrated to new username system.
+        ///     Gets the per-username unique ID for this user. This will return 0 for users who have migrated to new username system.
         /// </summary>
         ushort DiscriminatorValue { get; }
         /// <summary>
--- a/src/Discord.Net.Core/Entities/Users/IUser.cs
+++ b/src/Discord.Net.Core/Entities/Users/IUser.cs
@@ -47,7 +47,7 @@ namespace Discord
         /// </returns>
         string GetDefaultAvatarUrl();
         /// <summary>
-        ///     Gets the per-username unique ID for this user. This will return 0 for users, who have migrated to new username system.
+        ///     Gets the per-username unique ID for this user. This will return "0000" for users, who have migrated to new username system.
         /// </summary>
         string Discriminator { get; }
         /// <summary>
--- a/src/Discord.Net.Core/Entities/Users/IUser.cs
+++ b/src/Discord.Net.Core/Entities/Users/IUser.cs
@@ -47,11 +47,11 @@ namespace Discord
         /// </returns>
         string GetDefaultAvatarUrl();
         /// <summary>
-        ///     Gets the per-username unique ID for this user.
+        ///     Gets the per-username unique ID for this user. This will return 0 for users, who have migrated to new username system.
         /// </summary>
         string Discriminator { get; }
         /// <summary>
-        ///     Gets the per-username unique ID for this user.
+        ///     Gets the per-username unique ID for this user. This will return 0 for users, who have migrated to new username system.
         /// </summary>
         ushort DiscriminatorValue { get; }
         /// <summary>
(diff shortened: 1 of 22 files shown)


test/Discord.Net.Tests.Unit/MentionUtilsTests.cs
```diff
@@ -16,10 +16,9 @@ public class MentionUtilsTests
         [Fact]
         public void MentionUser()
         {
-            Assert.Equal("<@!123>", MentionUtils.MentionUser(123u));
-            Assert.Equal("<@!123>", MentionUtils.MentionUser("123"));
-            Assert.Equal("<@!123>", MentionUtils.MentionUser("123", true));
-            Assert.Equal("<@123>", MentionUtils.MentionUser("123", false));
+            Assert.Equal("<@123>", MentionUtils.MentionUser(123u));
+            Assert.Equal("<@123>", MentionUtils.MentionUser("123"));
+            Assert.Equal("<@123>", MentionUtils.MentionUser("123"));
```
dimson-n:
```suggestion
```

Panthr75:
Might want to deprecate `Discriminator` and `DiscriminatorValue`, as most users have already transferred over to the new username system

d4n3436:
> Might want to deprecate `Discriminator` and `DiscriminatorValue`, as most users have already transferred over to the new username system

Bots will remain on the old username system
https://support-dev.discord.com/hc/en-us/articles/13667755828631#h_01H26BDRT8ASH1YVKG6HMGZ1J2

src/Discord.Net.Core/Entities/Users/IUser.cs
```diff
@@ -47,11 +47,11 @@ public interface IUser : ISnowflakeEntity, IMentionable, IPresence
         /// </returns>
         string GetDefaultAvatarUrl();
         /// <summary>
-        ///     Gets the per-username unique ID for this user.
+        ///     Gets the per-username unique ID for this user. This will return "0000" for users, who have migrated to new username system.
```
quinchs:
```suggestion
        ///     Gets the per-username unique ID for this user. This will return "0000" for users who have migrated to new username system.
```

src/Discord.Net.Core/Entities/Users/IUser.cs
```diff
@@ -47,11 +47,11 @@ public interface IUser : ISnowflakeEntity, IMentionable, IPresence
...
         /// </summary>
         string Discriminator { get; }
         /// <summary>
-        ///     Gets the per-username unique ID for this user.
+        ///     Gets the per-username unique ID for this user. This will return 0 for users, who have migrated to new username system.
```
quinchs:
```suggestion
        ///     Gets the per-username unique ID for this user. This will return 0 for users who have migrated to new username system.
```

src/Discord.Net.Core/Format.cs
```diff
@@ -108,16 +108,18 @@ public static string StripMarkDown(string text)
         }
 
         /// <summary>
-        ///     Formats a user's username + discriminator.
+        ///     Formats a user's username + discriminator. Returns the username if the user has migrated to new username system.
```
quinchs:
```suggestion
        ///     Formats a user's username and optional discriminator.
```

src/Discord.Net.Core/Format.cs
```diff
@@ -108,16 +108,18 @@ public static string StripMarkDown(string text)
...
         /// </summary>
         /// <param name="doBidirectional">To format the string in bidirectional unicode or not</param>
         /// <param name="user">The user whose username and discriminator to format</param>
         /// <returns>The username + discriminator</returns>
```
quinchs:
```suggestion
        /// <returns>The username + optional discriminator.</returns>
```

quinchs (changes requested):

Panthr75 (changes requested):
Update `SocketThreadUser.DisplayName` to match `SocketGuildUser.DisplayName` by including a nullability check with `GlobalName` before `Username`

quinchs (approved):

PR #2724, opened 2023-07-07
[Feature] Channel icon emojis

Misha-133:
This PR adds support for channel emojis

### Changes
- Add `IconEmoji` property to `IGuildChannel`
- add `IconEmoji` to modify channel properties


src/Discord.Net.Rest/Entities/Channels/ChannelHelper.cs
```diff
@@ -39,9 +39,17 @@ internal static class ChannelHelper
                     }).ToArray()
                     : Optional.Create<API.Overwrite[]>(),
                 Flags = args.Flags.GetValueOrDefault(),
+                IconEmoji = args.IconEmoji.IsSpecified
+                    ? args.IconEmoji.Value is Emote emote
+                        ? new API.Emoji { Id = emote.Id, Name = emote.Name }
+                        : args.IconEmoji.GetValueOrDefault(null) is Emoji emoji
```
dimson-n: Looks like default value never be used here because `IconEmoji.IsSpecified` already checked and passing through `true` branch.

Misha-133: Hmm, yeah, nice catch. Idk why I ended up with `GetValueOrDefault`, seems to work fine without it.

Misha-133:
This property seems to be no longer used so i'll close this PR.

PR #2725, opened 2023-07-07
[Feature] Media channel support

Misha-133:
[docs](https://github.com/discord/discord-api-docs/pull/6232)

This PR adds support for the new media channels. 

## Warning
This feature is not released & fully documented yet, so use this at your own risk as things might change at any moment.

### Changes
- add `IMediaChannel`, rest & socket entities.
- add new channel flag
- add `IsThumbnail` property to `FileAttachment` - one can be set for an image used as a thumbnail in a media channel post
- update everything to support new `ChannelType`
- **misc** Add some missing forum channel related methods to `IGuild`

#### This PR is based on PR #2720

### DIFF
--- a/src/Discord.Net.Rest/Entities/Guilds/RestGuild.cs
+++ b/src/Discord.Net.Rest/Entities/Guilds/RestGuild.cs
@@ -447,6 +447,64 @@ namespace Discord.Rest
         }
 
         /// <summary>
+        ///     Gets a forum channel in this guild.
+        /// </summary>
+        /// <param name="id">The snowflake identifier for the forum channel.</param>
+        /// <param name="options">The options to be used when sending the request.</param>
+        /// <returns>
+        ///     A task that represents the asynchronous get operation. The task result contains the text channel
+        ///     associated with the specified <paramref name="id"/>; <see langword="null"/> if none is found.
+        /// </returns>
+        public async Task<RestForumChannel> GetForumChannelAsync(ulong id, RequestOptions options = null)
+        {
+            var channel = await GuildHelper.GetChannelAsync(this, Discord, id, options).ConfigureAwait(false);
+            return channel as RestForumChannel;
+        }
+
+        /// <summary>
+        ///     Gets a collection of all forum channels in this guild.
+        /// </summary>
+        /// <param name="options">The options to be used when sending the request.</param>
+        /// <returns>
+        ///     A task that represents the asynchronous get operation. The task result contains a read-only collection of
+        ///     forum channels found within this guild.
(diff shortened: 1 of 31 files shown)


src/Discord.Net.Rest/Entities/Guilds/RestGuild.cs
```diff
@@ -726,11 +784,24 @@ public Task<RestCategoryChannel> CreateCategoryChannelAsync(string name, Action<
         /// <param name="options">The options to be used when sending the request.</param>
         /// <exception cref="ArgumentNullException"><paramref name="name" /> is <see langword="null"/>.</exception>
         /// <returns>
-        ///     The created category channel.
+        ///     The created forum channel.
```
Panthr75, line 787: Rename `category` on line 780 to `forum`, like here

src/Discord.Net.WebSocket/Entities/Guilds/SocketGuild.cs
```diff
@@ -337,6 +337,15 @@ public IReadOnlyCollection<SocketThreadChannel> ThreadChannels
         public IReadOnlyCollection<SocketForumChannel> ForumChannels
             => Channels.OfType<SocketForumChannel>().ToImmutableArray();
```
Panthr75, line 338: Might want to add a check to not include media channels here

Panthr75:
It would be nice to have a `MediaChannelProperties`, so that you aren't passing `ForumChannelProperties` to a `Create Media Channel` method -- which seems wrong.

src/Discord.Net.Rest/Entities/Guilds/RestGuild.cs
Misha-133: ah yup, nice catch

src/Discord.Net.WebSocket/Entities/Guilds/SocketGuild.cs
Misha-133: This way it matches the behavior of similar methods.
ex: `SocketGuild.TextChannels`, `SocketGuild.VoiceChannels`

Misha-133:
> It would be nice to have a `MediaChannelProperties`, so that you aren't passing `ForumChannelProperties` to a `Create Media Channel` method -- which seems wrong.

I don't see a reason to add more properties, since the media channel does not have any extra features. Also, this matches the `IStageChannel` reusing same `VoiceChannelProperties`

MrCakeSlayer (approved):

quinchs (approved):

PR #2739, opened 2023-07-24
[Refactor] Remove some unnecessary async/await

dimson-n:
There are many `async` functions that doesn't really need to be `async` (e.g. check preconditions and delegate other operations to subsequent call like some helper).

This PR removes some of unnecessary `async`/`await` usage in simple cases.

That may cause slightly improve CPU and memory usage and allows to compiler generate more efficient code. Yes, this is micro-optimizations but every little performance issue touches each library user.


Misha-133 (approved):
lgtm

Waterball12:
You lose some information for debugging. seems to be fine for overload but in the PR is not always the case. Should consider how worth is `slightly improve CPU and memory usage`.

https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task

dimson-n:
Good remark, but i'm prefer performance in case of frameworks like D.Net is (just imho).
Exception wrapping guarantees may be matter, but almost all tasks awaited inplace here as i can see.

PR #2758, opened 2023-08-13
[Fix]: Slash Command Localizations lost after AddCommandsToGuildAsync and deleteMissing = false

### LINKED ISSUE #2757: [Bug]: Slash Command Localizations lost after AddCommandsToGuildAsync and deleteMissing = false
### Check The Docs

### Verify Issue Source

### Check your intents

### Description

When adding slash commands to a guild via `AddCommandsToGuildAsync` with the `deleteMissing` parameter set to `false` (probably also with `AddCommandsGloballyAsync` but i haven't tried that yet), the localizations of the existing slash commands are lost.

The cause is (most likely) [this call here](https://github.com/discord-net/Discord.Net/blob/7bfeb98265db66f646e429e7d1df1c33b3fadbb1/src/Discord.Net.Interactions/InteractionService.cs#L484C49-L484C85): the parameter `withLocalizations` is not set and default is `false`.

Will verify if this is indeed the cause and put a PR together with a fix for `AddCommandsToGuild`/`GloballyAsync` and other places that come to my mind, maybe `RegisterCommandsToGuild`/`GloballyAsync` are also affected.

### Version

3.12.0

### Working Version

Don't know if it was working before...

### Logs

```shell
don't have any logs
```

### Sample

_No response_

### Packages

Fergun.Interactive v1.7.3

### Environment

- **OS**: Windows 10
- **Architecture**: x64
- **SDK**: .NET 7

MaxEtMoritz:
### Description

Fetch localizations when adding commands with `deleteMissing` = `false` to prevent them getting lost.

### Changes

- add missing `withLocalizations` parameter with value `true` to the calls to `RestClient.GetGlobal`/`GuildApplicationCommands` calls.

### Related Issues

- resolves #2757

### DIFF
--- a/src/Discord.Net.Interactions/InteractionService.cs
+++ b/src/Discord.Net.Interactions/InteractionService.cs
@@ -392,7 +392,7 @@ namespace Discord.Interactions
             if (!deleteMissing)
             {
 
-                var existing = await RestClient.GetGuildApplicationCommands(guildId).ConfigureAwait(false);
+                var existing = await RestClient.GetGuildApplicationCommands(guildId, true).ConfigureAwait(false);
                 var missing = existing.Where(x => !props.Any(y => y.Name.IsSpecified && y.Name.Value == x.Name));
                 props.AddRange(missing.Select(x => x.ToApplicationCommandProps()));
             }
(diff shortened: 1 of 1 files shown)


DeclanFrampton:
I would suggest introducing  optional localization parameter's rather than forcing true on the localization params.

MaxEtMoritz:
> I would suggest introducing optional localization parameter's rather than forcing true on the localization params.

Yeah, could do that, but it is not clear for me what the use case of the current behavior is.
Why would someone want to add Slash commands without deleting the existing ones, but discard all existing localizations in this operation?

quinchs (approved):

PR #2769, opened 2023-08-28
[Fix] Add missing parameters to `RespondWithModalAsync` methods and implement missing overloads

zobweyt:
### Description
This PR adds missing implementation of modal modification in module bases.

### Changes
- Add missing `modifyModal` parameters to the `RespondWithModalAsync<T>` methods. (b7ee974)
- Add missing `RespondWithModalAsync<T>` overload to module bases. (e63665d)
- Fix inactive `options` parameter in the `RespondWithModalAsync` method. (46bcd72)
- Remove code duplication in `IDiscordInteractionExtensions` by using the `modalInfo.FromModal` method. (42f448b)
- Remove code duplication in `RestInteractionModuleBase` by extracting a `HandleInteractionAsync` method. (1a788d4)

### DIFF
--- a/src/Discord.Net.Interactions/Extensions/IDiscordInteractionExtensions.cs
+++ b/src/Discord.Net.Interactions/Extensions/IDiscordInteractionExtensions.cs
@@ -84,23 +84,8 @@ namespace Discord.Interactions
 
         private static async Task SendModalResponseAsync(IDiscordInteraction interaction, string customId, ModalInfo modalInfo, RequestOptions options = null, Action<ModalBuilder> modifyModal = null)
         {
-            var builder = new ModalBuilder(modalInfo.Title, customId);
-
-            foreach (var input in modalInfo.Components)
-                switch (input)
-                {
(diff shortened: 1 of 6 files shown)


Misha-133:
Could you also add [this overload](https://github.com/discord-net/Discord.Net/blob/dev/src/Discord.Net.Interactions/Extensions/IDiscordInteractionExtensions.cs#L57C9-L83)?

zobweyt:
should the `modifyModal` parameter come before `options`? almost everywhere it's ordered last and least used, also there will be no need to manually specify the parameter name or pass `null` to the `options`

Misha-133:
> should the `modifyModal` parameter come before `options`? almost everywhere it's ordered last and least used, also there will be no need to manually specify the parameter name or pass `null` to the `options`

Yeah, it should
But in some places doing that would introduce a breaking change. So you should order params in logical order where it won't 
break existing code.

Misha-133 (approved):

PR #2770, opened 2023-08-28
[Fix] URL encode timestamps sent as query parameters

Sauceke:
### Description

Methods calling to the `threads/archived/*` endpoints don't escape the '+' character in the `before` query parameter, resulting in a DATE_TIME_TYPE_PARSE error. URL escaping the timestamp solves this.

### Changes

- Update ~all three~ two of the archived thread enumeration methods

### Related Issues

none

EDIT: Sorry, I just realized one of these endpoints expects a different date format. Should have RTFM.
The other 2 methods are still in scope, reverted changes for the third one.
(The third method still seems to be incorrect in one of the branches, but that isn't related to this PR.)

### DIFF
--- a/src/Discord.Net.Rest/DiscordRestApiClient.cs
+++ b/src/Discord.Net.Rest/DiscordRestApiClient.cs
@@ -618,11 +618,13 @@ namespace Discord.API
 
             if (limit.HasValue)
             {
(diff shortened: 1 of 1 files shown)


Sauceke:
Sorry, I just realized one of these methods uses a different date format. Should have RTFM.
The other 2 methods are still in scope, reverted changes for the third one.
(The third method still seems to be incorrect in one of the branches, but that isn't related to this PR.)

Misha-133 (approved):

PR #2774, opened 2023-09-11
[Fix] Some IDiscordClient methods are not properly implemented in clients

Panthr75:
### Description
Fixes some bugs related to `IDiscordClient` methods not being properly implemented in `RestClient`, `SocketClient`, and `ShardedSocketClient` due to `BaseDiscordClient` *"blankly"* implementing them.

### Changes
- Added `CreateGlobalApplicationCommandAsync` and `BulkOverwriteGlobalApplicationCommandsAsync` to `ShardedSocketClient`
- `ShardedSocketClient` properly implements `IDiscordClient.BulkOverwriteGlobalApplicationCommand` and `CreateGlobalApplicationCommand`
- `SocketClient` properly implements `IDiscordClient.BulkOverwriteGlobalApplicationCommand` and `CreateGlobalApplicationCommand`
- `RestClient` properly implements `IDiscordClient.BulkOverwriteGlobalApplicationCommand` and `CreateGlobalApplicationCommand`

### Related Issues
- Bulk Overriding Global Commands does not work with `IDiscordClient`

### DIFF
--- a/src/Discord.Net.WebSocket/DiscordShardedClient.cs
+++ b/src/Discord.Net.WebSocket/DiscordShardedClient.cs
@@ -519,6 +519,49 @@ namespace Discord.WebSocket
             client.WebhooksUpdated += (arg1, arg2) => _webhooksUpdated.InvokeAsync(arg1, arg2);
             client.AuditLogCreated += (arg1, arg2) => _auditLogCreated.InvokeAsync(arg1, arg2);
         }
+
+        public async Task<SocketApplicationCommand> CreateGlobalApplicationCommandAsync(ApplicationCommandProperties properties, RequestOptions options = null)
+        {
+            var model = await InteractionHelper.CreateGlobalCommandAsync(this, properties, options).ConfigureAwait(false);
+
+            SocketApplicationCommand entity = null;
+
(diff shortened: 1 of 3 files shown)


src/Discord.Net.WebSocket/DiscordShardedClient.cs
```diff
@@ -519,6 +519,49 @@ private void RegisterEvents(DiscordSocketClient client, bool isPrimary)
...
+
+            foreach (var shard in _shards)
+            {
+                entity = shard.State.GetOrAddCommand(model.Id, (id) => SocketApplicationCommand.Create(shard, model));
+            }
+
+            System.Diagnostics.Debug.Assert(entity != null, "There should be at least one shard to get the entity");
+
+            //Update it in case it was cached
+            entity.Update(model);
```
Misha-133: what would happen if the command is cached in multiple shard states? it would be updated in only one shard with this impl

Panthr75: Updated to all shards. I wonder though if I should return the last entity (like the current implementation).

Misha-133 (approved):

PR #2832, opened 2024-01-22
[Feature] Add Icon size parameter to `GetGuildIconUrl` CDN method

Misha-133:
This PR adds a size parameter to the `GetGuildIconUrl` CDN method


src/Discord.Net.Core/CDN.cs
```diff
@@ -102,11 +102,12 @@ public static string GetDefaultUserAvatarUrl(ulong userId)
         /// </summary>
         /// <param name="guildId">The guild snowflake identifier.</param>
         /// <param name="iconId">The icon identifier.</param>
+        /// <param name="size">The size of the image to return in horizontal pixels. This can be any power of two between 16 and 2048.</param>
```
quinchs, line 105: > This can be any power of two between 16 and 2048.

Should we add a check for this instead of blindly adding it as the size param?

Misha-133: It's the same behavior as in all other endpoints that have a `size` parameter

MrCakeSlayer (approved):

PR #2835, opened 2024-01-29
[Feature] Moar select menu option validations

Misha-133:
This PR adds some additional validations to `SelectMenuOptionBuilder`


orobertofc:
This would have saved my ass

quinchs (approved):

PR #2893, opened 2024-03-28
Add embed parameter for webhooks in SendMessageAsync

akaJuliaan:
### Description
This PR adds a `embed` parameter to the `SendMessageAsync` method for Webhooks, like it is already in the `SendMessageAsync` method for Channels.

### Changes
- Add `embed` parameter
- Update Webhook sample

### DIFF
--- a/samples/WebhookClient/Program.cs
+++ b/samples/WebhookClient/Program.cs
@@ -24,6 +24,6 @@ class Program
 
         // Webhooks are able to send multiple embeds per message
         // As such, your embeds must be passed as a collection.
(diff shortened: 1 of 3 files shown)


akaJuliaan:
The build for the samples logically fails because the new parameter does not yet exist in version 3.13.1 which they are referring to.

SylveonDeko:
Embed parameter is not needed as you can pass a single embed in an embed array into the embeds parameter. Yes it's inconvenient but redundant parameters aren't useful

akaJuliaan:
If redundant parameters are unwanted, why is there the exactly same `embed` parameter for the channels?

SylveonDeko:
Because removing it would be a breaking change.

PR #2910, opened 2024-04-15
Fix: RestDMChannel relies on recipient 

quinchs:
### Summary
The `RestDMChannel` relies on `Recipient` when in actuality it can be unspecified in the case of a user app dm interaction.

### DIFF
--- a/src/Discord.Net.Rest/Entities/Channels/RestDMChannel.cs
+++ b/src/Discord.Net.Rest/Entities/Channels/RestDMChannel.cs
@@ -29,7 +29,9 @@ namespace Discord.Rest
         /// <summary>
         ///     Gets a collection that is the current logged-in user and the recipient.
         /// </summary>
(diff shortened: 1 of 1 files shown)


src/Discord.Net.Rest/Entities/Channels/RestDMChannel.cs
```diff
@@ -29,23 +29,26 @@ public class RestDMChannel : RestChannel, IDMChannel, IRestPrivateChannel, IRest
         /// <summary>
         ///     Gets a collection that is the current logged-in user and the recipient.
         /// </summary>
-        public IReadOnlyCollection<RestUser> Users => ImmutableArray.Create(CurrentUser, Recipient);
+        public IReadOnlyCollection<RestUser> Users => Recipient is null
+            ? Array.Empty<RestUser>()
```
Misha-133:
```suggestion
            ? ImmutableArray<RestUser>.Empty
```

Misha-133 (approved):

PR #2912, opened 2024-04-15
Fix: Don't depend on WebSocket for Interaction service

quinchs:
### Summary
This PR removes the pretty pointless dependency on interaction service, as we want to support full HTTP interactions without needing websockets


src/Discord.Net.Rest/IRestClientProvider.cs
```diff
@@ -0,0 +1,8 @@
+using Discord.Rest;
+
+namespace Discord.Rest;
+
+public interface IRestClientProvider
+{
+    DiscordRestClient RestClient { get; }
+}
```
Misha-133: no xmldoc
911289378234585170 gif

Misha-133 (changes requested):

Liege72 (approved):

PR #2913, opened 2024-04-20
Remove generic exceptions.

### LINKED ISSUE #1456: Do not throw the basic Exception type
In several places in the codebase, the basic Exception type is instantiated upon an error, in some cases with a more derived type as the InnerException.

Please consider not doing this. It makes handling exceptions in BaseDiscordClient.Log more complicated. For the case where there is no InnerException ([e.g.](https://github.com/discord-net/Discord.Net/blob/2c93363653d7935cd142b664b7abc63fd2641110/src/Discord.Net.WebSocket/ConnectionManager.cs#L51)) provide a derived Exception type. If there is an InnerException ([e.g](https://github.com/discord-net/Discord.Net/blob/2c93363653d7935cd142b664b7abc63fd2641110/src/Discord.Net.WebSocket/ConnectionManager.cs#L46)) wrap it in a derived Exception type, or do not wrap it.

AnalogFeelings:
### Description
Removes instances of generic exceptions in the code.

### Changes
- Removes instances of the generic Exception class.
- Removes weird `return Task.Delay(0);` in connection manager.

### Related Issues
- Closes #1456.

### DIFF
--- a/experiment/Discord.Net.BuildOverrides/BuildOverrides.cs
+++ b/experiment/Discord.Net.BuildOverrides/BuildOverrides.cs
@@ -263,7 +263,7 @@ namespace Discord
                 var result = await client.PostAsync($"{ApiUrl}/overrides/{id}/dependency", new StringContent($"{{ \"info\": \"{name}\"}}", Encoding.UTF8, "application/json"));
 
                 if (!result.IsSuccessStatusCode)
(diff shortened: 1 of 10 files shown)


AnalogFeelings:
the reason its a draft is because I want feedback on any exceptions that should be changed :suspect:

Misha-133 (approved):

PR #2922, opened 2024-05-08
Fix: NRE with rest interactions & no api call (again)

quinchs:
when trying to resolve a guild member from resolvable data, the client *assumes* the guild is not null.. which is not the case when api req are off.


Misha-133 (approved):
lgtm

PR #2924, opened 2024-05-09
Fix null error when using a `SelectMenuBuilder` constructor

Hampo:
### Description
Fixed a bug where `Options` can be set to `null` when using a specific `SelectMenuBuilder` constructor

### Changes
- Made `Options` be set to `options ?? new()`

### Related Issues
Never created one


Misha-133:
Hey, thanks for the contribution, but that is in fact wrong. `Options` can be set to `null` and that is intended behavior

Hampo:
> Hey, thanks for the contribution, but that is in fact wrong. `Options` can be set to `null` and that is intended behavior

Then should `AddOption` initialise the property if it's not set? What's the use case of a `null` `Options`?

Misha-133:
> Then should AddOption initialise the property if it's not set?

Hm, yeah that would be correct, might be just an oversight in the current implementation

> What's the use case of a `null` `Options`

auto-populated select menus (user, role, mentionable, channel)

Hampo:
Ah cool - I'll change this PR to update `AddOption` then

Hampo:
Updated another reference I found too

Misha-133 (approved):

PR #2932, opened 2024-05-16
Config Bypass for CanSendResponse

### LINKED ISSUE #2900: Request: Easy bypass for Defer 3 Second clock sync issue
# The Issue

This code here:
https://github.com/discord-net/Discord.Net/blob/d14d89dd18b618f8e37e11a45aa2f26d70dfbfe3/src/Discord.Net.Rest/Entities/Interactions/InteractionHelper.cs#L19-L22

wherein `ResponseTimeLimit = 3.0`

Is triggered when trying to `DeferAsync`, and throws an exception if it's false.

This means that the end system clock must be *exactly* in sync with Discord's server clocks\*, otherwise it *will* misbehave and throw invalid exceptions.
*\*(If we assume latency of ~1 second from network travel and initial processing, the local system clock only needs to be 2 seconds delayed for this call to be a guaranteed reject every time)*

While of course we all dream of running services in datacenters with perfect atomic clock sync, in reality bots are running wherever, even eg on a home computer (that's where I run while testing my bot for example), and in real world applications, a few seconds of offset of the system clock is very normal.

# The Proposed Solution

In general I disagree with the idea that this check should be made at all (should let the Discord remote server respond with a refusal if it's too late)\*, however I'd be more than happy with just a way to manually disable the check that I would personally be using at least in any `dev` environment. If I can just `DeferAsync(unchecked: true)`, or set `DiscordSocketConfig.NoDeferTimeCheck = true` or anything like that, that would suffice.

*\*(Noting as well that eg if the delay is 2.9 seconds, the check will pass and send along... and then after network latency, Discord will end up rejecting it anyway, ie we already rely on Discord's servers handling appropriate rejections here)*

The other valid solution here is to reformulate the check to base the comparison _relative to when the interaction was received by Discord.Net_ rather than relative to Discord's clocks, as that will work as-expected for any/all cases as long as local system clock doesn't jump during processing (can use `Environment.TickCount64` to ensure even that case is not mishandled).

WhyNot180:
### Description

This adds an option in the config to easily bypass the internal time check on responses. This should be useful for use in a dev environment where latency issues are more prevalent.

This check is a little redundant, since Discord will refuse a late request anyways, but this should allow for a little more leeway without making too many changes.

### Changes

- Adds `ResponseInternalTimeCheck` bool to `DiscordConfig` and `BaseDiscordClient`
- Adds check to each use of `InteractionHelper.CanSendResponse`

### Related Issues

- resolves #2900

### DIFF
--- a/src/Discord.Net.Core/DiscordConfig.cs
+++ b/src/Discord.Net.Core/DiscordConfig.cs
@@ -206,6 +206,14 @@ namespace Discord
         public bool UseInteractionSnowflakeDate { get; set; } = true;
 
         /// <summary>
+        ///     Gets or sets whether or not DeferAsync will make an internal expiration check.
+        /// </summary>
+        /// <remarks>
+        ///     This should generally be set to <see langword="false"/> in a development environment due to potential latency issues.
+        /// </remarks>
+        public bool DeferAsyncTimeCheck { get; set; } = true;
+
+        /// <summary>
         ///     Gets or sets if the Rest/Socket user <see cref="object.ToString"/> override formats the string in respect to bidirectional unicode.
         /// </summary>
         /// <remarks>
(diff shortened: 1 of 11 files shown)


Misha-133 (approved):
LGTM

PR #2958, opened 2024-07-12
fix: crash when poll is null in RestMessageComponent

Zaratusa:
### Description
The poll isn't check for null values everywhere which causes crashes when using the Rest API

### Changes
- Fixed missing null checks for poll in RestMessageComponent


Misha-133:
Nice catch! Thank you for the contribution

Misha-133 (approved):

PR #2979, opened 2024-08-11
Create a (Discord)Color using a string

Adrigorithm:
## Description
It allows for Colo~~u~~rs to be created from strings using constructor overloading. Please accept my first PRussy to Discord.Net :^)

## Changes
- new constructor taking a string (and an enum for possible future color formats))
- tests for this constructor

### DIFF
--- a/src/Discord.Net.Core/Entities/Roles/Color.cs
+++ b/src/Discord.Net.Core/Entities/Roles/Color.cs
@@ -200,6 +200,13 @@ namespace Discord
         public static implicit operator uint(Color color)
             => color.RawValue;
 
+        /// <summary>
+        /// Converts the string representation of a color to a Color object.
+        /// </summary>
+        /// <param name="rawValue">String to be parsed to a color</param>
+        /// <param name="colorType">Color format of the string</param>
+        /// <returns>A Color object with a color value parsed from a string</returns>
+        /// <exception cref="ArgumentOutOfRangeException">The argument is not a number or of incorrect length</exception>
         public static Color Parse(string rawValue, ColorType colorType = ColorType.CssHexColor)
         {
             if (TryParse(rawValue, out var color, colorType))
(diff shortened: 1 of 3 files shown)


Misha-133:
Since it's parsing a string value, the logic should go into a static `Parse` method
Could you do that? And also make a `bool TryParse()` one please

Adrigorithm:
Added the bitmasking and uh the `if (rawValue.StartsWith("0x"))` was needed because people can actually create hex codes that include leading 0's :(

src/Discord.Net.Core/Entities/Roles/Color.cs
```diff
@@ -200,6 +200,94 @@ public Color(float r, float g, float b)
         public static implicit operator uint(Color color)
             => color.RawValue;
 
+        /// <summary>
+        /// Converts the string representation of a color to a Color object.
```
Misha-133:
```suggestion
        ///     Converts the string representation of a color to a Color object.
```

src/Discord.Net.Core/Entities/Roles/Color.cs
```diff
@@ -200,6 +200,94 @@ public Color(float r, float g, float b)
...
+        public static Color Parse(string rawValue, ColorType colorType = ColorType.CssHexColor)
+        {
+            if (TryParse(rawValue, out var color, colorType))
+                return color;
+
+            throw new ArgumentOutOfRangeException(nameof(rawValue), "Value must be a number of valid length - matching the specified ColorType");
+        }
+
+        /// <summary>
+        /// Converts the string representation of a color to a Color object. The return value indicates whether the conversion succeeded.
```
Misha-133:
```suggestion
        ///     Converts the string representation of a color to a Color object. The return value indicates whether the conversion succeeded.
```

Misha-133 (approved):

itslychee:
image

PR #2999, opened 2024-09-11
option to dump gw payloads on errors

Misha-133:
just a neat feature for debugging lib issues


MrCakeSlayer (approved):

quinchs:
Might be worth including a warning in the xml doc about leaking potential PII via the exception *cough cough dcs cough cough*

Misha-133:
real

PR #3022, opened 2024-11-08
Add Opus.

SineVector241:
### Description
I've added nearly all of the opus's functions and implemented safe handlers for thread safety and reference safety. So it is easier to dispose of opus native objects using an IDisposable wrapper implementation.

Things that are missing are CTL definitions and Error Codes however these can optionally be implemented. Also this code only works with libopus v1.5 and above.

If you want, I can add all the opus cross platform binaries or workflow to this project.

### Changes
- Added NativeOpus.cs
- Added OpusDecoderSafeHandle.cs
- Added OpusDREDDecoderSafeHandle.cs
- Added OpusDREDSafeHandle.cs
- Added OpusEncoderSafeHandle.cs
- Added OpusMSDecoderSafeHandle.cs
- Added OpusMSEncoderSafeHandle.cs
- Added OpusRepacketizerSafeHandle.cs
- Created directory /Opus and /Opus/SafeHandlers in Discord.NET.V4.Audio
- Allowed unsafe blocks in Discord.NET.V4.Audio (You could probably make the NativeOpus code be able to be compiled without having to use unsafe but it requires a bit more thinking.)
- 
### Related Issues
N.A.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -210,3 +210,6 @@ docs/api/\.manifest
 
 # Codealike UID
 codealike.json
+
+# This is so coderush data isn't included...
+.cr/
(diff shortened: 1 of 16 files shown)


SineVector241:
Also note, this code is part of my other project/library (OpusSharp) however I have added it for freedom of use for Discord.NET

kalebbroo:
As a user I would love to see this implemented!

SineVector241:
Some functions still aren't correct. I'm still going through them and double checking them as well as documenting them. I will update the pr after I finish some tests on the fixes I've made.

SineVector241:
This is outdated now

SineVector241:
Severely outdated, just install the lib at this point...

PR #3091, opened 2025-04-05
Fixed modals with value type properties

Suiram1701:
### Description
It is not realy possible to use [RespondWithModalAsync<T>](https://github.com/Suiram1701/Discord.Net/blob/a542c33fe7d57a03830d00f64d3a31fef2c9331c/src/Discord.Net.Interactions/Extensions/IDiscordInteractionExtensions.cs#L57) with provided modal value due to problems with the property access and unfavorable string conversion (in my case with DateTime).

The changes allow it to use value types and search first after a ToString override before casting.

### Changes
- Cast the property value into object in the PropertyGetter methods in [ReflectionUtils.cs](https://github.com/Suiram1701/Discord.Net/blob/a542c33fe7d57a03830d00f64d3a31fef2c9331c/src/Discord.Net.Interactions/Utilities/ReflectionUtils.cs#L176)
- Improved the string conversion done in [RespondWithModalAsync<T>](https://github.com/Suiram1701/Discord.Net/blob/a542c33fe7d57a03830d00f64d3a31fef2c9331c/src/Discord.Net.Interactions/Extensions/IDiscordInteractionExtensions.cs#L57)

### Related Issues
N/A

### DIFF
--- a/src/Discord.Net.Interactions/Utilities/ReflectionUtils.cs
+++ b/src/Discord.Net.Interactions/Utilities/ReflectionUtils.cs
@@ -177,7 +177,8 @@ namespace Discord.Interactions
         {
             var instanceParam = Expression.Parameter(typeof(T), "instance");
             var prop = Expression.Property(instanceParam, propertyInfo);
-            return Expression.Lambda<Func<T, object>>(prop, instanceParam).Compile();
+            var cast = Expression.Convert(prop, typeof(object));
+            return Expression.Lambda<Func<T, object>>(cast, instanceParam).Compile();
         }
 
         internal static Func<T, object> CreateLambdaPropertyGetter(Type type, PropertyInfo propertyInfo)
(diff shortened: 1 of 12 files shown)


src/Discord.Net.Interactions/Utilities/ReflectionUtils.cs
```diff
@@ -227,5 +229,29 @@ internal static Func<object[], object[], T> CreateLambdaMemberInit(TypeInfo type
...
+        {
+            type ??= typeof(T);
+
+            do
+            {
+#if NET6_0_OR_GREATER
+                var method = type.GetMethod(nameof(ToString), bindingAttr: BindingFlags.Public | BindingFlags.Instance, types: Type.EmptyTypes);
+#else
+                var method = type.GetMethods(BindingFlags.Public | BindingFlags.Instance)
+                    .SingleOrDefault(m => m.Name == nameof(ToString) && m.GetParameters().Length == 0);
```
Cenngo, line 247: I could not remember the actual performance  of Get* reflection methods but ideally we shouldnt be calling these reflection methods on every respond call, inside a loop. Caching the result and storing it in the ModalInfo should be better. However after going over the modal code, I think it was an oversight on my part that ComponentTypeConverters are not being used to convert the Property values into API objects. I think it would be better to implement a `Write` logic similar to the one used in Slash Command TypeConverters to future-proof this if/when Discord decides to actually roll out new Modal inputs. Also it would make the logic modular and allow user customization.

Suiram1701: I've moved the check whether the type overrides ToString in a lazy field in the TextComponentInfo. At first I wanted implement method like you mentiont
```csharp
public virtual object Write(InputComponentInfo componentInfo, object value)
```
with the string conversion as a default implemention. But this isn't the target of this PR and you definitly have more knowladge in the area.

Cenngo: I will try to create a ComponentTypeConverter workflow for a solution like you and I mentioned hopefully in the near future. That said, you are right, focus of this PR is different.

Cenngo (approved):

Suiram1701:
@Misha-133 Could you merge this PR if its alright or do wait for something?

PR #3139, opened 2025-05-28
Make `MessageComponent.Empty` public

d4n3436:
### Description
This PR changes the access modifier of `MessageComponent.Empty` so it becomes public (no idea why it wasn't in the first place)

### Changes
- Changed access modifier of `MessageComponent.Empty` to `public`


Misha-133 (approved):
You could set components to `null` to remove then, but I guess a `null`-object might come a nice addition

PR #3148, opened 2025-06-23
fix(IsLoadableModule): return false for abstract classes

OverwrittenCode:
### Description
`IsLoadableModule` did not check the property `TypeInfo.IsAbstract`. Therefore, an unnecessary warning log stated that it could not be loaded.

Now, if a class is abstract, it will be ignored safely and all the inheritors will continue to load any commands.

### Changes
- fix(IsLoadableModule): return false for abstract classes
- fix: handle other attributes
- refactor: simplify attribute logic
- fix: correct terminology in debug message

### Related Issues
N/A

### DIFF
--- a/src/Discord.Net.Interactions/Builders/ModuleClassBuilder.cs
+++ b/src/Discord.Net.Interactions/Builders/ModuleClassBuilder.cs
@@ -63,7 +63,7 @@ namespace Discord.Interactions.Builders
                 result.Add(type.AsType(), moduleInfo);
             }
 
-            await commandService._cmdLogger.DebugAsync($"Successfully built {built.Count} Slash Command modules.").ConfigureAwait(false);
+            await commandService._cmdLogger.DebugAsync($"Successfully built {built.Count} interaction modules.").ConfigureAwait(false);
 
             return result;
         }
--- a/src/Discord.Net.Interactions/Builders/ModuleClassBuilder.cs
+++ b/src/Discord.Net.Interactions/Builders/ModuleClassBuilder.cs
@@ -18,7 +18,7 @@ namespace Discord.Interactions.Builders
         {
             static bool IsLoadableModule(TypeInfo info)
             {
-                return info.DeclaredMethods.Any(x => x.GetCustomAttribute<SlashCommandAttribute>() != null);
+                return !info.IsAbstract && info.DeclaredMethods.Any(x => x.GetCustomAttribute<SlashCommandAttribute>() != null);
             }
 
             var result = new List<TypeInfo>();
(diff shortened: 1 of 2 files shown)


src/Discord.Net.Interactions/Builders/ModuleClassBuilder.cs
```diff
@@ -18,7 +18,7 @@ public static async Task<IEnumerable<TypeInfo>> SearchAsync(Assembly assembly, I
         {
             static bool IsLoadableModule(TypeInfo info)
             {
-                return info.DeclaredMethods.Any(x => x.GetCustomAttribute<SlashCommandAttribute>() != null);
+                return !info.IsAbstract && info.DeclaredMethods.Any(x => x.GetCustomAttribute<SlashCommandAttribute>() != null);
```
Cenngo: I think doing something like this for the attribute check would fix the problem I was mentioning.
```suggestion
               return !info.IsAbstract && info.DeclaredMethods.SelectMany(x => x.GetCustomAttributes()).Any(x => x switch
    {
        SlashCommandAttribute or ComponentInteractionAttribute or ContextCommandAttribute or AutocompleteCommandAttribute or ModalInteractionAttribute => true,
        _ => false
    });
```
I tested this but can you please also test it so we don't run into any issues later.

Also the only other part as far as I can see is on #L61, it should say interaction module instead of slash command module. We cannot add comments on uneffected lines for some reason so I have to write it this way sorry. Other than that the PR should be good to go.

Cenngo (changes requested):

OverwrittenCode:
@Cenngo your suggestion looks completely fine - I have:
- Accepted your suggestion and slightly modified it so the switch statement is no longer needed.
- Corrected the debug message

~~Do you think we can mark this PR as Ready for review?~~

Hold on, we might be missing some.

Screenshot of the [`src/Discord.Net.Interactions/Attributes/Commands`](https://github.com/OverwrittenCode/Discord.Net/tree/Mark-Abstract-As-Not-Loadable-Module/src/Discord.Net.Interactions/Attributes/Commands) folder for reference:

src/Discord.Net.Interactions/Attributes/Commands

- [x] `AutocompleteCommandAttribute`
- [x] `ComponentInteractionAttribute`
- [x] `ContextCommandAttribute`
- [ ] `MessageCommandAttribute`
- [x] `ModalInteractionAttribute`
- [x] `SlashCommandAttribute`
- [ ] `UserCommandAttribute`

Should I include `MessageCommandAttribute` and `UserCommandAttribute` before marking this PR ready for review?

Also, it may help to sort the `is` pattern matching alphabetically so it is easier to compare to the source folder.

Cenngo:
> Should I include `MessageCommandAttribute` and `UserCommandAttribute` before marking this PR ready for review?

That wouldnt be necessary since both of those already inherit the `ContextCommandAttribute`. The PR should be good to go as is.

Cenngo (approved):

OverwrittenCode:
Perfect, PR is now ready to merge.

PR #3177, opened 2025-08-19
Make Cacheable constructors public

GreemDev:
### Description
This PR simply makes the constructors for the 2 different Cacheable structs public. 

I want to use this mechanism in my bot's code without duplicating the class in my own project but obviously I can if this functionality is not desired.

I did not add an XML doc, however I can add one if that is desired.

### Changes
`internal` constructor for `Cacheable<TCachedEntity, TDownloadableEntity, TRelationship, TId>` made `public`
`internal` constructor for `Cacheable<TEntity, TId>` made `public`

### Related Issues
N/A


Misha-133 (approved):
I don't see a reason for it to be `internal`.

PR #3185, opened 2025-09-06
[Feature] Component Designer

quinchs:
## Summary

This PR adds a 'html' like way to design and build components, specifically v2 components, providing design-time error analysis based off of the constraints of the Discord API.

Currently, the support is pretty basic, only allowing discords defined components, but in the future this could be expanded apon to create stateful and custom components. 

#### Performance

The custom XML parser is a one-and-done full parser, and could be greatly improved with incremental parsing, given the source tree from the C# generator. This would greatly speed up the parsing of the CXML.

As of right now, after the parse step, the entire component node tree is regenerated, which again could benefit from incremental mutations

#### Testing

There are no tests for any of this, and that smells, some good tests for parsing, diagnostic reporting, and so on are needed.

### Future prospects

As mentioned earlier, the possibilities with this feature can extend out to a full dom-like renderer and state manager, making stateful cards super trivial to implement:
```cs
public class Counter : CXmlComponent
{
    public int Count { get; set; }
    
    public override IMessageComponentBuilder Render()
    {
        return cx(
            $"""
            <text>
                You've clicked {Count} times!
            </text>
            <row>
                <button customId="clickme" onClick={HandleClick}/>
            </row>
            """
        );
    }

    public void HandleClick(IDiscordInteraction interaction)
    {
        Count += 1;
    }
}
```

The possibilities are quite literally endless with this.

#### Custom IDE language plugin
It would also be trivial to implement a custom LSP for the components, and provide both VS, VSC, and jetbrains plugins to support them in the `StringSyntax` attribute used for the basic highlighting in the CXML strings


quinchs:
Moved to https://github.com/discord-net/ComponentDesigner

PR #3189, opened 2025-09-17
[Feature] Modal Select Components Support for IF

Cenngo:
This PR adds support for FileUpload, UserSelect, RoleSelect, ChannelSelect, MentionableSelect and SelectMenu to Interaction Framework. 

To make modal generation/response flow more modular and make granular control possible, a new TypeConverter type called ModalComponentTypeConverter has been introduced. The new TypeConverter exposes a `WriteAsync<>()` method to allow the alteration of input component parameters when a `RespondWithModal<>()` extension method is called. 4 default TypeConverters are included in this PR for the following types: Array, Enum, IConvertible, Nullable. 
With the default setup, User/Channel/Role/Mentionable select values can be automatically mapped to `IUser[]`/`IRole[]`/`IChannel[]` arrays. For SelectMenu components, just like with message component SelectMenus, the converter falls back to a TypeReader for converting the returned string values to the underlying CLR type. Enums can be used to auto-populate SelectMenu options and returned values can be mapped to regular enums if a single-select is used, to flags enums if a multi-select is used. `[HideAttribute]` used for SlashCommand parameters, can be used to hide select enum values,
and a new `[SelectMenuOptionAttribute]` can be used to define description, emote, and IsDefault values for individul enum values. A new [ModalSelectMenuOptionAttribute] can be used on properties marked with [ModalSelectMenuInputAttribute] to add select menu options decleratively. Nullable types can be used which use the ModalComponentTypeConverter assigned to the underlying type for convertion. TextInput and single-select menu values can be converted to `IConvertible` types. Instance values of the component properties are used to supply the default values of components. The default behaviour of all of the converters is that they use the property value of the modal instance to populate their respective components' default values. As always, every default behaviour defined in the default typeconverters can be overridden by defining custom typecoverters for the respective clr type. Text display component cannot have any assigned typeconverters and must be a type of string, its content is populated using the property value of the instance and falls back to the content provided in the attribute. [HideAttribute] now features an overrideable Predicate() method to selectively hide enum options on runtime. 

Changes:
- Added description field to [InputLabelAttribute]
- Added [ModalFileUploadInputAttribute], [ModalMentionableSelectInputAttribute], [ModalRoleSelectInputAttribute], [ModalSelectMenuInputAttribute], [ModalTextDisplayAttribute], [ModalUserSelectInputAttribute]
- Created IModalComponent entities to pave the road for more non-input components

Breaking Changes:
- ModalBuilder.AddTextComponent() has been renamed to .AddTextInputComponent()
- ModalBuilder.Components property now inherits from a stripped down version of IInputComponentBuilder which is called IModalComponentBuilder
- InputComponentInfo now uses the new ModalComponentTypeConverter instead of ComponentTypeConverter with Write capabilities
- ModalInfo.Components property now inherits from a stripped down version of InputComponentInfo which is called ModalComponentInfo (new InputComponents property can be used as a drop-in replacement)


src/Discord.Net.Interactions/Attributes/HideAttribute.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    /// <summary>
+    ///     Can be optionally implemented by inherited types to conditionally hide an enum value.
+    /// </summary>
+    /// <remarks>
+    ///     Only runs on prior to modal construction. For slash command parameters, this method is ignored.
+    /// </remarks>
+    /// <param name="interaction">Interaction that <see cref="IDiscordInteractionExtentions.RespondWithModalAsync{T}(IDiscordInteraction, string, T, RequestOptions, Action{ModalBuilder})"/> is called on.</param>
+    /// <returns>
+    ///     <see cref="true"/> if the attribute should be active and hide the value.
```
Misha-133:
```suggestion
    ///     <see langword="true"/> if the attribute should be active and hide the value.
```

Cenngo: yea my mistake but already fixed it

src/Discord.Net.Interactions/Builders/Modals/Components/RoleSelectComponentBuilder.cs
```diff
@@ -0,0 +1,73 @@
...
+    /// </summary>
+    /// <param name="roles">The roles to add as a default value.</param>
+    /// <returns>
+    ///     The builder instance.
+    /// </returns>
+    public RoleSelectComponentBuilder AddDefaultValues(IEnumerable<IRole> roles)
+    {
+        _defaultValues.AddRange(roles.Select(x => new SelectMenuDefaultValue(x.Id, SelectDefaultValueType.Role)));
+        return this;
+    }
```
Misha-133: can we simplify this with `params IEnumerable<>` maybe?
The lang feature was introduced in c#13 iirc, though I don't recall whether it can be used with net8.0

src/Discord.Net.Interactions/Builders/ModuleClassBuilder.cs
```diff
@@ -659,6 +679,147 @@ private static void BuildTextInput(TextInputComponentBuilder builder, PropertyIn
...
+                    case RequiredInputAttribute requiredInput:
+                        builder.IsRequired = requiredInput.IsRequired;
+                        break;
+                    case InputLabelAttribute inputLabel:
+                        builder.Label = inputLabel.Label;
+                        builder.Description = inputLabel.Description;
+                        break;
+                    case ModalSelectMenuOptionAttribute selectMenuOption:
+
+                        if (!Emote.TryParse(selectMenuOption.Emote, out var emote) && !string.IsNullOrEmpty(selectMenuOption.Emote))
```
Misha-133: I'm fairly certain select menu options can have a unicode emoji too. This only covers custom guild/app ones.

I'd inverse the order of expressions here too for better readability.

Cenngo: i didnt know about the unicode emote ill try to implement it over the weekend bu about the logic inversion, i cant cuz roslyn is dumb and wont compile saying 'use of uninitialized variable'. Thats what i tried first

src/Discord.Net.Interactions/Builders/Modals/Components/RoleSelectComponentBuilder.cs
Cenngo: i will try and see

src/Discord.Net.Interactions/Builders/ModuleClassBuilder.cs
Cenngo: The emote property now can be in either an emote or and emoji format.

Misha-133 (approved):

PR #3193, opened 2025-10-10
Remove unsupported SDK targets

Misha-133:
With .net 10 we get even more build issues due to supporting the unsupported .net framework and .net releases. This PR finally drops support of those, to help us work more efficiently on the library while not dragging a long tail of legacy behind

P.S. 
you can vote in the poll in our [support server](https://discord.gg/dnet); poll will be up for 1 week.


SylveonDeko:
Finally!

src/Discord.Net.Interactions/Discord.Net.Interactions.csproj
```diff
@@ -20,12 +20,8 @@
...
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="8.0.0" />
-    <PackageReference Include="System.Collections.Immutable" Version="8.0.0" />
-    <PackageReference Include="System.Reactive" Version="6.0.1" />
-  </ItemGroup>
-  <ItemGroup Condition="$(TargetFramework.StartsWith('net4')) AND '$(MSBuildRuntimeType)' == 'Core'">
-    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="9.0.9" />
+    <PackageReference Include="System.Reactive" Version="6.1.0" />
```
d4n3436: This reference can be removed (it's unused)

```suggestion
```

src/Discord.Net.Core/Discord.Net.Core.csproj
```diff
@@ -14,15 +14,10 @@
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
-    <PackageReference Include="System.Interactive.Async" Version="6.0.1" />
-    <PackageReference Include="System.Collections.Immutable" Version="8.0.0" />
-    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.4" />
+    <PackageReference Include="System.Interactive.Async" Version="6.0.3" />
```
d4n3436: Microsoft recommends replacing `System.Interactive.Async` with `System.Linq.AsyncEnumerable` (which is now part of .NET 10, and is provided to older versions via a NuGet package)
https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/10.0/asyncenumerable
It's a prerelease version but it should be safe to replace.
```suggestion
    <PackageReference Include="System.Linq.AsyncEnumerable" Version="10.0.0-rc.1.25451.107" />
```

d4n3436 (changes requested):

src/Discord.Net.Core/Discord.Net.Core.csproj
Misha-133: well it seems it isn't a simple drop in replacement. I'm gonna have to refactor some things.

d4n3436: Looks like [the method](https://github.com/dotnet/reactive/blob/1ce85f4066360193cea9da5b94def4c32552c3be/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ForEach.cs#L30) was not transferred to the new package, which makes sense.

PR #3216, opened 2025-12-30
Improve .NET 9.0+ locking performance

MarkCiliaVincenti:
Uses a source generator so does not add a dependency.


Misha-133 (approved):
Thanks for the contribution

MarkCiliaVincenti:
Cool! This trusts you will keep those InternalsVisibleTo, otherwise you may want to consider not using it as a source generator, but so far all's well and it works.

PR #3217, opened 2025-12-30
Switch lock backport package to #if defs instead

SquirrelKiev:
The addition of a new package and the details surrounding that new package somewhat irks me, so this PR removes said new package and replaces it with some simple `#if NET9_0_OR_GREATER`s. Should accomplish the same thing but without adding new stuff to the csproj.

### DIFF
--- a/src/Discord.Net.Commands/Map/CommandMapNode.cs
+++ b/src/Discord.Net.Commands/Map/CommandMapNode.cs
@@ -12,7 +12,11 @@ namespace Discord.Commands
 
         private readonly ConcurrentDictionary<string, CommandMapNode> _nodes;
         private readonly string _name;
+#if NET9_0_OR_GREATER
         private readonly Lock _lockObj = new();
+#else
+        private readonly object _lockObj = new();
+#endif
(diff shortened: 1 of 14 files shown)


MarkCiliaVincenti:
The package is set to run in source generator mode, which means that it won't actually add a dependency.

SquirrelKiev:
Not formally no but from what I can tell it just copies the runtime package into the source of the project.
But even if that's fine, the whole point of the package is to provide the APIs that lock provides to <NET 9, but we're not even using those - the entire package can be replaced with some `#if NET9_0_OR_GREATER`s with no consequence or extra build dependencies. This feels like the package was just added for the sake of adding the package.

MarkCiliaVincenti:
Right, but those directives are both an eyesore and limiting in scope. Yes, currently only the lock keyword is used but the polyfill allows for using the whole API.

For starters, using the preprocessor directives won't allow you to lock with a timeout, or you'd have to use preprocessor directives on the timeout lock call too.

Misha-133 (approved):

PR #3221, opened 2026-01-03
[Fix] Modal Write invocation without instance and missing ChannelTypes

Cenngo:
- Fixes the modal building logic to run `ModalComponentTypeConverter.WriteAsync()` even when no modal instance is provided. 
- Adds the missing channelTypes property to the ChannelSelect builder and info classes.
- Removes the public setter requirement from modal component properties and adds enforcement to only the input components
- Added channel type constraint mapping from property type to single channel select
- Misc refactors for code brevity
- Added Id parameters to modal components

### DIFF
--- a/src/Discord.Net.Interactions/TypeConverters/ModalComponents/DefaultArrayModalComponentConverter.cs
+++ b/src/Discord.Net.Interactions/TypeConverters/ModalComponents/DefaultArrayModalComponentConverter.cs
@@ -1,5 +1,7 @@
+using Discord.Interactions.Utilities;
 using Discord.Utils;
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
@@ -12,6 +14,7 @@ internal sealed class DefaultArrayModalComponentConverter<T> : ModalComponentTyp
     private readonly Type _underlyingType;
     private readonly TypeReader _typeReader;
     private readonly ImmutableArray<ChannelType> _channelTypes;
+    private readonly ImmutableArray<EnumSelectMenuOption> _enumOptions;
 
     public DefaultArrayModalComponentConverter(InteractionService interactionService)
     {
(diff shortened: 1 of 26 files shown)


src/Discord.Net.Interactions/TypeConverters/ModalComponents/DefaultArrayModalComponentConverter.cs
```diff
@@ -134,49 +138,46 @@ public override Task WriteAsync<TBuilder>(TBuilder builder, IDiscordInteraction
...
-                selectMenu.WithDefaultValues(SelectMenuDefaultValue.FromChannel(channel));
-                break;
-            case IMentionable mentionable:
-                selectMenu.WithDefaultValues(mentionable switch
+            IEnumerable<IUser> defaultUsers => defaultUsers.Select(SelectMenuDefaultValue.FromUser).ToList(),
+            IEnumerable<IRole> defaultRoles => defaultRoles.Select(SelectMenuDefaultValue.FromRole).ToList(),
+            IEnumerable<IChannel> defaultChannels =>
+                defaultChannels.Select(SelectMenuDefaultValue.FromChannel).ToList(),
+            IEnumerable<IMentionable> defaultMentionables => defaultMentionables
+                .Where(x => x is IUser or IRole or IChannel)
```
Misha-133: Not exactly an issue, but
`mentionable` only includes users & roles, not channels. In DNet we also have `IChannel` implement `IMentionable` for the `Mention` property abstraction, but that's not the case on discord's end.
ig we should still handle the case with channels so we don't throw if the dev provides a channel entity, but we probably would want to warn the dev in some way.

Cenngo: Alright, then this should do it

Misha-133 (approved):

src/Discord.Net.Interactions/TypeConverters/ModalComponents/EnumModalComponentConverter.cs
Misha-133:
```suggestion
        if (option.Type is not ComponentType.SelectMenu and not ComponentType.TextInput)
```
Probably a logic error

Misha-133 (approved):

PR #3222, opened 2026-01-04
feat: libdave support

quinchs:
## Summary
This PR adds `libdave` support into the audio client, as well as some small audio client refractors.

### Remarks
A new project, `Discord.Net.Dave` was created in the `tools/` folder that provides the low level bindings (`libdave` class) and some higher level wrappers around them (`Dave`, `DaveSession`, etc).

Since this is a low level binding, manual memory management is required, the current allocs/FFI boundaries should be investigated for any leaking memory. 

### Notice on breaking changes
This PR is implemented in a way to prevent breaking change, `libdave` e2ee support is gated behind a new configuration option `EnableVoiceDaveEncryption` within the socket config. The default value for the new configuration option is `false`, so users will have to opt-in to `libdave` support. This is up for discussion as Discord plans to require `libdave` e2ee by march.

I don't know how we want to distribute `libdave` along with dnet when it becomes required, do we bundle builds in with the WS project? I'm happy for feedback on this.

---

If you have an app with audio functionality, It'd be great if you can test this PR and report any bugs you find with it.


docs/guides/voice/libdave.md
```diff
@@ -0,0 +1,38 @@
+---
+uid: Guides.Voice.LibDave
+title: Using libdave e2ee
+---
+
+# Setup
+
+You'll first need a build of Discords' [libdave](https://github.com/discord/libdave) library as either `libdave.dll`, `libdave.so` or `libdave.o` in your apps executing directory.
```
Misha-133: Or available in the `PATH`

src/Discord.Net.WebSocket/Audio/AudioClient.cs
```diff
@@ -294,116 +370,165 @@ internal async Task RemoveInputStreamAsync(ulong userId)
...
             try
             {
                 switch (opCode)
                 {
                     case VoiceOpCode.Ready:
-                        {
-                            await _audioLogger.DebugAsync("Received Ready").ConfigureAwait(false);
-                            var data = (payload as JToken).ToObject<ReadyEvent>(_serializer);
+                    {
+                        var data = (payload as JToken).ToObject<ReadyEvent>(_serializer);
```
Misha-133: What's the point of doing `as` cast since if it fails the `ToObject` will cause an NRE to be thrown anyway? a hard cast would be (negligibly, but still) faster while achieving the same result.

This probably isn't specifically related to the PR tho, iirc that's the case in the WS package too.

tools/Discord.Net.Dave/Dave.cs
```diff
@@ -0,0 +1,129 @@
...
+
+    /// <summary>
+    ///     The protocol version representing dave e2ee being disabled.
+    /// </summary>
+    public const int DISABELD_PROTOCOL_VERSION = 0;
+
+    /// <summary>
+    ///     The expected epoch of a new MLS group.
+    /// </summary>
+    public const ulong MLS_NEW_GROUP_EXPECTED_EPOCH = 1;
```
Misha-133: Capital snek case seems a bit weird imho. even as per microsoft's c# [naming conventions](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names#naming-conventions) public class members should be in `PascalCase`

tools/Discord.Net.Dave/Utils.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Discord.LibDave;
+
+/// <summary>
+///     A class containing utilities related to <c>libdave</c>
```
Misha-133:
```suggestion
///     A class containing utilities related to <c>libdave</c>.
```
no dot :(

tools/Discord.Net.Dave/Dave.cs
quinchs: I usually use capital snake case for constants to distinguish them between properties/fields, down to change this though

src/Discord.Net.WebSocket/Audio/AudioClient.cs
quinchs: I was following the existing pattern in the switch, happy to update this

Misha-133 (approved):

quinchs:
It might be a good idea to expose either the `DaveSession` object on `IAudioClient`, or expose some method to get the verification code from the `IAudioClient`

docs/guides/voice/libdave.md
```diff
@@ -0,0 +1,20 @@
+---
+uid: Guides.Voice.LibDave
+title: Using libdave e2ee
+---
+
+# Setup
+
+You'll first need a build of Discords' [libdave](https://github.com/discord/libdave) library as either `libdave.dll`, `libdave.so` or `libdave.o` in your apps executing directory or `PATH`.
```
MrCakeSlayer: `Discord's` - Discord is not plural by default, this apostrophe is in the wrong place.
`app's` - Missing apostrophe :P

src/Discord.Net.WebSocket/Audio/DaveSessionManager.cs
```diff
@@ -0,0 +1,293 @@
...
+        _client = client;
+        _decryptors = [];
+        _preparedTransitions = [];
+        _session = Dave.CreateSession();
+        Encryptor = Dave.CreateEncryptor();
+        _logger = client.Discord.LogManager.CreateLogger($"Dave #{clientId}");
+
+        _session.OnMLSFailure += (source, reason) =>
+        {
+            // I don't like this :/
```
MrCakeSlayer: Lol, probably could use a more descriptive comment

MrCakeSlayer (approved):

moiph:
Will test this out this weekend

Update: weekend arrived, so far successful testing. Flood of debug output from logs, about 50 per second of "Buffer under run".

Buffer under run
(D:\a\libdave\libdave\cpp\src\dave\frame_processors.cpp:191) Encrypted frame is too small to contain min supplemental bytes

src/Discord.Net.WebSocket/Audio/Streams/DaveDecryptStream.cs
```diff
@@ -0,0 +1,91 @@
...
+using Discord.LibDave.Binding;
+using Discord.Logging;
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Discord.Audio.Streams;
+
+public sealed class DaveDecryptStream : AudioOutStream
```
Misha-133, line 15: No xmldocs in a public class

src/Discord.Net.WebSocket/Audio/Streams/DaveEncryptStream.cs
```diff
@@ -0,0 +1,75 @@
+﻿using Discord.LibDave;
+using Discord.LibDave.Binding;
+using Discord.Logging;
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Discord.Audio.Streams;
```
Misha-133, line 8: Here too

src/Discord.Net.WebSocket/Audio/AudioClient.cs
```diff
@@ -99,10 +107,25 @@ internal AudioClient(SocketGuild guild, int clientId, ulong channelId)
...
+
+            _dave = null;
+
+            // if (Discord.LibDaveEnabled is null)
+            // {
+            //     if()
+            // }
+
+            // if (Discord.LibDaveEnabled)
+            //     _dave = new(this, clientId);
```
Misha-133: Feels a bit weird to leave this in

src/Discord.Net.WebSocket/Audio/AudioClient.cs
```diff
@@ -126,16 +149,55 @@ internal Task StopAsync(StopReason stopReason)
...
 
+        private async ValueTask SetupLibDave()
+        {
+            var isAvailable = Dave.CheckAvailability();
+            switch (Discord.LibDaveEnabled, isAvailable)
+            {
+                case (null, false):
+                    await _audioLogger.WarningAsync(
+                        "libdave will be required for receiving and transmitting audio by March 1st, 2026, please" +
+                        " checkout the documentation for enabling libdave support: https://docs.discordnet.dev/guides/voice/libdave.html"
```
Misha-133:
```suggestion
                        " check out the documentation for enabling libdave support: https://docs.discordnet.dev/guides/voice/libdave.html"
```
not a git operation :)

src/Discord.Net.WebSocket/Audio/AudioClient.cs
```diff
@@ -271,139 +383,205 @@ internal async Task CreateInputStreamAsync(ulong userId)
...
-                            if (data.Mode != DiscordVoiceAPIClient.Mode)
-                                throw new InvalidOperationException($"Discord selected an unexpected mode: {data.Mode}");
+                        if (data.Mode != DiscordVoiceAPIClient.Mode)
+                            throw new InvalidOperationException($"Discord selected an unexpected mode: {data.Mode}");
 
-                            SecretKey = data.SecretKey;
-                            await SetSpeakingAsync(false);
-                            _keepaliveTask = RunKeepaliveAsync(_connection.CancelToken);
+                        SecretKey = data.SecretKey;
+                        //await SetSpeakingAsync(false);
```
Misha-133: Is this not needed anymore? imho this should either explain why is it left out or just be removed

brunoamancio:
Hi, please take a look at the the PR I created to be merged into this branch: https://github.com/discord-net/Discord.Net/pull/3241

PR #3252, opened 2026-03-15
Handle TaskCanceledException in TypingNotifier

bc3tech:
### Description

When typing is canceled by a Task Cancellation token, `TaskCanceledException` is thrown, not `OperationCanceledException`. Adding this as a swallow to the typing indicator.

### Changes

Added catch block to typing indicator to capture canceled tasks.

### DIFF
--- a/src/Discord.Net.Rest/Utils/TypingNotifier.cs
+++ b/src/Discord.Net.Rest/Utils/TypingNotifier.cs
@@ -38,6 +38,7 @@ namespace Discord.Rest
                 }
             }
             catch (OperationCanceledException) { }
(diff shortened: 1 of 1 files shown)


quinchs:
image

bc3tech:
do you have a UT/IT that validates this works as expected, then? I am cancelling a typing indicator and getting a TCE up in my app

quinchs:
Are you able to share a stacktrace? the code you changed already handles `TaskCanceledException` because of inheritance. There is no UT for this, but by just looking at the code I can assert that `TaskCanceledException` does not get thrown there

PR #3265, opened 2026-04-27
Add GET voice-state REST wrappers

yury-opolev:
Wraps two voice-state GET endpoints that round out the PATCH pair already in `DiscordRestApiClient`:

- `GET /guilds/{guild.id}/voice-states/@me`
- `GET /guilds/{guild.id}/voice-states/{user.id}`

Both return `null` on 404 (the user is not currently in any voice channel of the guild).

Exposed as `GetMyVoiceStateAsync` and `GetUserVoiceStateAsync` on `RestGuild` and `SocketGuild`, returning a new `RestVoiceState` entity that mirrors the relevant fields from the REST response (channel id, mute/deaf/suppress/streaming/video flags, voice-session id, request-to-speak timestamp).

## Why a new `RestVoiceState` rather than reusing `IVoiceState`

The REST response carries `channel_id`, not a full channel object. Implementing `IVoiceState` would require returning `null` for `IVoiceChannel VoiceChannel` even when the user *is* in voice (since `null` already means "not in voice" per the existing interface contract), which would be misleading. Exposing `VoiceChannelId : ulong?` directly avoids that ambiguity; callers who need the channel can resolve it via `guild.GetVoiceChannelAsync(state.VoiceChannelId.Value)`.

Happy to revisit if maintainers prefer a different shape (e.g. implementing `IVoiceState` with a documented null contract, or pre-fetching the channel via an extra REST call inside `GuildHelper`).

## Tests

Skipped — the change is REST glue and the 404→null mapping isn't reachable from the unit-test project's public surface. Live integration coverage exists in spirit via the existing `RestGuildFixture` flow but I did not add a test there to keep the patch self-contained. Happy to add tests if maintainers point at the right entry point.

### DIFF
--- a/src/Discord.Net.WebSocket/Entities/Users/SocketVoiceState.cs
+++ b/src/Discord.Net.WebSocket/Entities/Users/SocketVoiceState.cs
@@ -34,6 +34,10 @@ namespace Discord.WebSocket
         ///     Gets the voice channel that the user is currently in; or <see langword="null" /> if none.
         /// </summary>
         public SocketVoiceChannel VoiceChannel { get; }
+
+        /// <inheritdoc />
+        public ulong? VoiceChannelId { get; }
+
         /// <inheritdoc />
         public string VoiceSessionId { get; }
         /// <inheritdoc/>
@@ -58,6 +62,7 @@ namespace Discord.WebSocket
         internal SocketVoiceState(SocketVoiceChannel voiceChannel, DateTimeOffset? requestToSpeak, string sessionId, bool isSelfMuted, bool isSelfDeafened, bool isMuted, bool isDeafened, bool isSuppressed, bool isStream, bool isVideo)
         {
             VoiceChannel = voiceChannel;
+            VoiceChannelId = voiceChannel.Id;
             VoiceSessionId = sessionId;
             RequestToSpeakTimestamp = requestToSpeak;
 
--- a/src/Discord.Net.Core/Entities/Guilds/IGuild.cs
+++ b/src/Discord.Net.Core/Entities/Guilds/IGuild.cs
@@ -1481,5 +1481,26 @@ namespace Discord
         ///     Modifies the current user in this guild.
         /// </summary>
         Task ModifyCurrentUserAsync(Action<SelfGuildUserProperties> props, RequestOptions options = null);
+
+        /// <summary>
+        ///     Gets the current user's voice state in this guild.
+        /// </summary>
+        /// <param name="options">The options to be used when sending the request.</param>
+        /// <returns>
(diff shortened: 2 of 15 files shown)


Misha-133:
I think we should still reuse the `IVoiceState` interface. While it is technically breaking the contract, it's a part of an outdated API model so imho it's fine. Also adding a separate `VoiceChannelId` property should help with it. 
I will add the changes to the PR.

Also thanks for the contribution :))

Misha-133 (approved):

src/Discord.Net.WebSocket/Entities/Users/SocketVoiceState.cs
```diff
@@ -58,6 +62,7 @@ private enum Flags : byte
         internal SocketVoiceState(SocketVoiceChannel voiceChannel, DateTimeOffset? requestToSpeak, string sessionId, bool isSelfMuted, bool isSelfDeafened, bool isMuted, bool isDeafened, bool isSuppressed, bool isStream, bool isVideo)
         {
             VoiceChannel = voiceChannel;
+            VoiceChannelId = voiceChannel.Id;
```
ROBdk97, line 65:
```suggestion
            VoiceChannelId = voiceChannel?.Id;
```
Missing null check


This is causing a error for me if you join to/from no voice channel
```
Error handling Dispatch (VOICE_STATE_UPDATE)
System.TypeInitializationException: The type initializer for 'Discord.WebSocket.SocketVoiceState' threw an exception.
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Discord.WebSocket.SocketVoiceState..ctor(SocketVoiceChannel voiceChannel, Nullable`1 requestToSpeak, String sessionId, Boolean isSelfMuted, Boolean isSelfDeafened, Boolean isMuted, Boolean isDeafened, Boolean isSuppressed, Boolean isStream, Boolean isVideo)
   at Discord.WebSocket.SocketVoiceState..cctor()
   --- End of inner exception stack trace ---
   at Discord.WebSocket.DiscordSocketClient.ProcessMessageAsync(GatewayOpCode opCode, Nullable`1 seq, String type, Object payload)
   at Discord.WebSocket.SocketVoiceState..ctor(SocketVoiceChannel voiceChannel, Nullable`1 requestToSpeak, String sessionId, Boolean isSelfMuted, Boolean isSelfDeafened, Boolean isMuted, Boolean isDeafened, Boolean isSuppressed, Boolean isStream, Boolean isVideo)
   at Discord.WebSocket.SocketVoiceState..cctor()
```

I will create a Issue

ROBdk97: #3275

PR #3266, opened 2026-04-27
Add play audio sample

josago97:
### Description
An example of audio playback in a voice channel has been added.


Misha-133:
Hey
Thanks for the PR
I genuinely do think having an actual example of voice API usage is beneficial, but I don't feel comfortable with having the sample download 3rd party binaries (and other resources)
I think it should be presented in form of an instruction, so it's up to the end user ro decide whether to download & install tools/libraries or not. Also that would allow it to stay portable across platforms.

We have an [existing doc page](https://docs.discordnet.dev/guides/voice/sending-voice.html) that is lacking in that regard. It was never fully refined and is also using old language constructs, while more modern approaches exist.

I also would suggest using a package manager like winget on windows/ apt/yum/other ones on linux. It generally leads to better user experience not having to.

josago97:
Hello, I understand the concern about downloading and running third-party programs. My idea in creating this example is to make it as user-friendly as possible, like I've seen in other audio libraries, and to keep it always up-to-date. The binaries are obtained from their official sites: Ffmpeg is obtained from the public link on its official website, libdave is taken from the official Discord repository, and Opus is obtained from this repository. Therefore, they can be considered reliable, and these links are referenced in the Discord.NET documentation. Furthermore, anyone can view the code and see what will be downloaded.

I can modify it to support Linux systems when downloading the binaries, if possible.
I can also require user confirmation before anything is downloaded, so that everything is done with their permission.
Regarding using Winget, the problem I see is that it will install more things than necessary and may modify environment variables unnecessarily. 

I await your instructions on what changes to make :)

Misha-133:
Hey
Sorry for a delayed response, I had some stuff happening IRL.

Regarding the PR. I'm not particularly familiar with the sites, so i did not recognize them; nor will a regular lib user. 
Even if these are first party distributions, I think the sample should
a. explicitly prompt the user that it will download binaries from a 3rd party source (could be a "yes/no" question in console with a warning message listing the sources & packages)
b. have comments/xmldoc in code describing what each site is and having links to pages.

Also Linux is the primary platform the lib will likely be used on, so support for one in the sample would be appreciated :3

josago97:
Done :3
In theory, it should be ready to work on Windows, Mac, and Linux. In practice, I've tested it working correctly on Windows and Linux (via WSL).

Misha-133 (approved):

PR #3268, opened 2026-05-15
feat: add support for getting messages from a guild (with filters)

OMEGA3065:
### Description
This Draft/PR would add support for getting all messages from a guild based on the specified filters.
This uses the "new" discord API endpoint: https://docs.discord.com/developers/resources/message#search-guild-messages

### Changes
- added a function to get filtered messages from a guild


src/Discord.Net.Core/Utils/Preconditions.cs
```diff
@@ -72,6 +72,22 @@ public static void NotNullOrWhitespace(Optional<string> obj, string name, string
         private static ArgumentException CreateNotEmptyException(string name, string msg)
             => new ArgumentException(message: msg ?? "Argument cannot be blank.", paramName: name);
 
+        /// <exception cref="ArgumentException">String size may not be equal to <paramref name="size"/>.</exception>
+        public static void NotEqualSize(Optional<string> obj, int size, string name, string msg = null) { if (obj.IsSpecified && obj.Value.Length == size) throw new ArgumentException(message: msg ?? $"String size may not be equal to {size}.", paramName: name); }
```
Misha-133: I think this should be called `Length` instead of `Size`, since the property is called `string.Length`

src/Discord.Net.Core/Utils/Preconditions.cs
```diff
@@ -402,6 +418,25 @@ public static void Options(string name, string description)
 
         #endregion
 
+        #region Array Size
+
+        /// <exception cref="ArgumentException">Array size may not be equal to <paramref name="size"/>.</exception>
+        public static void NotEqualSize<T>(Optional<T[]> obj, int size, string name, string msg = null) { if (obj.IsSpecified && obj.Value.Length == size) throw new ArgumentException(message: msg ?? $"Array size may not be equal to {size}.", paramName: name); }
```
Misha-133: Same here

src/Discord.Net.Rest/Entities/Guilds/GuildHelper.cs
```diff
@@ -178,6 +178,29 @@ public static async Task<GuildIncidentsData> ModifyGuildIncidentActionsAsync(IGu
...
+        public static async Task<GuildMessageSearchData> GetGuildMessagesAsync(IGuild guild, BaseDiscordClient client, GetGuildMessagesParams args, RequestOptions options = null)
+        {
+            var model = await client.ApiClient.GetGuildMessagesAsync(guild.Id, args, options);
+
+            var builder = ImmutableArray.CreateBuilder<RestMessage>();
+            if (model.ParseMessages)
+                foreach (var msg in model.Messages)
+                {
+                    var author = MessageHelper.GetAuthor(client, guild, msg.Author.Value, msg.WebhookId.ToNullable());
+                    var channel = GetChannelAsync(guild, client, msg.ChannelId, options).Result;
```
Misha-133: 1. This should use the async API instead of blocking the thread with a `.Result`
2. this should use `guild.GetChannelAsync` instead, since it would allow it to get channels from `SocketGuild`'s state cache instead of performing a REST API call for each channel

Misha-133 (changes requested):

src/Discord.Net.Rest/Entities/Guilds/GuildHelper.cs
OMEGA3065: > 2. this should use `guild.GetChannelAsync` instead, since it would allow it to get channels from `SocketGuild`'s state cache instead of performing a REST API call for each channel

Should I allow it download the data or only allow getting from cache?

Misha-133 (approved):

PR #3284, opened 2026-08-21
Give some extra info when no selectmenu options are given

RobinLaevaert:
### Description
Currently when building a select menu when no options are given and sending a message with it, we get the following error:

The server responded with error 50035: Invalid Form Body
Inner Errors:
BASE_TYPE_BAD_LENGTH: Must be between 1 and 25 in length.

It can probably be figured out what exactly is going wrong, but it's not super clear at first.

### Changes
Add some checks and error throwing when building select menu component

### DIFF
--- a/src/Discord.Net.Core/Entities/Interactions/MessageComponents/Builders/SelectMenuBuilder.cs
+++ b/src/Discord.Net.Core/Entities/Interactions/MessageComponents/Builders/SelectMenuBuilder.cs
@@ -438,6 +438,12 @@ public class SelectMenuBuilder : IInteractableComponentBuilder
     {
         var options = Options?.Select(x => x.Build()).ToList();
 
+        if (options == null)
+            throw new ArgumentNullException(nameof(Options), "Options must have a value.");
+
+        if (options.Count is 0 or > MaxOptionCount)
+            throw new ArgumentOutOfRangeException(nameof(Options), $"Number of select menu options must be in range [1, {MaxOptionCount}]");
+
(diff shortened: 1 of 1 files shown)


src/Discord.Net.Core/Entities/Interactions/MessageComponents/Builders/SelectMenuBuilder.cs
```diff
@@ -438,6 +438,12 @@ public SelectMenuComponent Build()
     {
         var options = Options?.Select(x => x.Build()).ToList();
 
+        if (options == null)
```
Misha-133:
```suggestion
        if (options is null)
```

src/Discord.Net.Core/Entities/Interactions/MessageComponents/Builders/SelectMenuBuilder.cs
```diff
@@ -438,6 +438,12 @@ public SelectMenuComponent Build()
...
+            throw new ArgumentNullException(nameof(Options), "Options must have a value.");
+
+        if (options.Count is 0 or > MaxOptionCount)
```
Misha-133: This should should use `Preconditions.AtLeast / AtMost`

src/Discord.Net.Core/Entities/Interactions/MessageComponents/Builders/SelectMenuBuilder.cs
Misha-133: These checks are only valid in case the select menu is a regular string select one. It should check whether it's a string select, or a user/role/channel/mentionable one.

Misha-133 (changes requested):

Misha-133 (approved):

